PR #22, opened 2026-05-06
feat(web): add job filters and artifact labels

ebarti:
## Summary
- add projection-backed job facets and exact multi-value/date/fit score filters
- update the Jobs toolbar with location, company, job title, discovered date, and fit score controls
- canonicalize duplicate resume PDF artifacts and show explicit tailored resume labels

## Validation
- pnpm api:test
- pnpm web:build
- pnpm check
- Browser: verified Jobs filters and Artifacts labels at http://127.0.0.1:5173/ against API port 8767

### DIFF
--- a/apps/api/src/projections.ts
+++ b/apps/api/src/projections.ts
@@ -893,12 +893,13 @@ function collectArtifacts(
     );
     for (const row of rows) {
       if (!row.path) continue;
-      const key = `${row.artifact_type}:${row.path}`;
+      const artifactType = canonicalArtifactType(row.artifact_type);
+      const key = `${artifactType}:${row.path}`;
       if (seen.has(key)) continue;
       seen.add(key);
       out.push({
         artifactId: row.artifact_id || key,
-        artifactType: row.artifact_type || "artifact",
+        artifactType,
         status: row.status || "active",
         localPath: row.path,
         sizeBytes: nullableNumber(row.size_bytes),
@@ -923,12 +924,13 @@ function collectArtifacts(
     );
     for (const row of rows) {
       if (!row.path) continue;
-      const key = `${row.artifact_type || "artifact"}:${row.path}`;
+      const artifactType = canonicalArtifactType(row.artifact_type);
+      const key = `${artifactType}:${row.path}`;
       if (seen.has(key)) continue;
       seen.add(key);
       out.push({
         artifactId: String(row.row_id ?? key),
-        artifactType: row.artifact_type || "artifact",
+        artifactType,
         status: row.status || "active",
         localPath: row.path,
         sizeBytes: nullableNumber(row.size_bytes),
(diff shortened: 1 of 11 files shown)


apps/api/src/projections.ts
```diff
@@ -893,12 +893,13 @@ function collectArtifacts(
     );
     for (const row of rows) {
       if (!row.path) continue;
-      const key = `${row.artifact_type}:${row.path}`;
+      const artifactType = canonicalArtifactType(row.artifact_type);
```
ebarti, line 896: PR Reviewer - Codex

Severity: Medium
Gate: Pass
Evidence: `apps/api/src/projections.ts:896` and `apps/api/src/projections.ts:927` now canonicalize artifact aliases in the TS projection refresher, but `workers/automation/src/jobhunter/infrastructure/projections/projection_builder.py:540-607` still dedupes on raw `(artifact_type, local_path)` and stores the raw `artifact_type`. Both projection builders share the `operations_projections` watermark; if Python advances it, TS can skip at `apps/api/src/projections.ts:398-401`, leaving raw/duplicate artifact projection rows.

This TS-side canonicalization fixes the API refresher path, but the Python mirror can still materialize the old aliases and duplicates. Because the Python file is not in this PR diff, this comment is anchored to the closest changed TS line that introduced the canonicalization requirement.

Required fix: Apply the same `resume_pdf -> tailored_resume_pdf`, `tailored_resume -> tailored_resume_txt`, and `cover_letter -> cover_letter_txt` canonicalization in the Python ProjectionBuilder, or canonicalize/dedupe on API read so Python-built projections preserve alias compatibility.

ebarti: PR Reviewer - Codex

Severity: Medium
Gate: Pass
Evidence: `apps/api/src/projections.ts:896` and `workers/automation/src/jobhunter/infrastructure/projections/projection_builder.py:564` now canonicalize artifact types during rebuild, but both refresh loops return without rebuilding when there are no dirty jobs and the shared `operations_projections` watermark is current (`apps/api/src/projections.ts:398-401`, Python `ProjectionBuilder.refresh()` at `workers/automation/src/jobhunter/infrastructure/projections/projection_builder.py:191-192`). Existing `artifact_list_projections` rows that were materialized before this change can therefore keep raw `resume_pdf` / duplicate canonical rows until a future event dirties each job. The API read path still returns projection rows as stored (`apps/api/src/read-model.ts:247-260`) instead of canonicalizing/deduping stale rows.

The Python parity part of the earlier finding is now fixed for future rebuilds, but the upgrade path is still incomplete. A local database that already has duplicate legacy/canonical artifact projection rows and an up-to-date watermark will hit the early return before the new canonicalization runs, so `/v1/artifacts` can continue returning duplicate/raw artifact rows even though the PR goal is to dedupe legacy/canonical resume PDF artifacts.

Required fix: Add a one-time projection backfill/version bump for this canonicalization change, or canonicalize and dedupe artifact rows on the read path (`listArtifacts` and job detail artifacts) so already-materialized projection rows are normalized even when no new `job_events` rows exist.

PR #35, opened 2026-05-07
feat(workers): add pipeline activities + JobPipelineWorkflow + ApplyWorkflow

ebarti:
## Summary

PR 2 of the Temporal + Worker Reliability stack. Defines one Temporal **Activity** per pipeline stage and adds the **`JobPipelineWorkflow`** + **`ApplyWorkflow`** workflows that drive them. Wires the worker registry from PR 1 to bind the real workflows + activities so `jobhunter worker` boots into production wiring (no more bootstrap no-op).

What landed:

- `refactor(pipeline)`: split `pipeline.py` into a `pipeline/` package so `pipeline/workflow.py` has somewhere to live; `STAGE_ORDER`, `run_pipeline`, etc. are re-exported from `pipeline/__init__.py` so existing imports keep working.
- `feat(workers)`: per-stage Temporal activities — `discover`, `enrich`, `score`, `tailor`, `cover`, `pdf`, `apply`, `profile_import`. Each is a thin adapter over the existing stage runner (or `apply.launcher.main` / `actions.run_local_action` where that's cleanest). Activities heartbeat for cancellation propagation; the apply activity runs `apply_main` on a worker thread under `asyncio.shield` and heartbeats every 15 s.
- `feat(workers)`: `JobPipelineWorkflow` drives the configured pipeline stages serially **in batch mode** against eligible jobs in the local DB. Stage eligibility is owned by the underlying runner via `state.set_stage_state`, not by the workflow. Passing `"apply"` raises a non-retryable `ApplicationError` pointing at `ApplyWorkflow`. Default `RetryPolicy(max_attempts=3)` per stage.
- `feat(workers)`: `ApplyWorkflow` is a single-activity, **per-job** workflow with the apply-specific retry policy (`max_attempts=2`) and parameter shape (so PR 3 can register `apply` as `mode=workflow` cleanly). `apply_activity` re-raises transient failures so the retry policy can fire; `LookupError` is wrapped in a non-retryable `ApplicationError`.
- `feat(workers/temporal)`: pass the `jobhunter` package through the workflow sandbox so workflow modules can construct activity-input dataclasses at the workflow boundary without sandbox-init failures (the proxy mechanism otherwise refuses to instantiate frozen dataclasses imported through `imports_passed_through()`).
- `feat(cli)`: `jobhunter worker` imports `WORKFLOWS, ACTIVITIES` from `infrastructure.temporal.registry` and passes them to `build_worker(...)`. The PR-1 `_BootstrapNoOpWorkflow` is now skipped because the registry is non-empty.
- `test(workers)`: full coverage — happy-path test per activity, `JobPipelineWorkflow` (ordered execution, unknown-stage rejection, apply-stage rejection, validation_mode forwarding, activity failure, cancellation propagation), `ApplyWorkflow` (success + transient-retry + retry-recovery + non-retryable LookupError paths), real-registry build under the production sandbox runner, registry decoration smoke test.

## Stack context

This is **PR 2 of 7** in [the Temporal + Worker Reliability stack](../blob/temporal/pipeline-workflow/docs/plans/proposed/2026-05-07-temporal-and-worker-reliability-stack.md).

- Base: `temporal/foundation` (PR 1).
- Will rebase to `main` once PR 1 lands.

## What's NOT in this PR

- JSON-RPC `fire_and_forget` cut-over to `mode="workflow"` — that's **PR 3**. `infrastructure/rpc/` is unchanged here.
- `apply_runs` table collapse + workflow-history projector — **PR 4**.
- UI Workflow Runs view + invalidation routing — **PR 5**.
- Removing `actions.run_local_action` — it stays as the synchronous entry point used by tests and the CLI's pre-Temporal command paths. PR 3 removes it from the JSON-RPC layer.

## Residual risks (deferred)

These were flagged by the PR reviewer and are intentionally deferred — see the inline thread replies for details:

- **Per-(TenantId, JobId) batching for non-apply stages.** Today the discover / enrich / score / tailor / cover / pdf runners are batch-only — they walk DB selectors over the entire `jobs` table. Only `ApplyWorkflow` is truly per-job. Refactoring the runners to accept a single `job_url` so `JobPipelineWorkflow` can drive per-job batches is tracked in [`docs/backlog.md`](../blob/temporal/pipeline-workflow/docs/backlog.md) under "Worker Reliability".
- **Heartbeat-tier `apply_activity` cancellation test.** The existing cancellation test exercises workflow-side propagation through a synthetic long-running activity; adding a test that verifies `asyncio.shield(apply_task)` keep-alive, the `wait_for` re-loop, and the executor-thread-vs-cancellation race in the real `apply_activity` is a deeper test-scaffolding refactor.
- **Top-level `jobhunter.materials` / `jobhunter.profile` package placement.** The reviewer noted these new top-level packages add a fourth namespace per concept. Moving them under `infrastructure/` (or to a new `infrastructure/temporal/activities/` namespace) is a wider reorganisation that touches the registry, the workflow file, and every test that imports from these paths. Can land as a follow-up if the namespace count starts to bother people.

## How to test locally

```bash
# 1. Boot the local Temporal dev server (see docs/local-development.md):
temporal server start-dev

# 2. In another shell, boot the JobHunter worker with the real registry:
uv --project workers/automation run jobhunter worker
# Expected log line:
#   JobHunter worker running on task queue jobhunter-default with 2 workflow(s) and 8 activity(ies) — Ctrl-C to stop.
# That message confirms the bootstrap no-op is no longer in play.

# 3. Confirm the doctor probe still reports Temporal reachable:
uv --project workers/automation run jobhunter doctor
```

## Verification

All commands run in this worktree against `temporal/pipeline-workflow` after the fixer commits.

| Command | Result |
| --- | --- |
| `uv --project workers/automation sync --extra dev` | ok |
| `uv --project workers/automation run --extra dev pytest -q` | **618 passed in 6.45 s** |
| `uv --project workers/automation run --extra dev ruff check .` | **All checks passed!** |
| `pnpm api:check` | ok (tsc clean) |
| `pnpm api:test` | **55 passed** in 759 ms |

QA: skipped — automation-only, no UI surface change.


workers/automation/src/jobhunter/pipeline/workflow.py
```diff
@@ -0,0 +1,236 @@
...
+    persistence still happens inside the stage runner via ``set_stage_state``.
+    """
+    pending = Pending(attempt_count=0, max_attempts=0)
+    enqueued = transition(pending, StageTransition.Enqueue)
+    if isinstance(enqueued, TransitionRejected):
+        return enqueued.reason
+    started = transition(enqueued, StageTransition.Start)
+    if isinstance(started, TransitionRejected):
+        return started.reason
+    return None
```
ebarti: PR Reviewer - Opus

Severity: High
Gate: Fail
Evidence: `_check_state_machine` constructs a fresh `Pending(attempt_count=0, max_attempts=0)` on every call (workflow.py:93-100). Per `domain/pipeline/state_machine.py` rows 1-2, `Pending → Queued` and `Pending → Running` are *always* allowed by the table, so this function can only ever return `None`. The `stages_skipped` branch in `JobPipelineWorkflow.run` (line 116-118) is therefore unreachable in production — only the test that monkeypatches `_check_state_machine` can exercise it. The PR description claims the workflow "consults `StageStateMachine.transition()` for `Pending -> Queued -> Running` validity (skipping when rejected)", but skipping requires looking up the *actual current state* of the stage in the DB (e.g., `Succeeded`, `Exhausted`, `Blocked`) — that lookup is missing.
Reproduction: Reasoned from code — `transition(Pending(...), Enqueue)` always returns `Queued`; `transition(Queued(...), Start)` always returns `Running`. No DB lookup happens.

Practical consequence: when a job has stage `score` already in `Succeeded` and the workflow is asked to re-run `score`, the workflow will dispatch `score_activity`, which will call `set_stage_state(state="running")`, which will raise `ValueError("Invalid state transition for score: succeeded -> running")` from `state.py:_validate_stage_transition`. The activity surfaces this as an `ActivityError`, the workflow retries 3× per `_DEFAULT_RETRY`, and the pipeline records the stage as `failed` rather than `skipped`. That is the exact opposite of the brief: re-running an already-succeeded stage produces a hard failure with retries, not a clean skip.

Required fix: either
1. Wire a real per-(job, stage) state lookup into `_check_state_machine` (read from `job_stage_states` via an activity, since DB I/O is forbidden inside workflows) and skip when the current state is `Succeeded` / `Exhausted` / `Blocked`; or
2. Drop `_check_state_machine` entirely from this PR and document explicitly in the PR description and the workflow docstring that skip-on-already-succeeded is deferred to a later PR. Leaving the no-op veto in place implies a guarantee the workflow does not provide and will mislead PR 3 / PR 4 reviewers.

Either path needs an updated test that exercises the *real* veto path (not a monkeypatched one).

workers/automation/src/jobhunter/pipeline/workflow.py
```diff
@@ -0,0 +1,236 @@
...
+                workers=payload.workers,
+            ),
+            start_to_close_timeout=_APPLY_TIMEOUT,
+            retry_policy=_APPLY_RETRY,
+            heartbeat_timeout=timedelta(seconds=60),
+        )
+    raise workflow.ApplicationError(
+        f"Unknown stage: {stage}",
+        non_retryable=True,
+    )
```
ebarti, line 180: PR Reviewer - Opus

Severity: High
Gate: Fail
Evidence: `JobPipelineWorkflowInput` accepts `job_url` (line 56) and the apply branch passes it through (line 203). Every other branch — discover, enrich, score, tailor, cover, pdf — silently drops `payload.job_url`. The activities all forward to `run_pipeline(stages=[stage], ...)`, which is a *batch* selector that walks `_PENDING_SQL` over the entire `jobs` table (see `pipeline/runner.py:311` + `_count_pending`). Result: starting `JobPipelineWorkflow(tenant_id="local", job_url="https://example.com/job/X", stages=["enrich"])` enriches every pending job in the DB, not job X.
Reproduction: Reasoned from code. `_execute_stage("enrich", payload)` only forwards `tenant_id`, `workers`, `limit` (workflow.py:146-156); `enrich_activity` calls `run_pipeline(stages=["enrich"], workers=..., limit=...)` (enrichment/activities.py:32); `run_pipeline → _resolve_stages → _run_sequential` walks the DB-wide selector. There is no per-job filter anywhere in that chain.

This is a load-bearing question for the rest of the stack:
- The review brief says the workflow is "per (TenantId, JobId)". The code does not implement that — it implements per-batch.
- The workflow ID convention in PR 3 / PR 4 will be incompatible with batch semantics: you cannot have a deterministic workflow ID per `(tenant, job)` if a single workflow run mutates many jobs.
- The `apply_runs` collapse in PR 4 assumes 1 workflow ↔ 1 (job, stage) — that mapping is currently broken for 6 of 7 stages.

Required fix: pick one and commit to it before this lands:
A) Per-job: introduce a per-job code path through the runners (e.g., `run_pipeline(stages=[...], job_url=payload.job_url)`) so each activity processes only the workflow-bound job. This is a meaningful runner change, not just a workflow change.
B) Per-batch: rename to `BatchPipelineWorkflow`, drop `job_url` from the workflow input, and rework the brief / downstream PR plan so PR 3-5 do not assume per-job. Move the per-job path to `JobPipelineWorkflow(job_url=...)` as a separate workflow.

Either path needs an explicit decision in the PR description and an updated planning doc — and the corresponding tests need to assert the chosen semantics.

workers/automation/src/jobhunter/apply/activities.py
```diff
@@ -0,0 +1,87 @@
...
+    except asyncio.CancelledError:
+        activity.logger.info("apply_activity cancelled")
+        raise
+    except Exception as exc:  # noqa: BLE001 — apply path must surface as structured failure
+        return ApplyActivityOutput(
+            status="failed",
+            applied=0,
+            failed=0,
+            error=str(exc),
+        )
```
ebarti: PR Reviewer - Opus

Severity: Medium
Gate: Pass
Evidence: `apply_activity` catches *every* exception other than `CancelledError` (lines 69-75) and returns `ApplyActivityOutput(status="failed", ...)`. Because the activity returns normally rather than raising, Temporal does not classify the call as a failure and the configured `RetryPolicy(maximum_attempts=2)` (workflow.py:80, apply/workflow.py:51) is never triggered. Effectively, apply has zero retry — `max_attempts=2` is decorative.
Reproduction: Reasoned from code. The retry policy only fires when an activity raises; structured-failure returns are treated as success by Temporal. `test_workflow_apply.py::test_apply_workflow_returns_failure_when_apply_main_raises` confirms `apply_main_mock.assert_called()` — but it is called *once*, not twice. (Test passes only because `assert_called` is satisfied by a single call.)

This may be intentional (apply is destructive — blind retries could resubmit applications), but the brief explicitly called for "tighter `max_attempts=2`". Two interpretations are possible:
- The retry policy is a placeholder for future behavior. Then the inline comment in `apply/activities.py` and `apply/workflow.py` should say so explicitly, and the policy block should be marked TODO.
- The retry policy is supposed to engage on transient failures (network, browser flakiness) and the swallow-all `except` is wrong. Then the activity should re-raise — at minimum on the exception classes that are safe to retry — and let the policy decide.

Required fix: pick one explicitly and document it in the activity docstring. If you keep the swallow-all, drop or comment-out the apply retry policy, since it cannot fire. If you keep the policy, narrow the bare `except Exception` to *non-retryable* failures only and re-raise the rest. Also tighten the test: `apply_main_mock.assert_called_once()` (not `assert_called`) so future regressions surface immediately.

workers/automation/src/jobhunter/infrastructure/temporal/worker.py
```diff
@@ -21,6 +25,14 @@ async def run(self) -> str:
         return "noop"
 
 
+# Pass jobhunter through the sandbox: workflow modules import activity dataclasses
+# whose modules transitively pull in rich/sqlite/etc. Activities run outside the
+# sandbox, so determinism is preserved.
+_PASSTHROUGH_RESTRICTIONS = SandboxRestrictions.default.with_passthrough_modules(
+    "jobhunter"
+)
```
ebarti, line 45: PR Reviewer - Opus

Severity: Medium
Gate: Pass
Evidence: The inline comment on line 28-30 says the passthrough is needed because workflow modules transitively pull in `rich/sqlite/etc`. That is not actually the failure mode here — every activity module in this PR (`apply/activities.py`, `discovery/activities.py`, etc.) defers heavy imports inside the activity function bodies, so their *top-level* imports are only `dataclasses`, `typing`, and `temporalio`. None of the workflow imports transitively pull in `rich`, `sqlite3`, etc. at module-load time.

I reproduced the actual failure mode by trying narrower passthroughs:
```
SandboxRestrictions.default.with_passthrough_modules(
    "jobhunter.discovery.activities",
    "jobhunter.enrichment.activities",
    ...
)
```
fails with `RuntimeError: Restriction state not present. Using subclasses of proxied objects is unsupported.` The real reason the broad passthrough is needed: the workflow constructs activity *dataclasses* (e.g. `EnrichActivityInput(...)`) at the workflow boundary, and the sandbox proxy mechanism does not allow constructing instances of frozen dataclasses imported through `imports_passed_through()`.

Reproduction: Ran `Worker(..., workflow_runner=SandboxedWorkflowRunner(restrictions=...))` with `WORKFLOWS` from the registry against four progressively narrower passthrough configurations. `with_passthrough_modules("jobhunter")` succeeds; passing through only the activity packages (`jobhunter.discovery`, etc.) plus `jobhunter.domain` still fails; adding `jobhunter.pipeline.workflow` and `jobhunter.apply.workflow` to the narrow set succeeds.

Required fix:
1. Update the inline comment to reflect the actual failure: workflow modules construct frozen activity-input dataclasses, and the sandbox proxy chain breaks on subclass construction. "rich/sqlite/etc" is misleading.
2. Either tighten the passthrough to the explicit module list (the activity packages + the workflow modules themselves + `jobhunter.domain`) or document explicitly that the broad `jobhunter` passthrough is acceptable because all `jobhunter` code is trusted and determinism is preserved by activities running outside the sandbox. The current code does the latter implicitly; making it explicit would prevent a future contributor from flagging it as a security gap.

Not blocking for this PR — the broad passthrough is functionally correct and determinism is preserved. The concern is documentation accuracy and intent.

workers/automation/tests/test_temporal_registry.py
```diff
@@ -0,0 +1,28 @@
...
+        )
+
+
+def test_registry_activities_are_temporal_activity_definitions():
+    for activity_fn in ACTIVITIES:
+        defn = getattr(activity_fn, "__temporal_activity_definition", None)
+        assert defn is not None, (
+            f"{getattr(activity_fn, '__name__', activity_fn)!r} is in "
+            "ACTIVITIES but is missing the @activity.defn decorator."
+        )
```
ebarti, line 28: PR Reviewer - Opus

Severity: Medium
Gate: Pass
Evidence: All 11 new tests in this PR pass `workflow_runner=UnsandboxedWorkflowRunner()` (test_workflow_job_pipeline.py:70, test_workflow_apply.py:35, every `test_activity_*.py`). The only test that exercises the production `build_worker(...)` sandbox configuration is the pre-existing `test_temporal_worker.py::test_build_worker_binds_to_jobhunter_task_queue`, but it passes `workflows=[]` and `activities=[]` — so the sandbox machinery never tries to validate any real workflow module. There is no test that asserts `build_worker(env.client, workflows=WORKFLOWS, activities=ACTIVITIES)` succeeds under the production `SandboxedWorkflowRunner` with the real registry.
Reproduction: Ran a quick proof: `build_worker(env.client, workflows=WORKFLOWS, activities=ACTIVITIES)` succeeds today. But if a future contributor narrows or removes the `with_passthrough_modules("jobhunter")` call (or adds a new workflow module that imports something the proxy cannot handle), no existing test will catch the boot-time `RuntimeError: Failed validating workflow JobPipelineWorkflow` until the next time someone runs `jobhunter worker` against a live Temporal server.

The implementer flagged this divergence in the review brief. It is a real test-vs-prod gap.

Required fix: add one test that calls `build_worker(env.client, workflows=WORKFLOWS, activities=ACTIVITIES)` (i.e. the *real* sandboxed runner) and asserts construction does not raise. This adds a few hundred ms to the suite and pins down the production wiring forever. Suggested location: `test_temporal_registry.py` or a new `test_temporal_worker_registry.py`.

workers/automation/tests/test_activity_apply.py
```diff
@@ -0,0 +1,73 @@
...
+    kwargs = apply_main_mock.call_args.kwargs
+    assert kwargs["limit"] == 2
+    assert kwargs["target_url"] == "https://example.com/job"
+    assert kwargs["min_score"] == 8
+    assert kwargs["headless"] is True
+    assert kwargs["model"] == "haiku"
+    assert output.status == "ok"
+    assert output.applied == 2
+    assert output.failed == 0
+    assert output.error is None
```
ebarti, line 73: PR Reviewer - Opus

Severity: Medium
Gate: Pass
Evidence: The cancellation propagation test (`test_workflow_job_pipeline.py::test_workflow_cancel_propagates_to_activity_as_cancelled_error`, line 213-242) uses a synthetic `_long_running_activity` — not the real `apply_activity`. The real `apply_activity` has a distinctive structure: it runs `apply_main` in `loop.run_in_executor`, wraps the future in `asyncio.shield`, and heartbeats every 15 s. None of that is exercised by an existing test. The synthetic activity does a simple `await asyncio.sleep(...)` loop and re-raises `CancelledError` cleanly — `apply_activity` has substantially more moving parts.

Specifically untested in `apply_activity`:
- `asyncio.shield(apply_task)` keeps the underlying executor task alive when `wait_for` times out — does the next iteration successfully resume waiting on the same future?
- When `wait_for` raises `TimeoutError`, the activity calls `activity.heartbeat("apply still running")` and re-loops. Does heartbeating actually fire?
- When the activity is cancelled, `apply_task` (the executor future) is not cancelled because of the shield. The thread keeps running. Does Temporal observe the cancellation correctly via the `CancelledError` re-raise on line 67?

Reproduction: Reasoned from code. `test_activity_apply.py` only covers the happy path (apply_main returns `(2, 0)`).

Required fix: add at least one test that:
1. Patches `apply_main` to block on a `threading.Event` (so the executor task does not complete quickly).
2. Cancels the wrapping workflow.
3. Asserts the workflow surfaces `CancelledError` as `WorkflowFailureError.cause`, that `activity.heartbeat` was called at least twice (initial + at least one timeout-driven heartbeat), and that the apply_main thread receives a stop signal so the test does not leak a hanging thread.

This catches future regressions in the executor / shield / heartbeat dance, which is the most subtle code in the PR.

workers/automation/src/jobhunter/infrastructure/temporal/registry.py
```diff
@@ -0,0 +1,42 @@
...
+    enrich_activity,
+    score_activity,
+    tailor_activity,
+    cover_activity,
+    pdf_activity,
+    apply_activity,
+    profile_import_activity,
+]
+
+__all__ = ["ACTIVITIES", "WORKFLOWS"]
```
ebarti, line 42: PR Reviewer - Opus

Severity: Medium
Gate: Pass
Evidence: Per `AGENTS.md` Documentation Requirements ("TypeScript API plus Python worker architecture, local-first boundaries, orchestration, or phased migration constraints" → `docs/architecture.md`), this PR introduces:
- A new `WORKFLOWS, ACTIVITIES` registry pattern in `infrastructure/temporal/registry.py`.
- Two production workflows (`JobPipelineWorkflow`, `ApplyWorkflow`).
- Eight production activities.
- A new `pipeline/` package layout (`pipeline.py` → `pipeline/{__init__,runner,workflow}.py`).
- A workflow-sandbox passthrough policy that downstream contributors need to know about before adding new workflow modules.

`docs/architecture.md` line 349 currently reads: "Pipeline workflows + activities are introduced in later PRs of this stack; this layer is the substrate they bind to." That sentence is now factually stale — this PR *is* the "later PR". A reader of `architecture.md` will believe the workflow layer is still empty.

Reproduction: `git diff temporal/foundation..HEAD -- docs/` returns nothing in this PR.

Required fix: update `docs/architecture.md` "Workflow Orchestration (Local Temporal)" section to:
1. Describe `JobPipelineWorkflow` (one-per-batch, drives stages serially) and `ApplyWorkflow` (single-activity, apply-specific retry).
2. Describe the `WORKFLOWS, ACTIVITIES` registry contract (`infrastructure/temporal/registry.py`).
3. Mention the `pipeline.py` → `pipeline/` package split, since `pipeline/runner.py` is now the home of `run_pipeline` and `pipeline/workflow.py` is the home of `JobPipelineWorkflow`.
4. Note the workflow-sandbox passthrough policy (and the reason — workflow-side construction of activity-input dataclasses requires the proxy to be bypassed for `jobhunter`).

This is a real doc requirement under `AGENTS.md`, not optional polish — the PR adds meaningful new capabilities to the architecture surface.

workers/automation/src/jobhunter/discovery/activities.py
```diff
@@ -0,0 +1,47 @@
...
+from dataclasses import dataclass, field
+from typing import Any
+
+from temporalio import activity
+
+
+@dataclass(frozen=True)
+class DiscoverActivityInput:
+    tenant_id: str
+    workers: int = 1
```
ebarti, line 20: PR Reviewer - Opus

Severity: Low
Gate: Pass
Evidence: Every activity input dataclass in this PR carries `tenant_id: str` as a required field (`DiscoverActivityInput`, `EnrichActivityInput`, `ScoreActivityInput`, `TailorActivityInput`, `CoverActivityInput`, `PdfActivityInput`, `ApplyActivityInput`, `ProfileImportActivityInput`). None of the activities or their downstream `run_pipeline` / `apply_main` / `run_local_action` paths use `tenant_id`. It is captured at the workflow boundary and dropped.
Reproduction: `grep -n "tenant_id" workers/automation/src/jobhunter/{discovery,enrichment,scoring,materials,profile,apply}/activities.py` returns 8 matches — all on the dataclass field declaration. Zero usages downstream.

This appears to be a forward-looking placeholder for hosted multi-tenant work, but as-is it is dead data flowing through the workflow boundary that:
- Implies a contract the system does not enforce (the activity does not honor the tenant scope).
- Wastes Temporal payload size and history bytes on every activity invocation.
- Could mislead a reviewer of PR 3 / PR 4 into thinking the pipeline is already tenant-aware.

Required fix: pick one explicitly:
1. If this is intentional placeholder for hosted: add a comment in each input dataclass saying "captured for hosted-future tenant scoping; ignored locally" and reference the planning doc that promises to wire it.
2. If `tenant_id` should already be honored locally: thread it through to the runners (the `LOCAL_TENANT` constant from `jobhunter.domain.tenant` would be a natural default).
3. If neither, drop the field — re-adding it later is a 1-line dataclass change.

workers/automation/src/jobhunter/pipeline/workflow.py
```diff
@@ -0,0 +1,236 @@
...
+        return await workflow.execute_activity(
+            cover_activity,
+            CoverActivityInput(
+                tenant_id=payload.tenant_id,
+                min_score=payload.min_score,
+                limit=payload.limit,
+            ),
+            start_to_close_timeout=_DEFAULT_TIMEOUT,
+            retry_policy=_DEFAULT_RETRY,
+        )
```
ebarti, line 164: PR Reviewer - Opus

Severity: Low
Gate: Pass
Evidence: `TailorActivityInput` (materials/activities.py:22) and `CoverActivityInput` (materials/activities.py:69) accept `validation_mode: str = "normal"` as a settable field, but `JobPipelineWorkflowInput` does not expose it. The workflow constructs both activity inputs without passing `validation_mode` (workflow.py:168-179 + 180-190), so the workflow always uses the default `"normal"`. CLI users who run `jobhunter run --validation lenient` get the lenient path; users who trigger the same flow through `JobPipelineWorkflow` cannot.
Reproduction: Reasoned from code. CLI handler passes `validation_mode` to `run_pipeline` (cli.py:606-614). Workflow does not.

Required fix: add `validation_mode: str = "normal"` to `JobPipelineWorkflowInput` and forward it in the `tailor` and `cover` branches of `_execute_stage`. Alternatively, document explicitly in the workflow input docstring why validation mode is not configurable through the workflow surface (e.g., "only `normal` is supported for workflow-driven runs in PR 2; broader options ship in PR 5 with the UI").

workers/automation/src/jobhunter/pipeline/workflow.py
```diff
@@ -0,0 +1,236 @@
...
+# at the workflow boundary even though we never bind their values.
+_OUTPUT_TYPES: tuple[type, ...] = (
+    DiscoverActivityOutput,
+    EnrichActivityOutput,
+    ScoreActivityOutput,
+    TailorActivityOutput,
+    CoverActivityOutput,
+    PdfActivityOutput,
+    ApplyActivityOutput,
+)
```
ebarti: PR Reviewer - Opus

Severity: Low
Gate: Pass
Evidence: `_OUTPUT_TYPES` (lines 226-236) is defined and never referenced anywhere — `grep -rn "_OUTPUT_TYPES" workers/automation/` returns one match (this declaration). The comment claims it exists "so static analysers know we use these output dataclasses at the workflow boundary", but the workflow code never binds the values, so the static analyser would flag the imports anyway. The actual mechanism the workflow uses is the explicit `from … import (DiscoverActivityOutput, …)` in the `imports_passed_through()` block — and Ruff is configured to allow unused imports through `__all__` or `noqa` directives, not through dead module-level tuples.

Required fix: either
1. Remove `_OUTPUT_TYPES` and add `# noqa: F401` comments on the output-type imports inside the `imports_passed_through()` block (matches established Ruff suppression style elsewhere in the codebase).
2. Drop the output-type imports entirely — the workflow never references them in the code path. They are only needed if the dataclass needs to be available for runtime type introspection, which is not happening here.

Either way, the current `_OUTPUT_TYPES` tuple is a low-grade smell that future contributors will trip over.

workers/automation/src/jobhunter/materials/__init__.py
```diff
@@ -0,0 +1 @@
+"""JobHunter materials package — Temporal activities for tailor, cover, pdf stages."""
```
ebarti, line 1: PR Reviewer - Opus

Severity: Low
Gate: Pass
Evidence: This PR creates two brand-new top-level packages — `jobhunter.materials` and `jobhunter.profile` — solely to host one Temporal activity module each. The codebase already has:
- `jobhunter.domain.materials` (domain types)
- `jobhunter.infrastructure.materials` (LaTeX adapter, repository)
- `jobhunter.domain.profile`
- `jobhunter.infrastructure.profile` (PDF parser)

Adding `jobhunter.materials` and `jobhunter.profile` at the top level introduces a fourth namespace per concept and a different placement convention than `discovery/`, `enrichment/`, `scoring/`, `apply/` (which all already existed at the top level housing their stage runners — adding `activities.py` next to them was natural). For materials/profile, no such top-level package existed; the implementer chose to invent one rather than co-locate the activity with the existing concept owner.

Reproduction: `find workers/automation/src/jobhunter -maxdepth 2 -type d -name "materials*" -o -name "profile*"` returns 6 directories — 2 of which (`jobhunter/materials/` and `jobhunter/profile/`) are new in this PR.

Required fix: pick one explicitly:
1. Move these activities into `jobhunter.infrastructure.materials.activities` and `jobhunter.infrastructure.profile.activities`, co-locating them with the repositories they coordinate with. Update `registry.py` accordingly.
2. Move them into a new `jobhunter.infrastructure.temporal.activities` package, treating activities as cross-cutting infrastructure.
3. Document the chosen convention (top-level package per stage) in `docs/architecture.md` and explain why the materials/profile activities are special-cased to avoid future contributors making different choices.

Not blocking for this PR. The current placement works. The concern is consistency and discoverability for the next person adding an activity.

workers/automation/src/jobhunter/pipeline/workflow.py
ebarti: PR Fixer - Claude

Fixed in 2b89499. Dropped `_check_state_machine` and the `stages_skipped` branch from `JobPipelineWorkflow.run`. The function only ever returned `None` because the synthetic `Pending(0,0) → Enqueue → Start` transition is always allowed by the table — it never consulted the real per-(job, stage) state. Real eligibility is owned by the underlying stage runner via `state.set_stage_state`, and that's what the workflow docstring now says. Replaced the synthetic-veto test with `test_pipeline_workflow_rejects_unknown_stage_as_non_retryable`, which asserts the existing `_execute_stage` fallback raises `ApplicationError(non_retryable=True)`.

workers/automation/src/jobhunter/pipeline/workflow.py
ebarti: PR Fixer - Claude

Fixed in 2b89499. Dropped `job_url` from `JobPipelineWorkflowInput` and removed the `apply` branch from `_execute_stage`. `JobPipelineWorkflow` now commits to batch semantics — its docstring explicitly says batch mode against eligible jobs in the local DB. Passing `"apply"` raises a non-retryable `ApplicationError("apply is not orchestrated by JobPipelineWorkflow; use ApplyWorkflow")`. The per-job apply path stays in `ApplyWorkflow`. The mismatched workflow-ID convention you flagged for PR 3 / PR 4 is now consistent: `JobPipelineWorkflow` is per-(TenantId, batch-of-stages); `ApplyWorkflow` is per-(TenantId, JobId). Added `test_pipeline_workflow_rejects_apply_stage_with_pointer_to_apply_workflow`. Per-(TenantId, JobId) batching for non-apply stages requires the runners to accept a `job_url` — out of scope for PR 2 — tracked in `docs/backlog.md` under Worker Reliability and called out in the plan's PR 2 entry.

workers/automation/src/jobhunter/apply/activities.py
ebarti: PR Fixer - Claude

Fixed in 34c264b. Narrowed the `except` block in `apply_activity`: `CancelledError` stays re-raised, `LookupError` is wrapped in a non-retryable `ApplicationError` (operator errors fail fast, no retry burn), all other exceptions propagate verbatim so Temporal's `max_attempts=2` retry policy fires on transient browser / network / executor failures. `ApplyWorkflow` already catches `ActivityError` and converts it to `ApplyWorkflowResult(status="failed", ...)`, so the workflow surface is unchanged. Added three assertions on `ApplyWorkflow`: a `RuntimeError` is invoked exactly twice (max_attempts) before surfacing, a `RuntimeError` then success returns ok with `applied=1` (retry recovery), and a `LookupError` is invoked exactly once and surfaces as a non-retryable failure.

workers/automation/src/jobhunter/infrastructure/temporal/worker.py
ebarti: PR Fixer - Claude

Fixed in 3d39b91. Reproduced your finding: top-level imports of activity modules are stdlib + temporalio only. The real failure mode for a narrower passthrough is "Restriction state not present. Using subclasses of proxied objects is unsupported." — the workflow constructs frozen activity-input dataclasses (`EnrichActivityInput(...)` etc.) at the workflow boundary and the sandbox proxy refuses to instantiate them through `imports_passed_through()`. Updated the inline comment to state that, and explained why the broad `with_passthrough_modules("jobhunter")` scope is intentional (a narrower scope would have to enumerate every activity-input dataclass module and would couple the worker bootstrap to the activity layer).

workers/automation/tests/test_temporal_registry.py
ebarti: PR Fixer - Claude

Fixed in 3d39b91. Added `test_build_worker_validates_real_registry_under_production_sandbox` in `tests/test_temporal_worker.py`. It uses `WorkflowEnvironment.start_local()` (not the unsandboxed runner) and calls `build_worker(env.client, workflows=WORKFLOWS, activities=ACTIVITIES)` with the real registry imports. This pins the sandbox-passthrough policy: any future narrowing of `with_passthrough_modules("jobhunter")` or any new workflow module the proxy can't handle will surface as a test failure here instead of a boot-time `RuntimeError` against a live Temporal server.

workers/automation/tests/test_activity_apply.py
ebarti: PR Fixer - Claude

Out of scope for this fixer pass — documented as a residual risk. The existing cancellation test (`test_workflow_cancel_propagates_to_activity_as_cancelled_error`) exercises the workflow-side propagation. Adding a heartbeat-tier test that also exercises `asyncio.shield(apply_task)` keep-alive, the `wait_for` re-loop, and the executor-thread-vs-cancellation race is a deeper refactor of the activity test scaffolding (need a `threading.Event` patched into `apply_main`, a way to verify the executor task survives the shield, a way to assert the heartbeat fires during the timeout loop). Tracking it as a follow-up; flagging in the PR description as residual risk so it isn't lost.

workers/automation/src/jobhunter/infrastructure/temporal/registry.py
ebarti: PR Fixer - Claude

Fixed in 6afb9af. Updated `docs/architecture.md` "Workflow Orchestration (Local Temporal)" section to describe (1) the per-stage activities under each bounded context, (2) `JobPipelineWorkflow` (batch-mode, non-apply stages) and `ApplyWorkflow` (per-job, apply-specific retry), (3) the `pipeline.py → pipeline/` package split and `infrastructure/temporal/registry.py` as the single source of truth for `WORKFLOWS` and `ACTIVITIES`, and (4) the rationale for `with_passthrough_modules("jobhunter")` so future contributors know why the scope is broad before they try to narrow it. Also brought `docs/plans/proposed/2026-05-07-temporal-and-worker-reliability-stack.md` PR 2 entry into the worktree and corrected the "per (TenantId, JobId)" claim there.

workers/automation/src/jobhunter/discovery/activities.py
ebarti: PR Fixer - Claude

Fixed in 1777ae0 (and 34c264b for `ApplyActivityInput`). Added a one-line comment on each activity-input dataclass: `# tenant_id is currently informational; runners read from LOCAL_TENANT until tenant scoping lands.` Kept the field on the wire because the hosted tenant-scoping work (PR 3 / PR 4) needs it; the comment is the cheap fix that prevents future reviewers from flagging the dead field again.

workers/automation/src/jobhunter/pipeline/workflow.py
ebarti: PR Fixer - Claude

Fixed in 2b89499. Added `validation_mode: str = "normal"` to `JobPipelineWorkflowInput` and forwarded it to both the tailor and cover branches of `_execute_stage`. New `test_pipeline_workflow_forwards_validation_mode_to_tailor_and_cover` asserts both stage runners receive the workflow's `validation_mode` so workflow-driven runs reach parity with `jobhunter run --validation lenient`.

workers/automation/src/jobhunter/pipeline/workflow.py
ebarti: PR Fixer - Claude

Fixed in 2b89499. Dropped the `_OUTPUT_TYPES` tuple. The output dataclasses are no longer imported at all — the workflow code only constructs activity-input dataclasses, so the output types were never bound or referenced anywhere. The lone import line is gone with the tuple.

workers/automation/src/jobhunter/materials/__init__.py
ebarti: PR Fixer - Claude

Out of scope for this fixer pass — documented as a residual risk. Moving `jobhunter.materials/` and `jobhunter.profile/` to live under `infrastructure/` (or a new `infrastructure/temporal/activities/` namespace) is a wider package reorganisation that would touch the registry, the imports inside the workflow file, every test that imports from these paths, and any downstream PR that expects today's layout. Flagging in the PR description; it can land as a follow-up if the namespace count starts to bother people.

PR #36, opened 2026-05-07
feat(workers/rpc): cut over apply to Temporal workflow mode (delete fire_and_forget)

ebarti:
## Summary

Replaces the JSON-RPC ``fire_and_forget`` background-thread shortcut with a
``workflow`` dispatch mode. Handlers in ``workflow`` mode return a
``WorkflowStartSpec(workflow, args, workflow_id?, retry_policy?)``; the
``JsonRpcServer`` starts the workflow via an injected ``WorkflowStarter``
(default: ``client.start_workflow(...)`` against ``JOBHUNTER_TASK_QUEUE``)
and returns ``{ runId, workflowId, firstExecutionRunId }``. ``apply`` is
re-registered against ``ApplyWorkflow``. A new ``cancel_run`` JSON-RPC
method gives the existing TS cancel surface a cooperative path against
in-flight workflows.

The threading shortcut and the ``fire_and_forget`` mode constant are
deleted — rip-and-replace per ``MEMORY.md``.

## Stack context

- Stack plan: [`docs/plans/proposed/2026-05-07-temporal-and-worker-reliability-stack.md`](docs/plans/proposed/2026-05-07-temporal-and-worker-reliability-stack.md).
- This is **PR 3 of 7**.
- Base = ``temporal/pipeline-workflow`` (PR 2, #35).
- PR 1 (#34) and PR 2 (#35) are still open. Will rebase to ``main`` once
  both land.

## Scope narrowing

The original plan said ``run_stage`` and ``profile_import`` would also
flip to ``mode=workflow``. They stay ``mode=sync`` in this PR — both
return synchronous result shapes the TS callers parse (e.g.
``defaultProfileImporter`` extracts a profile draft from the
``profile_import`` response). Converting them changes existing sync
contracts and is deferred to its own follow-up. The plan entry has been
updated to reflect this; see the updated [PR 3 entry](docs/plans/proposed/2026-05-07-temporal-and-worker-reliability-stack.md#pr-3--cut-over-json-rpc-to-workflows).

## Behavior change

- ``apply`` JSON-RPC now starts a Temporal workflow instead of a
  background thread. Response shape is
  ``{ runId, workflowId, firstExecutionRunId }``; existing TS callers
  reading ``result.runId`` are unaffected (the comment in
  ``apps/api/src/local-actions.ts`` is updated to reflect the new
  source).
- New ``cancel_run`` method: ``params.runId`` is the workflow id;
  the handler calls ``client.get_workflow_handle(run_id).cancel()``
[...]

PR #34, opened 2026-05-07
feat(workers): add local Temporal foundation (client + worker + doctor)

ebarti:
## Summary

Adds **Local Temporal** as the workflow-engine substrate for the Python worker. Purely additive infrastructure — no pipeline activities, no `JobPipelineWorkflow`, no JSON-RPC change.

- New module `workers/automation/src/jobhunter/infrastructure/temporal/` with three files: `client.py` (async `get_temporal_client()` reading `TEMPORAL_ADDRESS` / `TEMPORAL_NAMESPACE`), `worker.py` (`build_worker(client, *, workflows, activities, task_queue=...)` returning a `temporalio.worker.Worker` bound to `JOBHUNTER_TASK_QUEUE`; registers a private bootstrap no-op workflow when both registries are empty so the worker can boot before pipeline workflows land), and `task_queues.py` (`JOBHUNTER_TASK_QUEUE = 'jobhunter-default'`).
- New CLI command `jobhunter worker` that connects, builds the Worker with empty registries, prints a startup line, and runs forever. `--task-queue` overrides the queue.
- `jobhunter doctor` gains a `Temporal` row that reports `reachable` / `unreachable (start with: temporal server start-dev)`.
- `temporalio>=1.6` added to `workers/automation/pyproject.toml` (resolved to 1.26.0). `pytest-asyncio>=0.23` added to the dev extras with `asyncio_mode = strict` so async tests opt in explicitly.
- Tests: `test_temporal_client.py`, `test_temporal_worker.py`, `test_doctor_temporal.py` cover env handling, queue binding, and both doctor code paths.
- Docs: `docs/local-development.md` gains a "Local Temporal dev server" subsection; `docs/architecture.md` gains a "Workflow Orchestration (Local Temporal)" subsection under Runtime Boundaries; `README.md` Requirements gains the dev-server line; `AGENTS.md` lists `jobhunter worker`.

## Why

Sourced from `docs/backlog.md` -> "Workflow Orchestration (Local Temporal)" and the stack plan at `docs/plans/proposed/2026-05-07-temporal-and-worker-reliability-stack.md`. **This is PR 1 of 7 in that stack** — it lays the worker-bootstrap substrate so the next PR can land pipeline activities + `JobPipelineWorkflow` against a working Worker registry.

## What's NOT in this PR

Deliberately out of scope (each is a future stacked PR):

- **No activities** — the per-context `*/activities.py` modules land in PR 2.
[...]

PR #37, opened 2026-05-07
refactor(workers/apply): collapse apply_runs into Temporal workflow runs

ebarti:
## Summary

PR 4 of 7 in the [Temporal + Worker Reliability stack](https://github.com/ebarti/JobHunter/blob/main/docs/plans/proposed/2026-05-07-temporal-and-worker-reliability-stack.md). Collapses the bespoke `apply_runs` + `apply_run_events` SQLite tables into the Temporal workflow run history. After this PR:

- The `apply_runs` and `apply_run_events` tables are **dropped** on `init_db`.
- `apply_run_projections` is **owned by the Python `ProjectionBuilder`**, sourced directly from `job_events` keyed by `payload.run_id`.
- The TypeScript API reads `apply_run_projections` directly; its `apply_runs → apply_run_projections` projector and `apply_runs` reads are deleted.
- The launcher's queue lock moves to `job_stage_states.apply.state` (the canonical "is this job locked / succeeded / failed" row that `state.set_stage_state` already maintains).
- `SqliteApplyRunRepository` is deleted entirely; the `ApplyRun` aggregate stays in-memory inside `SubmitApplicationUseCase` / `ApplySaga`.

## Stack context

This is **PR 4 of 7**. Base is `temporal/jsonrpc-cutover` (PRs #34, #35, #36 still open ahead of this), not `main`.

## Approach choice — A (event-driven projection)

Picked **Approach A** from the PR plan: derive `apply_run_projections` from `job_events`. Approach B (workflow-history poller) was not needed because:

- The launcher already records every apply lifecycle moment via `record_job_event(stage='apply', payload={"run_id": ..., ...})` — `ApplyRunStarted`, `ApplicationSubmitted`, `ApplicationFailed`, `DryRunCompleted`, `ApplyManualSkip`, `LockReleased`.
- `SubmitApplicationUseCase` independently publishes the same events through `EventPublisher`.
- The `ProjectionBuilder` only needed to learn how to read these events and group them by `run_id`. No new poller, no `client.list_workflows(...)`, no watermark plumbing on top of the existing `operations_projections` watermark.

This keeps the projection materialiser synchronous + in-process, identical to every other `apply_run_projections` neighbour (job-list, dashboard, artifacts, job-detail).

## Behavior change — apply-run history wiped on migration

`init_db` runs `DROP TABLE IF EXISTS apply_runs / apply_run_events` on every worker boot from this PR onward. Existing local apply-run history from the bespoke tables will not survive the migration. Single-user codebase, no production users — accepted per the no-strangler memo (`feedback_no_strangler.md`). New apply runs after this PR materialise via `apply_run_projections` from `job_events`.

## What's NOT in this PR

- The UI Workflow Runs view (PR 5 of the stack — `temporal/runs-view`).
- Worker reliability cleanups: stage-state writer collapse (PR 6) and logs/reports as artifacts (PR 7).

## How to test locally

1. `temporal server start-dev` (already required by the stack).
2. `uv --project workers/automation run jobhunter worker` to boot the Temporal worker.
3. From the UI or `jobhunter apply --url ...`, run a dry-run apply.
4. Verify the apply-run row appears in `apply_run_projections`:
   ```bash
   sqlite3 ~/.jobhunter/jobs.db \
     "SELECT run_id, status, dry_run, started_at, finished_at FROM apply_run_projections ORDER BY started_at DESC LIMIT 1;"
   ```
5. Confirm `apply_runs` table is gone:
   ```bash
   sqlite3 ~/.jobhunter/jobs.db \
     "SELECT name FROM sqlite_master WHERE type='table' AND name LIKE 'apply%';"
   # Should return only: apply_run_projections
   ```

## Verification

| Command | Result |
| --- | --- |
| `uv --project workers/automation run --extra dev pytest -q` | 628 passed |
| `uv --project workers/automation run --extra dev ruff check .` | All checks passed |
| `pnpm api:check && pnpm api:test` | 64 tests passed |
| `pnpm -r check` | All packages clean |
| `grep -rn "apply_runs" workers/automation/src/ apps/api/src/` | only DROP TABLE migration + doc comments + method names like `_rebuild_apply_runs` (which now rebuilds `apply_run_projections`) |

## Test plan

- [x] New tests cover ApplyRunStarted → row open, ApplicationSubmitted → succeeded, ApplicationFailed → failed, DryRunCompleted → dry_run_complete, intermediate-event timeline collected, projection rebuild deterministic.
- [x] `init_db` migration drops both legacy tables (idempotent, smoke-tested via `test_apply_runs_table_dropped.py`).
- [x] TS `apply_run_projections` reads work without legacy `apply_runs` (smoke-tested via `apps/api/test/projections.test.ts`).
- [x] Existing apply queue selectors / launcher regressions pass against the new event-driven projection.


ebarti (commented):
Pre-merge review of PR #37 (apply_runs collapse). Findings inline below.

workers/automation/src/jobhunter/apply/launcher.py
```diff
@@ -460,43 +374,45 @@ def mark_result(
...
-            url=url,
-            run_id=task_id,
-            submission_result=result,
-            duration_ms=duration_ms,
-            dry_run=True,
-            worker_id=run_ctx.get("worker_id") if run_ctx else None,
-            model=run_ctx.get("model") if run_ctx else None,
-        )
         set_stage_state(
             conn,
```
ebarti, line 419: PR Reviewer - Claude

Severity: High
Gate: Fail
Evidence: Reproduced. The dry_run mark_result is broken in the production worker_loop sequence.

1. `acquire_job(target_url=...)` puts `job_stage_states.apply.state = 'running'` (launcher.py acquire_job → set_stage_state).
2. `worker_loop` then calls `mark_result(url, "dry_run", ...)` which executes this `set_stage_state(... "skipped", ...)` block with the default `validate_transition=True`.
3. `_validate_stage_transition` looks up `Running -> Skipped` in the §8.5 table (`_VALID_KIND_TRANSITIONS` in domain/pipeline/state_machine.py) — not present. Raises `ValueError: Invalid state transition for apply: running -> skipped`.

Reproduced with:
```
$ uv --project workers/automation run --extra dev python -c \
  "...acquire_job + mark_result('dry_run')..."
After acquire: running
EXCEPTION: ValueError: Invalid state transition for apply: running -> skipped
Final state: running
```

The existing test `test_dry_run_result_does_not_mark_job_applied` (test_apply_regressions.py:194) does NOT cover the production sequence — it calls mark_result directly without acquire_job, so the row defaults to `pending` (Pending -> Skipped IS valid).

Reproduction: Reproduced with command/steps.

In production the broad `except Exception` in `worker_loop` swallows the ValueError, logs "Launcher error: …", calls `release_lock` (which itself fails — separate finding), increments the failed counter. The dry-run is recorded as a worker failure even though the agent completed successfully, the row stays in `running`, and `DryRunCompleted` is never written to `job_events` so `apply_run_projections` never gets a row for the dry-run.

Required fix: pass `validate_transition=False` to this `set_stage_state(... "skipped", ...)` call. The launcher is the canonical lifecycle owner — admin-style overrides are appropriate here. Add a regression test that calls `acquire_job` then `mark_result("dry_run", ...)` end-to-end so the production sequence is covered.

workers/automation/src/jobhunter/apply/launcher.py
```diff
@@ -553,33 +459,48 @@ def mark_result(
...
+            payload={
+                "run_id": str(run_id),
+                "result": "ORPHANED: lock released by launcher",
+                "finished_at": now,
+                "duration_ms": 0,
+            },
         )
-        repository.save(completed)
     set_stage_state(
         conn,
```
ebarti, line 548: PR Reviewer - Claude

Severity: High
Gate: Fail
Evidence: Reproduced. The new `release_lock` cannot transition the stage row out of `running`.

Sequence:
1. `acquire_job(...)` puts the row in `running` (launcher.py acquire_job).
2. Worker exception path / orphan rescue calls `release_lock(url)`.
3. This `set_stage_state(conn, url, "apply", "pending", ...)` is invoked with the default `validate_transition=True`.
4. `_validate_stage_transition` looks up `Running -> Pending` in the §8.5 table — not present (only `Failed -> Pending`, `Blocked -> Pending`, `Exhausted -> Pending`, `Canceled -> Pending`, `Stale -> Pending` are valid). Raises `ValueError: Invalid state transition for apply: running -> pending`.

Reproduced with:
```
$ uv --project workers/automation run --extra dev python -c "...acquire_job + release_lock..."
After acquire: running
EXCEPTION: ValueError: Invalid state transition for apply: running -> pending (not allowed by the stage state machine)
```

Reproduction: Reproduced with command/steps.

This affects EVERY recovery path on a `running` row:
- `worker_loop` `KeyboardInterrupt` handler.
- `worker_loop` broad `except Exception` handler (including the dry_run failure from the sibling High finding).
- `_rescue_orphaned_running_apply` on worker boot — fails on the first orphan and aborts the whole sweep via the outer `try/except` (launcher.py:1166-1184). Permanent lock leak across worker restarts.

The same shape exists on the base branch (the bug pre-dates PR 4), but PR 4 makes the failure mode strictly worse: the previous noop path at least had `SqliteApplyRunRepository.save(completed)` writing a `Failed("ORPHANED", retryable=True)` row, which gave the read-side a record of the rescue. With `_NoopApplyRunRepository`, no aggregate-side persistence happens, the `record_job_event(... "LockReleased")` after this line never executes, the `conn.commit()` never executes, and the entire transaction aborts.

Required fix: pass `validate_transition=False` to this call. Same fix needed for `reset_failed` (line ~560 — `Failed -> Pending` IS valid for the failed/exhausted case it targets, but the `_has_succeeded_apply` early-skip already filters succeeded rows, so only failed/exhausted reach the call — that one is OK). The launcher's orphan sweep loop in `_rescue_orphaned_running_apply` should also iterate-and-continue on per-row exceptions so one stuck row doesn't poison the entire rescue pass.

workers/automation/src/jobhunter/apply/launcher.py
```diff
@@ -788,13 +673,44 @@ def _build_use_case() -> SubmitApplicationUseCase:
     )
 
 
-def _result_to_status_string(result: SubmissionResult) -> str:
-    """Map a ``SubmissionResult`` variant back to the legacy status string.
+class _NoopApplyRunRepository:
+    """No-op stand-in for the deleted ``SqliteApplyRunRepository``.
```
ebarti, line 736: PR Reviewer - Claude

Severity: High
Gate: Fail
Evidence: The implementer's claim "intermediate-event timeline collected" is not what the saga delivers in production.

`ApplySaga.run` (domain/apply/process_manager.py) records six rich saga events on the in-memory aggregate at every step: `SagaStarted`, `BrowserLaunched`, `BrowserLaunchFailed`, `AgentStarted`, `AgentTimedOut`, `AgentCrashed`, `AgentResult`, plus per-`AgentEvent` entries from `AgentResult.events`. After each step the saga calls `repository.save(run)`.

With this `_NoopApplyRunRepository.save` returning `None`, none of those events reach SQLite. The `SubmitApplicationUseCase._publish_event_records(...)` fan-out path is the only fallback — but `_build_use_case` (lines above) constructs `SubmitApplicationUseCase` WITHOUT `publisher=...`, so `self._publisher is None` and `_publish_event_records` short-circuits without publishing. Net result: the saga's `SagaStarted` / `BrowserLaunched` / `AgentStarted` / `AgentTimedOut` / `AgentCrashed` / `AgentResult` / per-AgentEvent timeline is **lost entirely** in production.

`apply_run_projections.events_json` only contains what `launcher.acquire_job` / `launcher.mark_result` write directly via `record_job_event` (i.e. `ApplyRunStarted`, `ApplicationSubmitted` | `ApplicationFailed` | `DryRunCompleted`, `LockReleased`, `ApplyManualSkip`). That's a 4–5x reduction in dashboard timeline detail vs the base branch where `SqliteApplyRunRepository.save` persisted aggregate events into `apply_run_events`.

Reproduction: Reasoned from code (verified call graph: `_build_use_case` → no `publisher` kwarg; `SubmitApplicationUseCase.__init__` defaults `publisher=None`; `_publish_event_records` → early return when None; `ApplySaga.run` calls `record_event` → `repository.save` → `_NoopApplyRunRepository.save` → no-op).

The new test `test_event_timeline_collected_per_run` in `test_apply_runs_projection_from_events.py:213` does NOT detect this regression: it manually seeds `ApplyRunEvent` via `record_job_event(... payload={"run_id": ...})`, never exercising the `ApplySaga` path.

Required fix (pick one):
1. Wire the `get_default_publisher()` into `_build_use_case` — `SubmitApplicationUseCase(publisher=get_default_publisher(), ...)` — and have `_publish_event_records` write the aggregate event timeline into `job_events` (it already publishes via the bus, but `record_job_event` requires a separate side-effect to land the row).
2. OR, persist saga events directly into `job_events` from `ApplySaga.run` (write through `record_job_event` keyed by `run.run_id`).
3. OR, restore the `SqliteApplyRunRepository`-shaped persistence: have a thin "ApplyRunEventRepository" that turns `apply_run.events` into `job_events` rows under the same run_id at terminal time, so the per-run timeline survives the noop transition.

Add a regression test that runs `SubmitApplicationUseCase.execute(...)` with the launcher's actual `_NoopApplyRunRepository` + the production `_build_use_case` wiring and asserts that `apply_run_projections.events_json` contains at least the `BrowserLaunched` / `AgentStarted` / `AgentResult` markers.

apps/api/src/projections.ts
```diff
@@ -405,7 +411,11 @@ export function refreshProjections(db: SqliteDatabase, tenantId = "local"): void
     rebuildJobProjections(db, tenantId, jobUrl);
   }
   if (dirtyJobs.size > 0) {
-    rebuildApplyRunProjections(db, tenantId);
+    // PR 4 of the Temporal stack: ``apply_run_projections`` is now
```
ebarti, line 414: PR Reviewer - Claude

Severity: High
Gate: Fail
Evidence: This change removes the only synchronous materialiser of `apply_run_projections` that the TS read path could rely on. With it gone, the table is materialised exclusively by the Python `ProjectionBuilder._rebuild_apply_runs`, which only runs when:

1. The CLI bootstraps (`cli.py:_bootstrap` → `builder.refresh()`), once per process.
2. Something publishes to the in-process bus → wildcard subscriber `_on_event` → `refresh()`.

The launcher is the producer of every `ApplyRunStarted` / `ApplicationSubmitted` / `ApplicationFailed` / `DryRunCompleted` row, but it writes through `record_job_event(conn, ..., publisher=None)` and the `_build_use_case` constructor passes no publisher to `SubmitApplicationUseCase`. So nothing in the apply path publishes — a long-lived `jobhunter worker` process can run dozens of apply lifecycles without ever refreshing `apply_run_projections`.

Concrete effect on the user-facing TS API:
- `loadLatestApplyRun` (this file, ~line 575) reads `apply_run_projections` directly and returns stale rows for every apply run since the last bootstrap.
- `rebuildDashboardProjection` (this file, dryRuns counter) reads `apply_run_projections` directly with the same staleness.
- `_LATEST_APPLY_RUN_JOIN` in Python `database.py` joins through the same stale table — so `get_jobs_by_stage("pending_apply")` / `get_jobs_by_stage("applied")` / `get_stats()["applied"]` / `get_stats()["apply_errors"]` see stale data too.

Reproduction: Reasoned from code + verified via `grep -rn "publisher" workers/automation/src/jobhunter/apply/` — no publisher path exists from the launcher to the bus.

Base-branch behaviour: the previous TS `rebuildApplyRunProjections(db, tenantId)` (the line being deleted here) read directly from `apply_runs` and re-materialised on every TS API request via the existing `refreshProjections(...)` call at the top of every read-model query. So the TS API was always self-fresh. PR 4 silently moves this responsibility to a Python writer that has no trigger.

Required fix (pick one):
1. Wire `get_default_publisher()` into the launcher's `_build_use_case` so saga events fan out through the bus and trigger the subscribed `_on_event` → `refresh()`. Also make the launcher's direct `record_job_event` calls publish (`record_job_event(... publisher=get_default_publisher())`) so `ApplyRunStarted` / `ApplicationSubmitted` / `ApplicationFailed` / `DryRunCompleted` themselves trigger refresh.
2. OR, restore the TS-side rebuild but key it off `job_events` (mirror the Python `_rebuild_apply_runs` logic) so the TS API stays self-fresh independent of the Python worker process.
3. OR, add a Python-side trigger (e.g. an after-write hook in `record_job_event` that calls `_rebuild_apply_runs` synchronously when stage='apply') so writes never outpace materialisation.

Add a regression test that boots a worker process, runs `apply.launcher.main(limit=1, ...)` against a stub agent that returns `Applied`, then queries the TS API and asserts `loadLatestApplyRun` returns the freshly-completed run.

workers/automation/src/jobhunter/apply/launcher.py
```diff
@@ -553,33 +459,48 @@ def mark_result(
...
     now = _utc_now()
-    for active in repository.list_active(LOCAL_TENANT):
-        if str(active.job_id) != url:
-            continue
-        completed = active.complete(
-            result=Failed(error="ORPHANED: lock released by launcher", retryable=True),
-            finished_at=now,
-            duration_ms=0,
+    run_id = (run_ctx.get("run_id") if run_ctx else None) or new_apply_run_id()
+    if _has_active_apply(conn, url):
```
ebarti, line 530: PR Reviewer - Claude

Severity: Medium
Gate: Pass
Evidence: This `release_lock` mints a NEW `run_id` via `new_apply_run_id()` whenever the caller doesn't pass one through `run_ctx`. The orphan rescue path in `_rescue_orphaned_running_apply` (launcher.py:1166-1184) calls `release_lock(row["job_url"])` without `run_ctx`, so on every orphan rescue:

1. A brand-new uuid `run_id` is generated.
2. An `ApplicationFailed` event is written with `payload.run_id = <new uuid>` and `started_at` absent.
3. The projection's `_collect_apply_events_by_run` groups by payload `run_id`, so the orphaned row's ORIGINAL `ApplyRunStarted` (with the original `run_id`) and this new `ApplicationFailed` (with the new `run_id`) are treated as TWO separate runs.
4. The original run stays in `apply_run_projections` with `status='starting'` forever — it never receives a terminal event.
5. The phantom new run appears with `status='failed'`, `started_at=NULL`, `finished_at=now`.

Reproduction: Reasoned from code (verified: `_rescue_orphaned_running_apply` does not look up the original `run_id` from the latest `ApplyRunStarted` event for the URL before calling `release_lock`).

Same drift happens in `worker_loop` exception/KeyboardInterrupt handlers when `run_ctx['run_id']` differs from whatever event-side run_id the launcher already wrote — but those paths normally do pass `run_ctx`, so this is mostly an orphan-rescue concern.

Required fix: in the orphan-rescue path, look up the latest `ApplyRunStarted` event for the URL and reuse its `payload.run_id` so the terminal event closes the SAME run row. Something like:
```python
row = conn.execute(
    "SELECT payload_json FROM job_events "
    "WHERE job_url = ? AND stage = 'apply' AND event_type = 'ApplyRunStarted' "
    "ORDER BY event_id DESC LIMIT 1", (url,)
).fetchone()
prior_run_id = (json.loads(row["payload_json"]) or {}).get("run_id") if row else None
release_lock(url, run_ctx={"run_id": prior_run_id} if prior_run_id else None)
```

Also worth noting: `_TERMINAL_EVENT_STATUS` in the projection includes `"LockReleased": ("failed", "failed")` but the launcher emits `LockReleased` without a `run_id` payload, so it's filtered out by `_collect_apply_events_by_run`. That dead mapping is misleading — either drop it from the table, or have `release_lock` emit `LockReleased` with the (looked-up) `run_id` so it actually drives the projection.

workers/automation/tests/test_apply_regressions.py
```diff
@@ -1,10 +1,20 @@
...
-import time
+"""PR 4 regressions for ``jobhunter.apply.launcher``.
+
+The bespoke ``apply_runs`` table is gone; the canonical lock now lives
+on ``job_stage_states.apply.state`` and the lifecycle is observable via
+the ``apply_run_projections`` table (sourced from ``job_events``).
+These tests cover the launcher contract that downstream callers
+(``cli.py``, ``actions.py``, ``pipeline.py``) still rely on.
+"""
+
```
ebarti, line 9: PR Reviewer - Claude

Severity: Medium
Gate: Pass
Evidence: The implementer's task brief specifically calls out: "Concurrency / lock semantics on the new `job_stage_states.apply` lock have a test (race-style)." The new `test_apply_regressions.py` does not include a race-style test.

Why this matters: PR 4 moves the lock primitive from `apply_runs.status IN ('starting','in_progress')` (a per-run row INSERTed inside `BEGIN IMMEDIATE`) to `job_stage_states.apply.state == 'running'` (a per-job-stage row UPSERTed inside `BEGIN IMMEDIATE`). The semantics LOOK equivalent — both rely on `BEGIN IMMEDIATE` + the SQLite write-lock — but verifying that two concurrent `acquire_job(url)` calls do not BOTH succeed is the critical regression guard for the migration.

Reproduction: Reasoned from code (verified `BEGIN IMMEDIATE` + `PRAGMA busy_timeout=10000` from `database.py:51`; both make the lock sound IN PRINCIPLE, but only an actual race test proves it).

Existing tests cover the SQL shape (`test_targeted_apply_takes_canonical_stage_lock`) but not the concurrency invariant.

Required fix: add a test that spawns two threads (each with its own `get_connection(db_path)` since `_local` is thread-local), each calling `acquire_job(target_url=URL)`, and asserts that exactly one returns a job dict and the other returns `None`. The test should also poll the `job_stage_states.apply` row to assert exactly one `running` row exists at the end.

Suggested skeleton:
```python
def test_concurrent_acquire_is_serialised(tmp_path, monkeypatch):
    db_path = Path(tmp_path) / "jobs.db"
    init_db(db_path)
    # ... insert one ready job ...
    monkeypatch.setattr("jobhunter.apply.launcher.get_connection", lambda: get_connection(db_path))

    results = []
    def _acquire():
        results.append(acquire_job(target_url=URL, worker_id=...))

    t1 = threading.Thread(target=_acquire)
    t2 = threading.Thread(target=_acquire)
    t1.start(); t2.start(); t1.join(); t2.join()

    successes = [r for r in results if r is not None]
    assert len(successes) == 1
```

workers/automation/src/jobhunter/infrastructure/projections/projection_builder.py
```diff
@@ -637,7 +648,7 @@ def _rebuild_dashboard(self) -> None:
         try:
             dry_runs = int(
                 self._conn.execute(
-                    "SELECT COUNT(*) FROM apply_runs WHERE dry_run = 1"
+                    "SELECT COUNT(*) FROM apply_run_projections WHERE dry_run = 1"
                 ).fetchone()[0]
```
ebarti, line 769: PR Reviewer - Claude

Severity: Low
Gate: Pass
Evidence: This `dry_runs` counter switches the FROM table from `apply_runs` to `apply_run_projections`, but does not match the TS-side counter in `apps/api/src/projections.ts:rebuildDashboardProjection` which joins `apply_run_projections arp LEFT JOIN jobhunter_deleted_jobs d ON d.job_url = arp.job_id AND d.restored_at IS NULL WHERE arp.dry_run = 1 AND d.job_url IS NULL` (i.e. excludes dry runs whose underlying job was soft-deleted).

The Python dashboard counts ALL `dry_run = 1` rows including those for soft-deleted jobs. Both the Python and TS refreshers write the same `dashboard_projections` row, so the user-visible value is whichever one ran most recently.

The same shape existed in the base branch (Python's old `apply_runs WHERE dry_run = 1` also lacked the deleted-jobs join), so this is a pre-existing divergence carried forward by PR 4 — but PR 4 is a natural moment to align them.

Reproduction: Reasoned from code (compared the two counters side-by-side).

Required fix (cosmetic, non-blocking): mirror the TS join here:
```sql
SELECT COUNT(*) FROM apply_run_projections arp
LEFT JOIN jobhunter_deleted_jobs d ON d.job_url = arp.job_id AND d.restored_at IS NULL
WHERE arp.dry_run = 1 AND d.job_url IS NULL
```

workers/automation/src/jobhunter/apply/launcher.py
ebarti: PR Fixer - Claude

Fixed in 7c9547e. ``mark_result(... "dry_run", ...)`` now passes ``validate_transition=False`` to the ``set_stage_state(... "skipped", ...)`` call with a one-line note that this is a launcher convention (Running -> Skipped is not in the canonical §8.5 table; the launcher owns the dry-run policy).

Regression test in 0c01118: ``test_acquire_job_then_mark_result_dry_run_completes_end_to_end`` exercises the full production sequence (``acquire_job`` -> Running, ``mark_result("dry_run", ...)`` -> Skipped, projection refresh -> ``apply_run_projections`` row with ``dry_run = 1`` and ``status = "dry_run_complete"``).

End-to-end reproducer (the exact sequence the reviewer ran) now completes without exception:
```
After acquire: running
After mark_result(dry_run): skipped
PASS
```

workers/automation/src/jobhunter/apply/launcher.py
ebarti: PR Fixer - Claude

Fixed in 7c9547e. ``release_lock`` now passes ``validate_transition=False`` to the ``set_stage_state(... "pending", ...)`` call with a one-line note (launcher owns the lock-release convention; Running -> Pending is not in the §8.5 table).

The orphan sweep robustness was also fixed in the same commit: ``_rescue_orphaned_running_apply`` now wraps each per-row ``release_lock`` in ``try / except Exception: log + continue``, so one corrupt payload (or any other per-row failure) no longer aborts the whole sweep.

Regression tests in 0c01118:
- ``test_release_lock_releases_running_row_back_to_pending`` covers the Running -> Pending transition end-to-end.
- ``test_orphan_rescue_continues_past_failing_row`` injects a per-row failure and asserts the sweep keeps going.

End-to-end reproducer:
```
After acquire: running
After release_lock: pending
PASS
```

workers/automation/src/jobhunter/apply/launcher.py
ebarti: PR Fixer - Claude

Fixed in 8510043. Two changes:

1. ``_build_use_case`` now wires ``get_default_publisher()`` from ``jobhunter.infrastructure.events`` into ``SubmitApplicationUseCase``, so the use case's ``ApplyRunStarted`` / ``ApplyRunEventRecorded`` / ``ApplicationSubmitted`` / ``ApplicationFailed`` events fan out to the in-process bus.

2. The launcher's ``run_job`` now persists the saga's intermediate event timeline (``SagaStarted`` / ``BrowserLaunched`` / ``BrowserLaunchFailed`` / ``AgentStarted`` / ``AgentTimedOut`` / ``AgentCrashed`` / ``AgentResult`` / per-``AgentEvent``) into ``job_events`` keyed by ``run_id`` via the new ``_persist_saga_event_timeline`` helper. Terminal lifecycle events are intentionally skipped to avoid duplicating what ``acquire_job`` / ``mark_result`` / ``release_lock`` already write.

Why we didn't go full "publisher-only fan-out": ``apply_run_projections`` is built by ``_collect_apply_events_by_run``, which re-reads from ``job_events``. Publishing to the bus alone wouldn't land the rows. Persisting the saga timeline directly into ``job_events`` is the smallest change that makes the projection see the full timeline.

Regression test in 0c01118: ``test_apply_saga_writes_full_event_timeline_to_job_events`` stubs out Chrome / Claude Code, drives an ``ApplyRun`` aggregate through the saga event shape, calls ``_persist_saga_event_timeline`` + ``mark_result("applied", ...)``, and asserts ``job_events`` carries SagaStarted / BrowserLaunched / AgentStarted / AgentResult (each with the same ``payload.run_id``) AND that ``apply_run_projections.events_json`` includes those event types.

apps/api/src/projections.ts
ebarti: PR Fixer - Claude

Fixed in 8510043. ``record_job_event`` in ``workers/automation/src/jobhunter/state.py`` now defaults to ``get_default_publisher()`` when no publisher is supplied (lazy import to avoid bootstrap cycles). Every ``record_job_event`` call therefore fans out through the process-wide ``InProcessEventBus``, and ``ProjectionBuilder._on_event`` (already wired in ``cli.py:_bootstrap``) refreshes the read-model after every write.

The launcher's ``acquire_job`` / ``mark_result`` / ``release_lock`` paths -- the producers of ``ApplyRunStarted`` / ``ApplicationSubmitted`` / ``ApplicationFailed`` / ``DryRunCompleted`` / ``LockReleased`` -- now automatically trigger projection refresh without changing their call sites.

Combined with the High #3 fix (saga timeline persisted to ``job_events`` from ``run_job``), the Python read-model stays self-fresh through long-lived ``jobhunter worker`` processes, so the TS API's ``loadLatestApplyRun`` / ``rebuildDashboardProjection`` / ``_LATEST_APPLY_RUN_JOIN`` callers see fresh rows. The TS-side rebuilder stays gone -- the no-strangler / single-write-path rule is preserved.

Regression test in 0c01118: ``test_record_job_event_default_publisher_refreshes_apply_run_projections`` proves the wildcard subscriber fires on every event and ``apply_run_projections`` reflects the just-written events after one ``ProjectionBuilder.refresh()``.

workers/automation/src/jobhunter/apply/launcher.py
ebarti: PR Fixer - Claude

Fixed in 7c9547e. ``release_lock`` now looks up the most recent ``ApplyRunStarted`` event for the URL via the new ``_latest_apply_run_started_run_id`` helper and reuses its ``payload.run_id`` when no ``run_ctx`` is provided. The terminal ``ApplicationFailed`` event is now keyed under the SAME ``run_id`` as the original ``ApplyRunStarted``, so the projection closes the same row instead of stranding the original in ``status='starting'``.

The ``LockReleased`` event now also carries ``payload.run_id`` so the dead ``_TERMINAL_EVENT_STATUS["LockReleased"]: ("failed", "failed")`` mapping in the projection actually fires.

Regression test in 0c01118: ``test_orphan_rescue_keeps_original_run_id`` runs ``acquire_job`` (which writes ``ApplyRunStarted`` with the canonical run_id), then ``release_lock(url)`` (no ctx -- the orphan-rescue shape), and asserts exactly one ``apply_run_projections`` row with the ORIGINAL ``run_id`` and ``status='failed'``. Verified end-to-end:
```
After acquire: running
Original run_id: f4e94281360b43dcbfbbbc9015d8ab29
After release_lock(no ctx): pending
Number of apply_run_projections rows: 1
  Row: run_id=f4e94281360b43dcbfbbbc9015d8ab29 status=failed
```

workers/automation/tests/test_apply_regressions.py
ebarti: PR Fixer - Claude

Added in 0c01118. ``test_acquire_job_concurrent_workers_only_one_succeeds`` boots two threads (each with its own thread-local SQLite connection -- ``get_connection`` is already thread-local, so the monkeypatch points each thread to the same DB path) racing on ``acquire_job(target_url=URL)``. Asserts exactly one returns a job dict and the other returns ``None``, and verifies exactly one ``running`` row in ``job_stage_states`` for the URL. Test passes deterministically -- the new ``BEGIN IMMEDIATE`` + ``UPSERT`` semantics serialise the workers as expected.

workers/automation/src/jobhunter/infrastructure/projections/projection_builder.py
ebarti: PR Fixer - Claude

Fixed in 294d463. The Python ``_rebuild_dashboard`` ``dry_runs`` counter now mirrors the TS join in ``apps/api/src/projections.ts``:

```sql
SELECT COUNT(*)
FROM apply_run_projections arp
LEFT JOIN jobhunter_deleted_jobs d
  ON d.job_url = arp.job_id AND d.restored_at IS NULL
WHERE arp.dry_run = 1 AND d.job_url IS NULL
```

Regression test in 0c01118: ``test_dashboard_dry_runs_excludes_soft_deleted_jobs`` seeds two jobs with dry-run lifecycles, soft-deletes one via ``jobhunter_deleted_jobs``, refreshes the projection, and asserts ``dashboard_projections.dry_runs == 1``.

apps/api/src/projections.ts
ebarti: PR Reviewer - Claude

Severity: High
Gate: Fail
Evidence: The fix is incomplete in multi-worker mode. Reproduced.

Subscribing the projection builder via ``cli.py:74: builder.subscribe_to(get_default_publisher())`` works in single-worker mode (default ``--workers 1``), but it silently breaks in multi-worker mode (``apply --workers > 1`` from ``launcher.py:1250-1266`` ``ThreadPoolExecutor``):

  1. ``_bootstrap`` runs in the main thread, so ``ProjectionBuilder._conn = init_db()`` is the **main-thread**-cached SQLite connection.
  2. Worker threads call ``acquire_job`` -> ``record_job_event`` (which now defaults to publishing through the bus).
  3. The wildcard subscriber ``ProjectionBuilder._on_event`` runs synchronously on the worker thread.
  4. ``self._conn.execute(...)`` from ``refresh()`` uses the main-thread connection from the worker thread -> ``sqlite3.ProgrammingError: SQLite objects created in a thread can only be used in that same thread``.
  5. The exception is caught by ``_on_event`` (``except Exception``) and logged as ``"ProjectionBuilder failed to refresh after %s"`` -- so the call returns success but the projection never refreshes.

Reproducer (run inside the worktree):

```
uv --project workers/automation run --extra dev python - <<EOF
import os, tempfile, threading
from pathlib import Path
db = Path(tempfile.mkdtemp()) / "jobs.db"
os.environ["JOBHUNTER_DB_PATH"] = str(db)
from jobhunter.database import init_db, get_connection
from jobhunter.infrastructure.projections.projection_builder import ProjectionBuilder
from jobhunter.infrastructure.events import get_default_publisher, reset_default_publisher
from jobhunter.state import record_job_event
reset_default_publisher()
conn_main = init_db(db)  # bootstrap thread connection
conn_main.execute("INSERT INTO jobs (url, title, site, fit_score, tailored_resume_path, application_url) VALUES (?,?,?,?,?,?)",
    ("https://example.com/job","Eng","ExampleCo",9,"/tmp/r.txt","https://example.com/apply"))
conn_main.commit()
ProjectionBuilder(conn_main).subscribe_to(get_default_publisher())
def worker():
    record_job_event(get_connection(db), "https://example.com/job", "apply", "ApplyRunStarted",
        payload={"run_id": "from-worker", "started_at": "2026-05-04T13:00:00+00:00"})
t = threading.Thread(target=worker); t.start(); t.join()
EOF
```

Output: ``ProjectionBuilder failed to refresh after ApplyRunStarted ... sqlite3.ProgrammingError``.

The default ``acquire_job`` happens to commit the row to ``job_events`` BEFORE ``publisher.publish`` fires (so the data is durable), and the next ``_bootstrap`` invocation will pick it up via the watermark backfill -- so this isnt data loss. But the briefs claim "the wildcard subscriber actually fires the projection refresh" only holds for ``--workers 1``. The new test ``test_record_job_event_default_publisher_refreshes_apply_run_projections`` runs everything on the main thread and therefore does not catch this regression.

Reproduction: Reproduced with command/steps above on commit 0c01118.

Required fix: pick one --
  (a) ``ProjectionBuilder._on_event`` opens its own thread-local connection via ``get_connection(db_path)`` (the same path the publisher would use) instead of reusing ``self._conn``; OR
  (b) detect the cross-thread case and skip / queue the refresh; OR
  (c) document explicitly that the in-process bus only fans out projection refreshes for single-worker mode and accept the watermark backfill on the next bootstrap as the multi-worker contract; AND extend the new test to actually spawn a worker thread that calls ``record_job_event`` and asserts the projection updates.

Until then, multi-worker apply runs leave the dashboard / read-model stale until the next CLI invocation -- a regression in the High #4 stated behaviour.

workers/automation/src/jobhunter/apply/launcher.py
ebarti: PR Reviewer - Claude

Severity: Medium
Gate: Pass
Evidence: ``_persist_saga_event_timeline`` only persists event types in the ``_SAGA_EVENT_TYPES_PERSISTED`` safelist (``{"SagaStarted", "BrowserLaunched", "BrowserLaunchFailed", "AgentStarted", "AgentTimedOut", "AgentCrashed", "AgentResult", "AgentEvent"}``).

The Claude Code adapter (``workers/automation/src/jobhunter/infrastructure/apply/claude_code_cli.py:170,232,248``) emits events with ``event_type`` values like ``"ClaudeLaunched"``, ``"AssistantText"``, ``"ToolUse"`` -- none of which are in the safelist. The saga then folds them straight into the aggregate timeline (``process_manager.py:271-278: event_type=str(raw_event.get("event_type", "AgentEvent"))``), preserving the raw name.

Result: the per-step "agent did X / used tool Y / wrote text Z" telemetry that previously landed in ``apply_run_events`` (now-dropped table -- see ``telemetry.py`` on the base) is **silently dropped** in the new path. The new test ``test_apply_saga_writes_full_event_timeline_to_job_events`` only asserts the saga-internal wrappers (``SagaStarted`` / ``BrowserLaunched`` / ``AgentStarted`` / ``AgentResult``) -- never ``ToolUse`` or ``AssistantText`` -- so it passes.

This is currently invisible because no read path renders the agent-stream timeline today (``apps/api/src/read-model.ts:recentApplyRuns`` does not project ``events_json``, and the web does not read it either), so user-facing severity is Medium not High. But:

  1. The briefs expectation -- "the saga events show up with the correct ``run_id``" -- doesnt hold for the agent-stream subset, which used to be the most useful slice.
  2. Anything built on top of ``apply_run_projections.events_json`` (its in the schema, just unused) will see an empty / sparse timeline.

Reproduction: Reasoned from code (``_SAGA_EVENT_TYPES_PERSISTED`` at ``launcher.py:761`` vs. ``claude_code_cli.py:172,234,250``).

Required fix: either (a) drop the safelist and persist any ``event_type`` that originated from the saga / agent (filter only the duplicate-with-launcher terminal types: ``ApplicationSubmitted`` / ``ApplicationFailed`` / ``DryRunCompleted`` / ``ApplyRunStarted``), or (b) extend ``_SAGA_EVENT_TYPES_PERSISTED`` to include ``ClaudeLaunched`` / ``AssistantText`` / ``ToolUse`` (and document the contract that any new agent-emitted ``event_type`` must be added). Then extend the test to construct an agent stream with a ``"ToolUse"`` event and assert it lands in ``job_events`` and ``events_json``.

apps/api/src/projections.ts
ebarti: PR Fixer - Claude

Fixed in 39602d4. ``ProjectionBuilder`` now takes a
``conn_factory: Callable[[], sqlite3.Connection]`` instead of a
fixed connection. ``refresh()``/``_on_event`` pull a per-call
connection from the factory and bind it (plus a per-call
``SqliteProjectionStore`` and ``SqliteEventWatermarkRepository``)
into ``threading.local`` for the duration of that refresh, so the
wildcard subscriber runs on whichever thread published the event
without ever touching another thread's SQLite handle.

Production wires ``conn_factory=get_connection`` from
``cli._bootstrap`` -- ``get_connection`` is itself thread-local
cached, so every worker thread gets its own conn for free.

The blanket ``except`` in ``_on_event`` stays (so a buggy projector
cannot break the bus), but the log call uses ``log.exception`` (=
``log.error(..., exc_info=True)``) so the traceback is never silent
again -- I left an explicit comment to that effect.

About the ``contextlib.closing(self._conn_factory())`` suggestion:
I dropped the ``close()`` from ``_on_event`` after watching the
worker reproducer hit ``ProgrammingError("Cannot operate on a
closed database.")``. In production the factory is
``get_connection(db_path)``, which returns the **same** thread-local
cached handle the publishing caller is currently writing to;
closing it from the wildcard subscriber yanked the conn out from
under the writer (and any subsequent ``record_job_event`` on the
same thread). The factory now owns connection lifetime --
``get_connection`` self-heals on next call if a stale handle is
ever cached -- and the builder never closes. Documented in
``_on_event``'s docstring.

New regression test
``test_record_job_event_from_worker_thread_refreshes_projection``
boots the subscription on the bootstrap thread, spawns a worker
thread that emits ``ApplyRunStarted`` + ``ApplicationSubmitted``
via ``record_job_event``, polls ``apply_run_projections`` for up to
1s, and asserts the row lands with ``status='succeeded'``. Verified
fail-then-pass: against the prior fixed-connection ``__init__`` the
poll times out (``ProgrammingError`` swallowed in the log); after
the fix it lands within milliseconds.

All call sites (CLI bootstrap + every existing test) now pass
``conn_factory=...``; the old positional ``conn`` ctor is gone
(no-strangler).

Verification (inside the worktree):
``uv --project workers/automation run --extra dev pytest -q`` -> 637 passed
``uv --project workers/automation run --extra dev ruff check .`` -> All checks passed
``pnpm api:check && pnpm api:test && pnpm -r check`` -> all pass
``git diff --check`` -> clean

workers/automation/src/jobhunter/apply/launcher.py
ebarti: PR Fixer - Claude

Fixed in e384e79. Dropped ``_SAGA_EVENT_TYPES_PERSISTED`` and
inverted to a small ``_LAUNCHER_OWNED_EVENT_TYPES`` blocklist
(``ApplyRunStarted`` / ``ApplicationSubmitted`` /
``ApplicationFailed`` / ``DryRunCompleted`` / ``ApplyManualSkip`` /
``LockReleased``) -- exactly the events the launcher already wrote
directly via ``acquire_job`` / ``mark_result`` / ``release_lock``,
which would otherwise duplicate. Set verified by grepping
``record_job_event(`` in ``launcher.py`` and listing each event type.
Every other saga-buffer event (``SagaStarted`` / ``BrowserLaunched``
/ ``AgentStarted`` / ``AgentResult`` / ``AgentTimedOut`` /
``AgentCrashed`` plus the agent-stream ``ClaudeLaunched`` /
``AssistantText`` / ``ToolUse`` and any new types the adapter adds)
now flows into ``job_events``.

One-line comment above the new constant documents the inversion.

Extended ``test_apply_saga_writes_full_event_timeline_to_job_events``
to emit ``ClaudeLaunched`` + ``AssistantText`` + ``ToolUse`` from
the saga timeline. The test now asserts each lands in
``job_events`` keyed by the same ``run_id`` AND in
``apply_run_projections.events_json``, and additionally that the
``ToolUse`` payload (``name`` + ``input``) survives the round-trip
unchanged.

PR #39, opened 2026-05-07
refactor(workers): collapse dual stage-state write path through canonical helper

ebarti:
## Summary

Eliminates the second runtime write path into `job_stage_states`.
`SqlitePipelineStateRepository.save` no longer emits its own bespoke
`UPDATE/INSERT INTO job_stage_states` SQL — it now routes every stage
through `state.set_stage_state` so the canonical helper is the single
source of truth for stage-state writes (validation, COALESCE semantics,
event-bus fan-out, and now optimistic locking too).

After this change `grep -rn "INSERT INTO job_stage_states\|UPDATE
job_stage_states" workers/automation/src/jobhunter/` returns exactly one
hit — the canonical helper's UPSERT.

## Stack context

PR 6 of 7 in the [Temporal + Worker Reliability stack](../blob/main/docs/plans/proposed/2026-05-07-temporal-and-worker-reliability-stack.md).
Sits on top of `temporal/runs-view` (PR 5).  PR 7 (logs/reports as
artifacts) builds on this branch next.

## Approach choice

**Extension of `set_stage_state` with `expected_version: int | None`.**
The canonical helper grows one optional keyword: when supplied, the upsert
is guarded by the row's current `version` column and the new row is
written at `version = expected_version + 1`; on mismatch
`OptimisticLockError` is raised.  When `None` (the default), the existing
unguarded UPSERT path is preserved exactly for stage runners that do not
need versioning.

I picked this over the "extract a private `_write_stage_row` helper"
alternative because the diff is smaller (no new abstraction layer), the
SQL stays in one function, and every existing caller of
`set_stage_state` keeps the same signature.  The repository becomes a
thin loop over `state.stages.items()` — `_stage_state_to_kwargs` plus the
helper plus a per-stage event emission.

## Behavior preserved

- Optimistic locking: the same `OptimisticLockError(job_url,
  expected_version, actual_version)` is raised on version conflict.  The
  pre-existing `test_optimistic_lock_conflict` regression still holds
  unchanged.
- Per-row UPSERT semantics (COALESCE on partial updates) unchanged.
- `state.version` still bumps to `state.version + 1` after a successful
  save().
- `save()` now emits one `record_job_event` per stage that actually
  changed state — type/level mapped from the StageState kind
  (`StageStarted`, `StageCompleted`, `StageFailed`, `StageBlocked`,
  `StageSkipped`, `StageExhausted`, `StageCanceled`, `StageReset`,
  `StageStale`).  Idempotent saves emit nothing.

## What's NOT in this PR

- Registering log/report files as artifacts via `record_job_artifact` —
  that is PR 7 (`worker-reliability/logs-as-artifacts`).
- Any change to the per-stage runners (they still pair `set_stage_state`
  with their own `record_job_event` calls — no double-events because the
  repo's emission only fires when called from the new aggregate path).
- Any change to `database.py`'s `_backfill_jobs_to_stage_states` — that
  is a one-shot init migration, not a runtime write path.

## Verification

Run inside the worktree (`/Users/eloibarti/Github/JobHunter-worktrees/wr-single-stage-writer`):

| Command | Result |
| --- | --- |
| `uv --project workers/automation run --extra dev pytest -q` | 649 passed |
| `uv --project workers/automation run --extra dev ruff check .` | All checks passed |
| `pnpm api:check && pnpm api:test` | tsc clean, 66 vitest passed |
| `pnpm -r check` | 6/6 workspace projects clean |
| `grep -rn "INSERT INTO job_stage_states\|UPDATE job_stage_states" workers/automation/src/jobhunter/` | 1 hit (`state.py:226` — the canonical helper) |

### DIFF
--- a/workers/automation/src/jobhunter/infrastructure/pipeline/sqlite_repository.py
+++ b/workers/automation/src/jobhunter/infrastructure/pipeline/sqlite_repository.py
@@ -1,18 +1,20 @@
 """SqlitePipelineStateRepository — adapter for persisting JobPipelineState in SQLite.
 
 Maps domain StageState PascalCase variants to/from the lowercase serialized
-strings stored in the ``job_stage_states`` table.  Implements optimistic
-locking via the ``version`` column (ddd-target.md S8.6).
+strings stored in the ``job_stage_states`` table.  Persistence is delegated
+to :func:`jobhunter.state.set_stage_state` so all stage-state writes share
+the same SQL, validation, and per-row event emission.  Optimistic locking
+uses the ``version`` column (ddd-target.md S8.6) via the helper's
+``expected_version`` parameter.
 """
 
 from __future__ import annotations
 
 import json
 import sqlite3
-from datetime import datetime, timezone
 from typing import Any
 
-from jobhunter.domain.pipeline.aggregate import JobPipelineState, OptimisticLockError
+from jobhunter.domain.pipeline.aggregate import JobPipelineState
 from jobhunter.domain.pipeline_types import (
     Blocked,
     Canceled,
@@ -31,10 +33,7 @@ from jobhunter.domain.pipeline_types import (
     serialize_stage_state,
 )
 from jobhunter.domain.tenant import TenantId
-
-
-def _utc_now() -> str:
-    return datetime.now(timezone.utc).isoformat()
+from jobhunter.state import record_job_event, set_stage_state
 
 
 def _json_loads(value: str | None, default: Any) -> Any:
@@ -44,12 +43,6 @@ def _json_loads(value: str | None, default: Any) -> Any:
         return json.loads(value)
     except Exception:
         return default
-
-
-def _json_dumps(value: Any) -> str | None:
-    if value in (None, "", [], {}, ()):
-        return None
-    return json.dumps(value, sort_keys=True)
 
 
 # ---------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


workers/automation/tests/test_pipeline_repository.py
```diff
@@ -167,3 +170,323 @@ def test_save_creates_new_rows_for_new_job(db):
...
+
+    agg = repo.load(LOCAL_TENANT, "https://example.com/job")
+    assert agg is not None
+
+    agg.set_stage_state(Stage.Enrich, Running(attempt_count=1, started_at="2026-05-01T00:00:00Z"))
+    agg.set_stage_state(Stage.Score, Running(attempt_count=1, started_at="2026-05-01T00:01:00Z"))
+    repo.save(agg)
+
+    events = _job_events(db, "https://example.com/job")
+    stages_with_events = sorted({stage for _, stage, _ in events})
```
ebarti: PR Reviewer - opus

Severity: Medium
Gate: Pass
Evidence: `uv --project workers/automation run --extra dev pyright workers/automation/tests/test_pipeline_repository.py` reports `error: Argument of type "set[str | None]" cannot be assigned to parameter "iterable" of type "Iterable[SupportsRichComparisonT@sorted]" in function "sorted" (reportArgumentType)` at this line. The base branch `temporal/runs-view` had zero pyright errors here; this PR introduces 11 new errors total in this test file.
Reproduction: Reproduced by running pyright on the worktree.

`stages_with_events = sorted({stage for _, stage, _ in events})` sorts a `set[str | None]` because `_job_events` returns `tuple[str, str | None, str | None]` (the `stage` and `level` columns are typed as nullable). `sorted` requires a comparable iterable, so passing optional strings is a type error and risks a runtime `TypeError` if any row in `job_events` has `stage=NULL` (the test only happens to pass because the seed data always has a non-null stage column).

Required fix: Either narrow the comprehension to `set[str]` by filtering out None (`{stage for _, stage, _ in events if stage is not None}`) or pass an explicit key (`sorted(s, key=lambda x: x or "")`). While here, also fix the related pyright errors at lines 329 and 336 — `reloaded.get_stage_state(Stage.Enrich).attempt_count` calls `.attempt_count` on a `StageState` union where only some variants (`Pending`, `Running`, `Succeeded`, `Failed`, `Exhausted`) carry that attribute; the test already asserts `isinstance(..., Running)` on line 328 but does not narrow the result before line 329, and line 336 has no isinstance guard at all. Add `assert isinstance(state, Running)` before the `.attempt_count` access (or rebind the result of `.get_stage_state(...)` to a local after asserting).

workers/automation/src/jobhunter/infrastructure/pipeline/sqlite_repository.py
```diff
@@ -232,95 +264,45 @@ def load(self, tenant_id: TenantId, job_url: str) -> JobPipelineState | None:
...
+                state.job_url,
+                stage_str,
+                new_state_str,
+                expected_version=state.version,
+                validate_transition=False,
+                **_stage_state_to_kwargs(stage_state),
+            )
+
+            if existing_states.get(stage_str) == new_state_str:
+                continue
```
ebarti, line 294: PR Reviewer - opus

Severity: Low
Gate: Pass
Evidence: `serialize_stage_state(state)` returns just the lowercase kind name (e.g., `"running"`, `"failed"`); see `workers/automation/src/jobhunter/domain/pipeline_types.py:serialize_stage_state` which is `state.kind.lower()`. The dedup compares `existing_states.get(stage_str) == new_state_str`.

The change-detection in this `save()` is **kind-only**: a stage transition like `Failed(error_code="A", message="foo") → Failed(error_code="B", message="bar")` (still kind=`Failed`, both serialize to `"failed"`) is treated as "no change" and emits no event, even though the persisted row's error fields are overwritten. Same for `Running(attempt_count=1) → Running(attempt_count=2)` and `Failed(retryable=True) → Failed(retryable=False)`. Today's only `repo.save()` callers (`CancelStageUseCase`, `RetryStageUseCase` in `domain/pipeline/use_cases.py`) only ever transition between distinct kinds, so the behavior is benign in production. But the docstring on `save()` says "When a row's persisted state actually changes, a per-stage event is emitted" — which is **misleading** because it implies any state change, not just a kind transition.

Required fix: Either tighten the docstring (e.g., "a per-stage event is emitted whenever the persisted state's *kind* changes") so future callers don't expect attempt_count / error_code / retryable churn to trigger an event, OR widen the comparison to include the relevant payload (e.g., compare against `(state, attempt_count, error_code, retryable)` from existing rows). Tightening the docstring is the smaller, safer change for this PR's scope.

workers/automation/src/jobhunter/state.py
```diff
@@ -188,6 +190,14 @@ def set_stage_state(
...
     ``ValueError`` is raised.  Pass ``validate_transition=False`` to bypass
     (e.g. during initial row creation or legacy migration).
+
+    When *expected_version* is supplied, the write is guarded by the row's
+    current ``version`` column (optimistic locking) and the new row is written
+    at ``version = expected_version + 1``.  If the existing row's version does
+    not match, ``OptimisticLockError`` is raised.  If no row exists yet the
+    INSERT is performed at the same bumped version.  Callers leave this as
+    ``None`` (the default) for the unguarded UPSERT path used by stage
+    runners.
```
ebarti, line 200: PR Reviewer - opus

Severity: Low
Gate: Pass
Evidence: With `expected_version=5` and no existing row, `new_version = 5 + 1 = 6`; the INSERT path succeeds with rowcount=1 and writes `version=6`. Verified by reproducing the SQL behavior in a `python3 -c` snippet against in-memory SQLite 3.53.0.
Reproduction: Reproduced with `python3 -c '...INSERT..ON CONFLICT...WHERE version=:expected_version'`.

The contract for `expected_version` says "If no row exists yet the INSERT is performed at the same bumped version." Concretely this means a caller passing `expected_version=5` for a brand-new row gets a row at `version=6`, even though no prior version 5 of that row ever existed. This is fine for the aggregate `save()` path (where `state.version` is the max across all rows and rows are saved together so they end up equal-versioned), but it is **silently incorrect** for any future caller that uses `set_stage_state(..., expected_version=N)` on a single row to assert "I am updating from version N → N+1" — if the row didn't exist, the helper INSERTs at `N+1` and returns success rather than raising. The docstring acknowledges this but does not flag the loaded gun: a stale write that beats out a concurrent INSERT will silently overwrite at a too-high version.

Required fix: Either (a) tighten the docstring to explicitly say "this is intended to be called from aggregate save() where ordering across rows guarantees consistent versions; do not call this for single-row optimistic-locked writes" with a concrete pointer, or (b) raise `OptimisticLockError(job_url, expected_version, 0)` when `expected_version > 0` and `cur.rowcount == 1` against an INSERT (no prior row). Option (a) is the smaller change and matches the "single-source-of-truth canonical helper" framing this PR is establishing — the caller's responsibility to use the right knob.

PR #38, opened 2026-05-07
feat(web): add Workflow Runs view with Temporal Web UI deep-link

ebarti:
## Summary

- Adds the **Workflow Runs view** at `/runs` — a new local web page that lists in-progress / failed / completed Temporal workflow runs (today: apply runs) and deep-links each row to the local Temporal Web UI for live debugging.
- Adds `GET /v1/workflow-runs` to the local TypeScript API, backed by `apply_run_projections` (now sourced from Temporal workflow histories per PR 4 of the stack).
- Adds the typed read path (`workflowRunsKeys` factory, `useWorkflowRunsListQuery` hook, `WorkflowRunSummary` schema) and wires the existing apply-run lifecycle events through the invalidation router so the table live-updates from the SSE stream.
- Adds a wider `<RunStatusBadge>` covering the full workflow-run state set (`canceled` / `terminated` / `timed_out` on top of the apply-run states).

## Stack context

- **Stack plan:** `docs/plans/proposed/2026-05-07-temporal-and-worker-reliability-stack.md`.
- **PR 5 of 7.** Stack so far: PR 1–4 are open / queued ahead of this PR.
- **Base branch:** `temporal/apply-runs-collapse` (NOT `main`). PR 6 / PR 7 will branch off `temporal/runs-view`.

## What's NOT in this PR

- **In-row "Cancel running workflow" button.** The JSON-RPC `cancel_run` handler + `CancelRunParamsSchema` already exist (PR 3 fixer); wiring an in-row UI button is a follow-up. A `// TODO` in `RunsView.tsx` pins the contract.
- **Worker Reliability cleanups.** PR 6 collapses the second stage-state write path; PR 7 registers logs + reports as artifacts.
- **Dedicated `WorkflowStarted` / `WorkflowCompleted` / `WorkflowFailed` / `WorkflowCanceled` event types.** The worker does not emit them yet — the existing apply-run events (`ApplyRunStarted` / `ApplicationSubmitted` / `ApplicationFailed`) drive the runs-list invalidation. The seam is set up so adding the `Workflow*` types and routing them through `workflowRunsKeys` is mechanical.
- **`runId` ≠ `workflowId` for any workflow.** Today they are equal (the Python `ApplyWorkflow` uses `info.workflow_id` as the timeline key). The schema exposes both as distinct fields so future non-apply workflows that separate the two land without a breaking read-model change.

## QA note

[...]

PR #41, opened 2026-05-07
feat(workers): add Langfuse observability via OpenTelemetry

ebarti:
## Summary

Wires the Python automation worker into Langfuse via OpenTelemetry. Every
LLM call, every Temporal workflow + activity, and every JSON-RPC dispatch
is now captured as an OTel span and exported via OTLP/HTTP (protobuf) to
the user's Langfuse instance. Init is idempotent and degrades gracefully
when Langfuse env vars are absent — the worker keeps running, only export
is off.

## Stack context

This is **PR 8** of the Temporal + Worker Reliability stack (added on
top of the original 7-PR plan). Plan:
[docs/plans/proposed/2026-05-07-temporal-and-worker-reliability-stack.md](docs/plans/proposed/2026-05-07-temporal-and-worker-reliability-stack.md).

- **Base:** `worker-reliability/logs-as-artifacts` (PR 7), **not** `main`.
- PRs ahead of this in the stack: #34, #35, #36, #37, #38, #39, #40.
- Branch: `observability/langfuse-otel`.

## What's emitted

| Source | Span name | `langfuse.observation.type` | Notes |
| --- | --- | --- | --- |
| `LLMClient.chat` (compat + native Gemini paths) | `llm.<model>` | `generation` | Sets `langfuse.observation.{model.name, model.parameters, input, output, usage_details}` and the GenAI semconv attrs (`gen_ai.request.model`, `gen_ai.response.model`, `gen_ai.usage.input_tokens`, `gen_ai.usage.output_tokens`). Token counts come from the OpenAI-compat `usage` block and the native Gemini `usageMetadata`. |
| Temporal workflow + activity execution | workflow / activity name | `span` (default) | Comes from `temporalio.contrib.opentelemetry.TracingInterceptor`, registered on both the **client** (`infrastructure/temporal/client.py`) and the **worker** (`infrastructure/temporal/worker.py`) so trace context propagates from the JSON-RPC dispatch span into the workflow run. |
| `JsonRpcServer.dispatch` | `rpc.<method>` | `span` | Sets `rpc.method`, `rpc.id`, `langfuse.trace.name=<method>`. Handler exceptions and method-not-found both flip the span to `StatusCode.ERROR`. |

The OTLP endpoint is `${LANGFUSE_BASE_URL}/api/public/otel/v1/traces`,
authentication is HTTP Basic over `base64(LANGFUSE_PUBLIC_KEY:LANGFUSE_SECRET_KEY)`,
and the `x-langfuse-ingestion-version: 4` fast-preview header is also
sent so traces show up in the Langfuse UI without ingestion lag.

## Behavior change

**Every LLM prompt and completion will be exported to the configured
Langfuse instance** when the credentials are present. `init_otel()` logs
a `WARNING` on startup so the user sees this before any LLM stage runs.
Opt-out paths:

- Unset any of `LANGFUSE_PUBLIC_KEY` / `LANGFUSE_SECRET_KEY` /
  `LANGFUSE_BASE_URL` — init logs a one-line warning and returns.
- Set `LANGFUSE_DISABLE=1` (or `true`) — init skips configuration even
  when credentials are present.

## What's NOT in this PR

- TS / web instrumentation (`apps/api`, `apps/web`, `packages/`).
- Distributed-trace propagation across the TS↔Python JSON-RPC boundary
  (would need TS to emit OTel context too — future PR).
- Dashboards / alerts / SLOs in the Langfuse UI.
- Prompt-versioning via Langfuse prompts.
- Cost calculation overrides — Langfuse computes cost from the token
  counts via its model registry.

## How to test locally

1. Set Langfuse credentials in `~/.jobhunter/.env`:

   ```env
   LANGFUSE_PUBLIC_KEY="pk-lf-..."
   LANGFUSE_SECRET_KEY="sk-lf-..."
   LANGFUSE_BASE_URL="https://us.cloud.langfuse.com"
   ```

2. Confirm the doctor row reports reachable:

   ```bash
   uv --project workers/automation run jobhunter doctor
   # ...
   # Langfuse  OK  reachable
   ```

3. Run any LLM-using command and check the Langfuse UI for the trace,
   for example:

   ```bash
   uv --project workers/automation run jobhunter score --limit 1
   ```

   Each `LLMClient.chat` call shows up in Langfuse as a generation with
   the prompt + completion + token usage attached. Workflow runs (when
   Temporal is up) appear with their activities as child spans.

4. Smoke-test the export path directly:

   ```bash
   uv --project workers/automation run python -c "
   from jobhunter.config import load_env; load_env()
   from jobhunter.infrastructure.observability import init_otel, shutdown_otel
   from jobhunter.infrastructure.observability.llm_spans import llm_generation_span
   init_otel()
   with llm_generation_span(model='test', messages=[{'role':'user','content':'hi'}], params={}) as record:
       record('pong', input_tokens=2, output_tokens=1)
   shutdown_otel()
   "
   ```

   Confirms `POST /api/public/otel/v1/traces 200` against the Langfuse
   ingest endpoint.

## Verification

| Command | Result |
| --- | --- |
| `uv --project workers/automation sync` | OK (4 OTel packages installed) |
| `uv --project workers/automation run --extra dev pytest -q` | **668 passed**, 3 noise warnings |
| `uv --project workers/automation run --extra dev ruff check .` | **All checks passed** |
| `pnpm api:check && pnpm api:test` | check OK, **66/66 tests passed** |
| `pnpm -r check` | all 6 workspaces OK |
| Live: `jobhunter doctor` with Langfuse env | `Langfuse  OK  reachable` |
| Live: `llm_generation_span` smoke export | `POST /api/public/otel/v1/traces HTTP/1.1 200` |

QA: skipped — automation/observability only, no UI surface change.

### DIFF
--- a/workers/automation/src/jobhunter/cli.py
+++ b/workers/automation/src/jobhunter/cli.py
@@ -1328,8 +1328,9 @@ def doctor() -> None:
     # Skip the network probe entirely if LANGFUSE_DISABLE is set — it's the
     # opt-out switch users flip when they don't want export running and they
     # shouldn't see a misleading "MISSING" or "unreachable" row.
-    lf_disabled = os.environ.get("LANGFUSE_DISABLE", "").strip().lower() in {"1", "true", "yes"}
-    if lf_disabled:
+    from jobhunter.infrastructure.observability import langfuse_disabled
+
+    if langfuse_disabled():
         results.append(("Langfuse", "[dim]disabled[/dim]", "LANGFUSE_DISABLE=1"))
     else:
         lf_pub = os.environ.get("LANGFUSE_PUBLIC_KEY", "").strip()
--- a/workers/automation/src/jobhunter/cli.py
+++ b/workers/automation/src/jobhunter/cli.py
@@ -1325,31 +1325,38 @@ def doctor() -> None:
                         "unreachable (start with: temporal server start-dev)"))
 
     # Langfuse OTLP ingest (observability target)
-    lf_pub = os.environ.get("LANGFUSE_PUBLIC_KEY", "").strip()
-    lf_sec = os.environ.get("LANGFUSE_SECRET_KEY", "").strip()
-    lf_url = os.environ.get("LANGFUSE_BASE_URL", "").strip().rstrip("/")
-    if not (lf_pub and lf_sec and lf_url):
-        results.append((
-            "Langfuse",
-            fail_mark,
-            "MISSING (set LANGFUSE_PUBLIC_KEY/SECRET_KEY/BASE_URL)",
-        ))
+    # Skip the network probe entirely if LANGFUSE_DISABLE is set — it's the
+    # opt-out switch users flip when they don't want export running and they
+    # shouldn't see a misleading "MISSING" or "unreachable" row.
+    lf_disabled = os.environ.get("LANGFUSE_DISABLE", "").strip().lower() in {"1", "true", "yes"}
+    if lf_disabled:
+        results.append(("Langfuse", "[dim]disabled[/dim]", "LANGFUSE_DISABLE=1"))
     else:
-        try:
-            resp = httpx.head(f"{lf_url}/api/public/otel/v1/traces", timeout=2.0)
-            # Any non-server-error response means the endpoint is alive.
-            # 405 (Method Not Allowed on HEAD) and 401 (auth required) both
-            # confirm the route exists.
-            if resp.status_code < 500:
-                results.append(("Langfuse", ok_mark, "reachable"))
-            else:
-                results.append((
-                    "Langfuse",
-                    fail_mark,
-                    f"unreachable (status={resp.status_code})",
-                ))
-        except Exception:  # noqa: BLE001 — any failure ⇒ unreachable
-            results.append(("Langfuse", fail_mark, "unreachable"))
+        lf_pub = os.environ.get("LANGFUSE_PUBLIC_KEY", "").strip()
+        lf_sec = os.environ.get("LANGFUSE_SECRET_KEY", "").strip()
+        lf_url = os.environ.get("LANGFUSE_BASE_URL", "").strip().rstrip("/")
+        if not (lf_pub and lf_sec and lf_url):
+            results.append((
+                "Langfuse",
+                fail_mark,
+                "MISSING (set LANGFUSE_PUBLIC_KEY/SECRET_KEY/BASE_URL)",
+            ))
+        else:
+            try:
+                resp = httpx.head(f"{lf_url}/api/public/otel/v1/traces", timeout=2.0)
+                # Any non-server-error response means the endpoint is alive.
+                # 405 (Method Not Allowed on HEAD) and 401 (auth required) both
+                # confirm the route exists.
+                if resp.status_code < 500:
+                    results.append(("Langfuse", ok_mark, "reachable"))
+                else:
+                    results.append((
+                        "Langfuse",
+                        fail_mark,
+                        f"unreachable (status={resp.status_code})",
+                    ))
+            except Exception:  # noqa: BLE001 — any failure ⇒ unreachable
+                results.append(("Langfuse", fail_mark, "unreachable"))
 
     # --- Render results ---
     console.print()
--- a/workers/automation/src/jobhunter/cli.py
+++ b/workers/automation/src/jobhunter/cli.py
@@ -8,6 +8,7 @@ import logging
 import time
 from typing import Optional
 
+import httpx
 import typer
 from rich.console import Console
 from rich.panel import Panel
@@ -65,6 +66,13 @@ def _bootstrap() -> None:
     load_env()
     ensure_dirs()
     init_db()
+    # Bootstrap OTel as early as possible so every span emitted by the
+    # rest of this CLI invocation flows to the configured Langfuse instance.
+    # init_otel() is idempotent and degrades gracefully when env vars are
+    # absent, so it's safe to call from every command.
+    from jobhunter.infrastructure.observability import init_otel
+
+    init_otel()
     try:
         # Pass a thread-local connection factory so the wildcard
         # subscriber (which fires on whichever thread published the
@@ -1184,7 +1192,14 @@ def worker(
         )
         await worker.run()
 
-    asyncio.run(_run())
+    from jobhunter.infrastructure.observability import shutdown_otel
+
+    try:
+        asyncio.run(_run())
+    finally:
+        # Flush any in-flight spans so the BatchSpanProcessor doesn't drop
+        # them on Ctrl-C.
+        shutdown_otel()
 
 
 @app.command()
@@ -1308,6 +1323,33 @@ def doctor() -> None:
     except (Exception, asyncio.TimeoutError):  # noqa: BLE001 — any failure ⇒ unreachable
         results.append(("Temporal", fail_mark,
                         "unreachable (start with: temporal server start-dev)"))
+
+    # Langfuse OTLP ingest (observability target)
+    lf_pub = os.environ.get("LANGFUSE_PUBLIC_KEY", "").strip()
+    lf_sec = os.environ.get("LANGFUSE_SECRET_KEY", "").strip()
+    lf_url = os.environ.get("LANGFUSE_BASE_URL", "").strip().rstrip("/")
+    if not (lf_pub and lf_sec and lf_url):
+        results.append((
+            "Langfuse",
+            fail_mark,
+            "MISSING (set LANGFUSE_PUBLIC_KEY/SECRET_KEY/BASE_URL)",
+        ))
+    else:
+        try:
+            resp = httpx.head(f"{lf_url}/api/public/otel/v1/traces", timeout=2.0)
+            # Any non-server-error response means the endpoint is alive.
+            # 405 (Method Not Allowed on HEAD) and 401 (auth required) both
+            # confirm the route exists.
+            if resp.status_code < 500:
+                results.append(("Langfuse", ok_mark, "reachable"))
+            else:
+                results.append((
+                    "Langfuse",
+                    fail_mark,
+                    f"unreachable (status={resp.status_code})",
+                ))
+        except Exception:  # noqa: BLE001 — any failure ⇒ unreachable
+            results.append(("Langfuse", fail_mark, "unreachable"))
 
     # --- Render results ---
     console.print()
--- a/workers/automation/src/jobhunter/infrastructure/observability/otel.py
+++ b/workers/automation/src/jobhunter/infrastructure/observability/otel.py
@@ -1,0 +1,114 @@
+"""Bootstrap OpenTelemetry to ship spans to a Langfuse OTLP/HTTP endpoint.
+
+Reads ``LANGFUSE_PUBLIC_KEY`` / ``LANGFUSE_SECRET_KEY`` / ``LANGFUSE_BASE_URL``
+straight from ``os.environ`` so deployments can override values that
+``jobhunter.config.load_env()`` already populated from ``~/.jobhunter/.env``.
+If any credential is missing, init logs a warning and returns without
+configuring — every command continues to run, only export is off.
+"""
+
+from __future__ import annotations
+
+import base64
+import logging
+import os
+
+from opentelemetry import trace
+from opentelemetry.exporter.otlp.proto.http.trace_exporter import OTLPSpanExporter
+from opentelemetry.sdk.resources import Resource
+from opentelemetry.sdk.trace import TracerProvider
+from opentelemetry.sdk.trace.export import BatchSpanProcessor
+
+from jobhunter import __version__ as JOBHUNTER_VERSION
+
+log = logging.getLogger(__name__)
+
+_initialized: bool = False
+_provider: TracerProvider | None = None
+_exporter: OTLPSpanExporter | None = None
+
+_OTLP_PATH = "/api/public/otel/v1/traces"
+
+
+def init_otel(*, service_name: str = "jobhunter", environment: str | None = None) -> None:
+    """Configure the global OTel ``TracerProvider`` once."""
+    global _initialized, _provider, _exporter
+
+    if _initialized:
+        return
+
+    if os.environ.get("LANGFUSE_DISABLE", "").strip().lower() in ("1", "true"):
+        log.info("Langfuse export disabled via LANGFUSE_DISABLE.")
+        _initialized = True
+        return
+
+    public_key = os.environ.get("LANGFUSE_PUBLIC_KEY", "").strip()
+    secret_key = os.environ.get("LANGFUSE_SECRET_KEY", "").strip()
+    base_url = os.environ.get("LANGFUSE_BASE_URL", "").strip().rstrip("/")
+    if not (public_key and secret_key and base_url):
+        log.warning(
+            "Langfuse OTel export disabled: set LANGFUSE_PUBLIC_KEY, "
+            "LANGFUSE_SECRET_KEY, and LANGFUSE_BASE_URL to enable."
+        )
+        _initialized = True
+        return
+
+    log.warning(
+        "Langfuse OTel export ENABLED — every LLM prompt + completion will be "
+        "shipped to %s. Set LANGFUSE_DISABLE=1 to opt out.",
+        base_url,
+    )
+
+    resource = Resource(
+        attributes={
+            "service.name": service_name,
+            "service.version": JOBHUNTER_VERSION,
+            "deployment.environment": environment or os.environ.get("JOBHUNTER_ENV", "local"),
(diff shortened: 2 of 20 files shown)


workers/automation/src/jobhunter/llm.py
```diff
@@ -134,15 +136,24 @@ def _chat_native_gemini(
...
-        resp.raise_for_status()
-        data = resp.json()
-        return data["candidates"][0]["content"]["parts"][0]["text"]
+        params = {"temperature": temperature, "max_tokens": max_tokens}
+        with llm_generation_span(model=self.model, messages=messages, params=params) as record:
+            resp = self._client.post(
+                url,
+                json=payload,
+                headers={"Content-Type": "application/json"},
+                params={"key": self.api_key},
```
ebarti: PR Reviewer - Code Owner

Severity: Blocker
Gate: Fail
Evidence: `workers/automation/src/jobhunter/llm.py:145` passes `params={"key": self.api_key}` to `httpx.Client.post`, producing the URL `https://generativelanguage.googleapis.com/v1beta/models/<model>:generateContent?key=<GEMINI_API_KEY>`. `init_otel()` in `workers/automation/src/jobhunter/infrastructure/observability/otel.py:92` calls `HTTPXClientInstrumentor().instrument()`, which globally patches every `httpx.Client` and emits a span per request whose `http.url` attribute contains the full URL including the query string. Reproduced locally with the OTel in-memory exporter:
```
Span name: GET
  http.method = GET
  http.url = http://localhost:9999/foo?key=SECRET-API-KEY-123
```
The user's brief confirms the live LLM path against `gemini-3.1-pro-preview` falls back to native via `_chat_compat`'s 403 → `_chat_native_gemini` switch, so this is the active code path against their Gemini account, not a hypothetical.
Reproduction: Reproduced with `python -c` invoking `HTTPXClientInstrumentor().instrument()` then `httpx.Client().get(..., params={'key': 'SECRET'})` and reading the captured span attributes.

The Gemini API key will be exfiltrated to the configured Langfuse instance in the `http.url` span attribute on every native-Gemini call (and on any other outbound httpx call that puts secrets in the query string). Langfuse stores trace attributes in plaintext.

Required fix: One of the following before merging:
1. Move the Gemini API key out of the URL and into the `x-goog-api-key` header in `_chat_native_gemini` (Google supports both — see https://ai.google.dev/gemini-api/docs/api-key); OR
2. Pass `request_hook` to `HTTPXClientInstrumentor().instrument(request_hook=...)` to scrub `?key=…` and any `Authorization: Bearer …` from the captured `http.url` / headers before the span is sent; OR
3. Drop the `HTTPXClientInstrumentor` auto-instrumentation entirely — the explicit `llm_generation_span` already captures everything you need from the LLM client, and the `_dispatch_streaming` / `_dispatch_workflow` spans already cover RPC. The auto-instrumentor is the leak vector; nothing in the brief requires it.
Also add a regression test that asserts no span attribute on any LLM call contains the value of `GEMINI_API_KEY` / `OPENAI_API_KEY`.

workers/automation/src/jobhunter/infrastructure/rpc/server.py
```diff
@@ -80,34 +83,45 @@ def dispatch(self, request: JsonRpcRequest) -> JsonRpcResponse | None:
...
-            return JsonRpcResponse.failure(
-                request.id,
-                JsonRpcError(INTERNAL_ERROR, "Internal error", data=str(exc)),
-            )
-        if request.id is None:
-            # Notification — no response.
-            return None
-        return JsonRpcResponse.success(request.id, result)
+        tracer = trace.get_tracer("jobhunter.rpc")
+        with tracer.start_as_current_span(f"rpc.{request.method}") as span:
```
ebarti, line 87: PR Reviewer - Code Owner

Severity: High
Gate: Fail
Evidence: `JsonRpcServer.dispatch()` correctly wraps the sync + workflow paths in a `tracer.start_as_current_span(f"rpc.{request.method}")` block, but `_dispatch_streaming()` (server.py line ~228) is invoked directly from `_handle_line()` (line ~224) and never opens a span. The architecture doc added in this PR (`docs/architecture.md`, "Every JSON-RPC dispatch") promises every dispatch gets a `rpc.<method>` span — broken for any handler registered with `mode="streaming"`.

Second issue on the same `dispatch` span: `_dispatch_workflow` is called from inside this span (line 102) and returns failure responses for: `workflow_starter is None`, `_RpcParamError`, generic `Exception` building the spec, the non-`WorkflowStartSpec` return-value branch, and `asyncio.run(self._workflow_starter(...))` raising. None of those failure branches calls `span.set_status(Status(StatusCode.ERROR, ...))` on the parent dispatch span. Workflow-start failures will appear as green / OK traces in Langfuse despite returning JSON-RPC errors.
Reproduction: Reasoned from code — `_handle_line` short-circuits to `_dispatch_streaming` before reaching `dispatch`; `_dispatch_workflow` failure branches return without touching the span status.

Required fix:
1. Wrap `_dispatch_streaming` in the same `tracer.start_as_current_span(f"rpc.{request.method}")` block with the same attribute set (rpc.method, rpc.id, langfuse.trace.name, langfuse.observation.type) and flip to ERROR on mid-stream exceptions.
2. In `_dispatch_workflow`, call `trace.get_current_span().set_status(Status(StatusCode.ERROR, ...))` (and `record_exception(exc)` for the exception branches) before returning each failure response.
3. Add a regression test that drives a streaming handler through `serve()` / `_handle_line` and asserts the `rpc.<method>` span is emitted; add a regression test that asserts a workflow-mode handler whose `_workflow_starter` raises produces a span with `StatusCode.ERROR`.

workers/automation/src/jobhunter/infrastructure/observability/otel.py
```diff
@@ -0,0 +1,114 @@
...
+        "Langfuse OTel export ENABLED — every LLM prompt + completion will be "
+        "shipped to %s. Set LANGFUSE_DISABLE=1 to opt out.",
+        base_url,
+    )
+
+    resource = Resource(
+        attributes={
+            "service.name": service_name,
+            "service.version": JOBHUNTER_VERSION,
+            "deployment.environment": environment or os.environ.get("JOBHUNTER_ENV", "local"),
```
ebarti, line 72: PR Reviewer - Code Owner

Severity: Medium
Gate: Pass
Evidence: `init_otel()` constructs a `Resource` at `otel.py:62` with `service.name`, `service.version`, and `deployment.environment` attributes — but no test in `test_otel_init.py` asserts these end up on emitted spans (or even on `_provider.resource`). The brief's "What to verify" list explicitly calls out `Resource attributes: service.name=jobhunter, service.version, deployment.environment from JOBHUNTER_ENV env or "local"`. Without coverage, a future refactor can silently drop or rename these and the only signal will be Langfuse traces missing the environment column.
Reproduction: Reasoned from code — grepped the test files for `service.name` / `JOBHUNTER_ENV` / `deployment.environment` and got no hits.

Required fix: Add a test that boots `init_otel()` with creds set and `JOBHUNTER_ENV=staging`, then asserts `otel_mod._provider.resource.attributes` contains `service.name == "jobhunter"`, `service.version == JOBHUNTER_VERSION`, and `deployment.environment == "staging"` (and that omitting `JOBHUNTER_ENV` defaults to `"local"`).

workers/automation/src/jobhunter/cli.py
```diff
@@ -1309,6 +1324,33 @@ async def _probe_temporal() -> None:
...
+    lf_url = os.environ.get("LANGFUSE_BASE_URL", "").strip().rstrip("/")
+    if not (lf_pub and lf_sec and lf_url):
+        results.append((
+            "Langfuse",
+            fail_mark,
+            "MISSING (set LANGFUSE_PUBLIC_KEY/SECRET_KEY/BASE_URL)",
+        ))
+    else:
+        try:
+            resp = httpx.head(f"{lf_url}/api/public/otel/v1/traces", timeout=2.0)
```
ebarti: PR Reviewer - Code Owner

Severity: Low
Gate: Pass
Evidence: The doctor probe at `cli.py:1327-1352` reads `LANGFUSE_PUBLIC_KEY` / `_SECRET_KEY` / `_BASE_URL` and reports MISSING / reachable / unreachable, but does not check `LANGFUSE_DISABLE`. A user who sets `LANGFUSE_DISABLE=1` will still see "Langfuse OK reachable" in `jobhunter doctor` output even though spans are not being exported.

Required fix: Either short-circuit with `results.append(("Langfuse", "[dim]disabled[/dim]", "LANGFUSE_DISABLE=1"))` when `LANGFUSE_DISABLE` is `1`/`true`, or add a "(export disabled)" suffix on the OK row. Cosmetic, but the brief calls out matching `is_otel_enabled()` semantics.

workers/automation/src/jobhunter/llm.py
ebarti: PR Fixer - claude

Fixed in 8493839.

Defense in depth applied:
1. `_chat_native_gemini` now sends the API key via the `x-goog-api-key` header instead of the `?key=` URL param, so OTel's `http.url` capture can no longer see it.
2. `init_otel()` registers a `request_hook` on `HTTPXClientInstrumentor` (`_scrub_url_credentials`) that REDACTs any `key=` query value still present on the captured URL, in case any future caller forgets the header convention.
3. New regression test `tests/test_llm_no_credential_leakage.py` wires an `InMemorySpanExporter` + `MockTransport`, drives the native Gemini path with `GEMINI_API_KEY=TEST_KEY_DEADBEEFCAFE`, and walks every span attribute asserting the substring is absent.

Verified before/after: with both `llm.py` + `otel.py` changes reverted from HEAD the test fails with `POST.http.url = https://...generateContent?key=TEST_KEY_DEADBEEFCAFE`; with the fix in place it passes.

workers/automation/src/jobhunter/infrastructure/rpc/server.py
ebarti: PR Fixer - claude

Fixed in c37b5ef.

1. `_dispatch_streaming` now opens the same `rpc.<method>` span that the sync `dispatch()` opens (with the same `rpc.method` / `rpc.id` / `langfuse.trace.name` / `langfuse.observation.type` attributes), wraps the iterator body, and marks `StatusCode.ERROR` on per-yield exception, on handler-start failure, and on the "did not return iterable" branch.
2. `_dispatch_workflow` now uses `trace.get_current_span()` (the dispatch caller already opened the span) and calls `span.set_status(Status(StatusCode.ERROR, "<short reason>"))` on every early-return failure: missing `workflow_starter`, handler raising `_RpcParamError`, handler raising any other exception while building the spec, handler returning a non-`WorkflowStartSpec`, and starter raising.
3. Added 6 new tests in `test_rpc_otel.py`:
   - `test_rpc_streaming_dispatch_emits_span`
   - `test_rpc_streaming_dispatch_marks_error_on_mid_stream_failure`
   - `test_rpc_workflow_mode_marks_span_error_on_starter_failure`
   - `test_rpc_workflow_mode_marks_span_error_on_missing_starter`
   - `test_rpc_workflow_mode_marks_span_error_on_handler_returning_non_spec`
   - `test_rpc_workflow_mode_marks_span_error_on_handler_raising_while_building_spec`

workers/automation/src/jobhunter/infrastructure/observability/otel.py
ebarti: PR Fixer - claude

Fixed in adee625.

Added `test_init_otel_sets_resource_attributes` to `test_otel_init.py`. It sets `JOBHUNTER_ENV=test` + Langfuse creds, calls `init_otel()`, then asserts the `TracerProvider.resource.attributes` dict carries `service.name == "jobhunter"`, `service.version == jobhunter.__version__`, and `deployment.environment == "test"`. After `shutdown_otel()` + clearing `JOBHUNTER_ENV` and re-resetting the global tracer once-gate, it re-inits and asserts the default `deployment.environment == "local"`.

workers/automation/src/jobhunter/cli.py
ebarti: PR Fixer - claude

Fixed in 3981ca7.

`doctor` now checks `LANGFUSE_DISABLE` (`{"1", "true", "yes"}`) before doing anything else for the Langfuse row. When set it appends `("Langfuse", "[dim]disabled[/dim]", "LANGFUSE_DISABLE=1")` and skips both the credential check and the `httpx.head` probe entirely. Added `test_doctor_langfuse_row_when_disabled` that patches `httpx.head` to raise `AssertionError` if invoked — confirming the network probe is short-circuited — and asserts the rendered output shows `disabled` + `LANGFUSE_DISABLE=1` and contains neither `MISSING` nor `unreachable`.

workers/automation/src/jobhunter/infrastructure/observability/otel.py
```diff
@@ -0,0 +1,143 @@
...
+
+
+def init_otel(*, service_name: str = "jobhunter", environment: str | None = None) -> None:
+    """Configure the global OTel ``TracerProvider`` once."""
+    global _initialized, _provider, _exporter
+
+    if _initialized:
+        return
+
+    if os.environ.get("LANGFUSE_DISABLE", "").strip().lower() in ("1", "true"):
```
ebarti: PR Reviewer - Code Owner

Severity: Medium
Gate: Pass
Evidence: `init_otel()` at `workers/automation/src/jobhunter/infrastructure/observability/otel.py:40` checks `os.environ.get("LANGFUSE_DISABLE", "").strip().lower() in ("1", "true")`, while `doctor` at `workers/automation/src/jobhunter/cli.py:1331` checks the same env var against `{"1", "true", "yes"}`. A user setting `LANGFUSE_DISABLE=yes` will see "disabled" in `jobhunter doctor` output but `init_otel` will still construct the exporter and ship spans on every CLI invocation. The two readers of the same flag must agree, otherwise the doctor row is actively misleading about whether export is on.
Reproduction: Reasoned from code — inspected both call sites side-by-side; documented behavior in README and architecture only mentions `LANGFUSE_DISABLE=1` so the `yes` value the doctor accepts has no other anchoring.

The `yes` alias was added in 3981ca7 to be lenient about how users opt out, which is fine — but `init_otel` was not updated to match. Pick one: either (a) extend `init_otel`'s tuple to `("1", "true", "yes")` (and add a test), or (b) drop `"yes"` from doctor's set so the two checks share the same truthy semantics. A small helper like `_is_disabled() -> bool` shared by both call sites would prevent this drifting again.

Required fix: Make the two readers agree on the accepted truthy values for `LANGFUSE_DISABLE`. Easiest path: factor a single `_langfuse_disabled() -> bool` function in `infrastructure/observability/otel.py`, import it from both `init_otel()` and `cli.doctor()`, and assert the parity in a small unit test.

workers/automation/src/jobhunter/infrastructure/observability/otel.py
```diff
@@ -0,0 +1,143 @@
...
+    key as a header (e.g. ``x-goog-api-key``), but if any future caller
+    accidentally puts ``?key=...`` back in the URL it would otherwise be
+    captured into the ``http.url`` span attribute and shipped to Langfuse.
+    The hook strips that single query param while preserving everything else.
+    """
+    try:
+        url = str(request.url)
+    except Exception:  # noqa: BLE001 — defensive: never break the request path
+        return
+    if "key=" not in url:
```
ebarti: PR Reviewer - Code Owner

Severity: Low
Gate: Pass
Evidence: `_scrub_url_credentials` at `workers/automation/src/jobhunter/infrastructure/observability/otel.py:110` early-returns with `if "key=" not in url`. This substring check is case-sensitive, but the inner scrub at line 118 uses `k.lower() == "key"` (case-insensitive). A URL like `https://example.com/x?KEY=secret` slips past the `"key=" not in url` check and is never scrubbed, even though the inner logic would have redacted it. Reproduced locally:
```
>>> from jobhunter.infrastructure.observability.otel import _scrub_url_credentials
>>> # span.attrs stays empty for KEY=, but scrubs key= correctly
>>> # 'https://example.com/x?KEY=secret&b=2' -> {} (no scrub)
>>> # 'https://example.com/x?key=secret' -> {'http.url': '...?key=REDACTED', ...}
```
Reproduction: Reproduced with a `FakeSpan` + `FakeReq` calling `_scrub_url_credentials` directly with mixed-case query keys.

This is purely defense-in-depth behind the actual fix (the Gemini key now flows via header, not URL), and Gemini's documented param is lowercase `key`, so the practical exposure is zero today. It only matters if a future caller re-introduces a `?KEY=` or `?Key=` URL — which is exactly the scenario the scrubber claims to defend against.

Required fix: Either lowercase the URL once before the membership test (`if "key=" not in url.lower():`) so the case-sensitive guard matches the case-insensitive scrub, or drop the early-return entirely and just always run the parse_qsl / urlencode path (the cost is negligible — one parse per outbound httpx request).

PR #141, opened 2026-06-08
feat(materials): canonical employer analysis (2-SDK ensemble)

ebarti:
## What

Phase 1: a reasoned, reproducible, **persisted** employer "ideal candidate" analysis that replaces the flakey `_extract_job_keywords` stopword heuristic and becomes the single source of truth driving downstream resume tailoring. The analysis is produced by a **two-SDK agent ensemble** — Claude Agent SDK + Codex SDK, run in parallel and reconciled by a Claude synthesizer pass (the Google/Antigravity leg is deferred; the port reserves a slot). This is a **backend + contracts + read-path vertical slice — there is no UI** (the inspector is Phase 5).

This is the first capability on the agent-SDK standard: all *new* AI runs through agent SDKs, never the legacy httpx `LLMClient`.

## Why

`_extract_job_keywords` produced flakey, ungrounded keywords from hardcoded stopword/high-signal lists. The new analysis: classifies requirements must-have vs nice-to-have with a 0–1 priority weight, ties every keyword to a quoted job-description evidence span (literal substring, deterministically validated), records role framing + inferred seniority + an ideal-candidate narrative, persists the full ensemble audit trail (per-model drafts + agreement + failures), and is cache-reproducible.

## How (locked decisions realised)

- **Domain** (`domain/materials/analysis.py`, `analysis_grounding.py`): `JobAnalysis`/`JobAnalysisDraft` Pydantic models (tier+weight D-14; quoted `evidence_span` D-15; `requirement_ref`+orphan flag D-17; rich narrative D-16); `EmployerAnalysis` aggregate (generation-versioned D-13; snapshot+version cache key D-11/D-12; per-model sub-analyses + failures + agreement D-08). Deterministic literal-substring grounding validator is the cardinal gate.
- **Ports + adapters + ensemble** (`domain/ports/materials.py`, `infrastructure/analysis/`): `AnalysisDraftPort`/`AnalysisSynthesizerPort` with Claude + Codex adapters (lazy-import, fully test-mockable SDK boundary); ensemble uses `asyncio.gather(..., return_exceptions=True)` so one SDK failure never cancels the healthy legs — partial failures are surfaced and persisted, never silently dropped. **No wall-clock timeout** (D-19); quality-only, top model + max effort (D-18).
- **Persistence**: canonical `job_employer_analysis` (+ sub-analysis/failure child tables), `SqliteEmployerAnalysisRepository` (cache lookup + supersede-not-destroy). Canonical rows, never `metadata_json` (D-09).
- **Lifecycle**: `AnalyzeJobUseCase` (cache short-circuit → ensemble → grounding re-validation → versioned persist → publish `EmployerAnalyzed`). Runs as the `_run_analyze` front-half sub-step of tailor (D-20); `TailorResumeUseCase` consumes the persisted analysis; **`_extract_job_keywords` removed outright (D-21, no shim)**. Standalone `analyze_job` JSON-RPC (D-10).
- **Cross-cutting sync**: `EmployerAnalyzed` event (Python factory + `DOMAIN_EVENT_TYPES` + `domain-types` mirror + `DomainEventUnion` + frontend invalidation handler + both parity tests); `analyze_job` (contracts `rpc.ts` + `local-actions.ts` + `handlers.py`); read-model field built identically by the Python projection builder **and** `apps/api/projections.ts`, served by `read-model.ts` as `JobDetail.employerAnalysis`, typed in `schemas.ts`.

## Validation

All SDK calls are **mocked** in tests — no live network/auth (live smoke is the operator's step).

- `ruff check .` → **All checks passed**
- `pytest -q` → **1206 passed, 1 failed** — the single failure (`test_suppress_backfilled_legacy_job_makes_selectors_treat_paths_inactive`) is **pre-existing on `main`**, unrelated to this change (materials-suppression backfill stats). 47 new tests added, all green.
- `pnpm api:check` → clean; `pnpm api:test` → **195 passed** (incl. new TS↔Python projection parity test)
- `pnpm web:check` → clean; `pnpm --filter @jobhunter/web test-d` → **no type errors**; web unit → **632 passed** (incl. `every-event-has-handler` + `invalidation-router` parity)

### ROADMAP Phase-1 success criteria
1. ✅ Tailoring produces a persisted `job_employer_analysis` record with must/nice-classified, weighted requirements.
2. ✅ Every keyword ties to a literal-substring JD evidence span; `_extract_job_keywords` no longer runs.
3. ✅ Re-analysis on the same snapshot hits the cache (reproducibility fixture passes); re-tailor reuses it, keyed by snapshot hash.
4. ✅ Served through the canonical read path (projection + DTO, Python↔TypeScript parity), readable back as an inspectable artifact.

### Could not be verified without live SDK auth
The end-to-end ensemble against real Claude/Codex sessions (actual extraction quality, real structured-output shapes, the EEO/subtext LLM-judge advisory signals) — that is the operator's live smoke step. The deterministic gates (grounding, schema/invariants, partial-failure path, cache) are fully covered by mocked tests.

## Note on commit range
This branch was created from local `main`, which is 9 commits ahead of `origin/main` with the user's own Phase-1 `.planning/` docs. Those additive doc commits appear in this PR's range; the only **code** change is the single `feat(materials)` commit.


workers/automation/src/jobhunter/infrastructure/analysis/prompts.py
```diff
@@ -0,0 +1,115 @@
...
+4. KEYWORDS: real skills/tools/quals, deduplicated. Link each to the \
+requirement it supports via requirement_ref (use the matching requirement id). \
+A keyword with no clear parent requirement may set requirement_ref to null. Do \
+NOT produce a bag of every noun, near-duplicate padding, or soft-skill filler.
+5. IGNORE BOILERPLATE: "fast-paced", "rockstar/ninja", "wear many hats", \
+"passionate self-starter", EEO/benefits/"about us"/legal sections are NOT \
+requirements or keywords.
+6. EEO: NEVER convert protected-class signals ("recent grad", "digital native", \
+"young/energetic", gendered language) into a requirement, keyword, or an "ideal \
+candidate" attribute.
```
ebarti, line 64: PR Reviewer - claude-opus-pr-reviewer

Severity: Medium
Gate: Pass
Evidence: AI-SPEC Section 6 (Online Guardrails, Dimension 9 "EEO red-flag screen") specifies the intervention as **Code (denylist/regex screen of requirement+keyword text against EEO red-flag terms)** — "Block from the requirement/keyword set + flag … record the screen hit as an audit note." The only EEO enforcement that ships is this prompt instruction (rule 6 here). A grep of `domain/materials/` and `infrastructure/analysis/` finds no deterministic denylist/regex screen; nothing strips a protected-class signal that a model emits anyway, and no audit note is recorded.
Reproduction: Reasoned from code (grep for eeo/denylist/protected-attribute in the analysis path returns only prompts.py) + AI-SPEC Section 6 table.

The spec frames the grounding/schema/partial-ensemble gates as deterministic code (and they are, correctly). EEO is the one online guardrail in the table whose code half is absent — it is prompt-only. Per the spec's own auditability discipline ("every displayed claim has an explicit source of truth"; the validator, not the prompt, is the real gate), a model that emits "recent grad" / "digital native" / gendered language as a requirement or keyword would pass through to the canonical record and downstream tailoring with no deterministic interception and no audit note.

This is Medium (the spec itself marks Dimension 9 priority Medium and the run does not abort), not a merge blocker — but it is a real gap against the locked guardrail table and a fairness-sensitive surface.

Required fix: Add a deterministic EEO red-flag screen in the domain layer (sibling to analysis_grounding.py) that scans every Requirement.text/evidence_span and ReasonedKeyword.keyword/evidence_span against a protected-attribute denylist, drops/flags matches from the requirement+keyword set before persistence, and records the screen hit as audit data on the EmployerAnalysis record — with a unit test that proves a protected-class phrase is removed and recorded. If deferring to a later phase is intended, state that explicitly in the PR description and the phase docs since Section 6 lists it as an online (generation-time) guardrail for this phase.

workers/automation/src/jobhunter/infrastructure/analysis/claude_analysis_adapter.py
```diff
@@ -0,0 +1,188 @@
...
+            raise RuntimeError("Claude structured-output retries exhausted")
+        candidate = getattr(message, "structured_output", None)
+        if candidate is not None:
+            structured = candidate
+    if structured is None:
+        raise RuntimeError("Claude Agent SDK returned no structured output for output_format")
+    if not isinstance(structured, dict):
+        # Some SDK shapes hand back a JSON string; tolerate that.
+        structured = json.loads(structured)
+    return structured
```
ebarti, line 86: PR Reviewer - claude-opus-pr-reviewer

Severity: Low
Gate: Pass
Evidence: `_structured_output_from_messages` is annotated `-> dict[str, Any]`. On line 81 `structured = json.loads(structured)` is reached when the SDK hands back a JSON string; `json.loads` can return a list or scalar, which line 82 then returns, violating the declared return type. The downstream caller `JobAnalysis.model_validate(structured)` (draft) would then raise a `ValidationError` on a non-dict rather than the clearer "no structured output" RuntimeError this function is meant to raise. (This is the IDE-flagged `Any|dict|None` vs `dict[str, Any]` diagnostic — confirmed real, but contained: the ensemble retry loop catches ValidationError and records the leg as a failure, so it cannot corrupt the canonical record.)

Required fix: After `json.loads`, assert/normalise to a dict — e.g. `if not isinstance(structured, dict): raise RuntimeError("Claude Agent SDK structured output was not a JSON object")` — so the contract holds and the failure reason is precise.

workers/automation/src/jobhunter/infrastructure/analysis/ensemble.py
```diff
@@ -0,0 +1,282 @@
...
+        return 1.0
+    union = a | b
+    if not union:
+        return 1.0
+    return len(a & b) / len(union)
+
+
+def _flag_non_unanimous(
+    drafts: tuple[JobAnalysisDraft, ...],
+    sets: list[set[str]],
```
ebarti: PR Reviewer - claude-opus-pr-reviewer

Severity: Low
Gate: Pass
Evidence: `_flag_non_unanimous(drafts, sets, *, kind)` declares a `drafts` parameter (line 155) but the body never references it — it `del kind` and computes purely from `sets` (lines 162-167). Both call sites (lines 125-126) pass `drafts` needlessly. Dead parameter; ruff does not flag unused function params so it slipped the lint gate. (This is the IDE-flagged "ensemble.py:155 unused drafts" diagnostic — confirmed real, cosmetic.)

Required fix: Drop the `drafts` parameter (and the matching call-site args) so the signature reflects what the function actually consumes. Note the separate IDE hint about `ensemble.py:269` accessing `.violations` on a `BaseException` is a FALSE positive — line 268 already narrows with `isinstance(error, GroundingError)` before the access.

workers/automation/src/jobhunter/infrastructure/analysis/prompts.py
ebarti: PR Fixer - claude-opus-pr-fixer

Fixed in 165fdb0. Added a deterministic, domain-layer EEO red-flag screen as a sibling to `analysis_grounding.py`: `workers/automation/src/jobhunter/domain/materials/analysis_eeo_screen.py`.

What it does:
- Word-boundary regex denylist over `Requirement.text` + `Requirement.evidence_span` and `ReasonedKeyword.keyword` + `ReasonedKeyword.evidence_span`, focused on the unambiguous EEOC ad-language signals: age ("recent grad", "digital native", "young/energetic"), explicitly gendered job-title nouns (salesman/foreman/waitress/...), and a few explicit protected-class phrasings. The gender pattern was deliberately tightened to whole gendered titles (NOT bare "man"/"men") to keep false positives near zero on legitimate text ("management", "human", "craftsmanship", "men and women" in an evidence span all pass).
- On a match it DROPS the offending requirement/keyword from the canonical set (plus any keyword orphaned by a dropped parent requirement) and records the drop as an `EeoScreenHit` audit note (kind, ref_id, protected-class category, matched phrase). It never aborts the run.

Wiring: `AnalyzeJobUseCase.execute_async` runs the screen right after the grounding re-validation and before persistence (`screen_eeo_red_flags(outcome.canonical)`), persists the cleaned canonical, and records the hits as canonical audit data via a new `eeo_screen_json` column (round-trips on `load()`).

Scope note: `EmployerAnalysis.to_read_model()` does NOT yet emit the hits, which keeps the cross-runtime projection parity (`apps/api` ↔ Python) intact — surfacing them in the inspector read model + the TS mirror is the natural Phase 5 inspector-UI slice. Verified: `apps/api` projection-parity test green (`pnpm --filter @jobhunter/api test` = 195 passed).

Tests: `workers/automation/tests/test_employer_analysis_eeo_screen.py` (drop+record for requirement and keyword, orphan-child removal, gendered title, false-positive guards), a use-case test in `test_analyze_job_use_case.py` proving a grounded protected-class requirement is dropped before persist and recorded, and a repository round-trip test of the audit hits.

workers/automation/src/jobhunter/infrastructure/analysis/claude_analysis_adapter.py
ebarti: PR Fixer - claude-opus-pr-fixer

Fixed in 70e3d12. After the `json.loads` fallback in `_structured_output_from_messages`, the result is re-checked and a precise error is raised when it is not a dict:

```python
if not isinstance(structured, dict):
    structured = json.loads(structured)
if not isinstance(structured, dict):
    raise RuntimeError("Claude Agent SDK structured output was not a JSON object")
return structured
```

The `-> dict[str, Any]` contract now holds, and a JSON array/scalar surfaces as the intended "no structured output" RuntimeError rather than a later `JobAnalysis.model_validate` ValidationError.

workers/automation/src/jobhunter/infrastructure/analysis/ensemble.py
ebarti: PR Fixer - claude-opus-pr-fixer

Fixed in 70e3d12. Dropped the unused `drafts` parameter (and the `kind` marker it `del`-eted) from `_flag_non_unanimous`; the signature is now `_flag_non_unanimous(sets: list[set[str]])` and both call sites pass only the precomputed sets. Confirmed the separate IDE hint about `ensemble.py` `.violations` on a `BaseException` is a false positive — `_raw_output_from_error` narrows with `isinstance(error, GroundingError)` before the access, so no change there.

PR #142, opened 2026-06-09
feat(materials): per-bullet provenance + granular controls

ebarti:
## What

Phase 2 of grounded resume tailoring: **every generated resume bullet now carries a canonical, FK-bound `BulletProvenance` record** (evidence × requirement × transform × control × rationale) consuming Phase 1's persisted `EmployerAnalysis`, with **never-fabricate enforced by a deterministic detector independent of the prompt**. Backend + contracts + read-path only (no UI — that is Phase 5).

## Why

The category's cardinal failure is fabrication (invented metrics/titles/dates/employers) and "opaque tailoring". Phase 2 makes every displayed bullet claim have an explicit source of truth in canonical rows, and adds a deterministic gate so a model that ignores the prompt still cannot ship an unsourced number/title/date/employer.

## How

**Domain**
- `BulletProvenance` entity + generation-versioned `BulletProvenanceSet` (`domain/materials/provenance.py`).
- Closed enums `TransformType` (verbatim / rephrase / reframe / synthesize_from_related / quantify_from_evidence) and `ControlRule` (rephrase always allowed; invent only for closely-related; never fabricate metrics/titles/dates/employers) in `value_objects.py`.
- `provenance_builder.py` computes one row per **rendered** bullet (same `resume_profile` helpers the assembler uses, so `generated_text` is the exact rendered line — the coverage anchor), maps change-type → `TransformType`, and binds `requirement_ids` as real FKs into the analysis. A fabricated evidence/requirement id raises `ProvenanceBindingError` before any row is built (GROUND-05).

**Deterministic never-fabricate detector** (`domain/materials/fabrication_detector.py`)
- Pure, prompt-independent (the Phase-2 analogue of Phase 1's grounding gate). Every numeric/date/percentage/money/title/company-suffixed-employer token in a generated bullet must trace to recorded profile evidence; a token that does not is hard-rejected at generation time.

**Use case + persistence**
- `TailorResumeUseCase` computes provenance + runs the detector after assembling the selected candidate's text. A fabrication downgrades validation so the resume is **not approved** and the last accepted generation's artifact + provenance are preserved (Anti-Pattern 4). An accepted generation persists rows via `BulletProvenanceRepository` transactionally with the artifact and publishes `BulletProvenanceRecorded`.
- Canonical `job_bullet_provenance` table (NOT `metadata_json`), generation-versioned + supersede-not-destroy.

**Event** `BulletProvenanceRecorded`: Python factory + `DOMAIN_EVENT_TYPES`; `packages/domain-types` mirror + union + `DOMAIN_EVENT_TYPES`; web `materials/handlers.ts` + invalidation-router registration + fixture; `every-event-has-handler` parity passes.

**Read path**
- Projected onto `artifact_list_projections.bullet_provenance_json` by **both** the Python builder and `apps/api/projections.ts` (parity), served on `ArtifactTailoringExplanation.bulletProvenance` by `read-model.ts` (a PDF artifact resolves it from the sibling tailored-resume row); DTO added to `schemas.ts`.
- The divergent legacy sibling-file / TS-recompute read paths are intentionally left untouched — that is Phase 4.

## ROADMAP Phase-2 success criteria

1. ✅ Each bullet has a `job_bullet_provenance` row linking profile evidence + job requirement, recording a transform from the closed taxonomy + a rationale.
2. ✅ Provenance is FK bindings, not free text — a fabricated evidence/requirement id is hard-rejected at generation time (fixture proves the reject).
3. ✅ The governing control rule is recorded per bullet.
4. ✅ A deterministic numeric/date/title detector runs independently of the prompt — metrics-hungry job + numberless profile ⇒ zero unsourced numerics (fixture proves it).
5. ✅ Provenance is generation-versioned and superseded with its artifact — a failed re-tailor never destroys the last accepted generation's provenance (fixture proves it).

## How validated

- `uv --project workers/automation run --extra dev ruff check .` → **pass**
- `uv --project workers/automation run --extra dev pytest -q` → **1230 passed, 1 failed** (only the pre-existing unrelated `test_suppress_backfilled_legacy_job_makes_selectors_treat_paths_inactive`)
- `pnpm api:check` → pass · `pnpm api:test` → **196 passed** (incl. new cross-runtime provenance parity test)
- `pnpm web:check` → pass · `pnpm --filter @jobhunter/web test --run` → **637 passed** (incl. `every-event-has-handler` + router parity) · `pnpm --filter @jobhunter/web test-d` → **10 passed**

Docs: `docs/architecture.md` + `docs/job-pipeline-architecture.md` updated for the new canonical audit data, event, and read path.

### DIFF
--- a/workers/automation/src/jobhunter/domain/materials/fabrication_detector.py
+++ b/workers/automation/src/jobhunter/domain/materials/fabrication_detector.py
@@ -1,0 +1,364 @@
+"""Deterministic never-fabricate detector — the cardinal grounding gate (CONTROL-03).
+
+The Phase-2 analogue of Phase 1's ``analysis_grounding.py``: a pure, deterministic
+check that runs **independently of the tailoring prompt**. Every numeric, date,
+percentage, money, title, and employer token that appears in a generated resume
+bullet MUST trace to recorded profile evidence. A token that does not is a
+fabrication (the category's cardinal failure — invented metrics/titles/dates/
+employers) and is HARD-REJECTED at generation time — never trusted from the model,
+and never expressible in a JSON schema or a prompt instruction.
+
+Why a separate deterministic detector at all (CONTROL-03 / GROUND-05):
+
+  * The prompt can ask the model not to fabricate; the model can ignore it. The
+    detector is the *real* gate — it does not ask, it checks.
+  * It runs against the **actual generated bullet text** (the rendered line the
+    user sees), not the model's self-reported provenance, so a model that lies
+    about its own grounding still fails here.
+
+The "evidence corpus" is assembled once from canonical profile data (experience
+bullets, evidence items, verified metrics, titles, companies, date ranges,
+education) — the source of truth a real fact must appear in. Membership is decided
+by literal containment after normalising only insignificant whitespace and case,
+so trivial formatting differences do not produce false rejections while genuine
+fabrications still fail.
+
+This module has NO LLM call and NO I/O. It is unit-tested directly.
+"""
+
+from __future__ import annotations
+
+import re
+from dataclasses import dataclass, field
+from typing import Any
+
+from jobhunter.domain.materials.value_objects import ControlRule
+from jobhunter.resume_profile import (
+    get_achievement_evidence,
+    get_education_entries,
+    get_experience_entries,
+    get_resume_constraints,
+)
+
+_WHITESPACE_RE = re.compile(r"\s+")
+
+# Numeric/metric tokens: money, percentages, multipliers, and bare numbers with a
+# unit or magnitude. Mirrors the metric grammar used by the quality evaluator so
+# the detector and the validator agree on what counts as a "number" in a bullet.
+_NUMERIC_RE = re.compile(
+    r"(?ix)"
+    r"(?:\$\s?\d+(?:[,.]\d+)*(?:\.\d+)?\s?(?:k|m|b|million|billion)?)"  # money
+    r"|(?:\b\d+(?:\.\d+)?\s?%)"  # percentage
+    r"|(?:\b\d+(?:\.\d+)?\s?x\b)"  # multiplier
+    r"|(?:\b\d[\d,]*(?:\.\d+)?\+?\b)"  # bare integer/decimal (incl. "10k", "5+")
+)
+
+# Four-digit years and explicit month/day-month-year date spans. Standalone small
+# integers are handled by ``_NUMERIC_RE``; this catches calendar dates that a
+# bullet might invent (e.g. "since 2019", "Jan 2020 - Mar 2021").
+_DATE_RE = re.compile(
+    r"(?ix)"
+    r"\b(?:19|20)\d{2}\b"  # a four-digit year
+    r"|\b(?:jan|feb|mar|apr|may|jun|jul|aug|sep|oct|nov|dec)[a-z]*\.?\s+(?:19|20)\d{2}\b"
+)
+
+# Leadership/seniority title tokens whose presence in a bullet implies a claimed
+# level. They are only a fabrication when the same signal is absent from the
+# profile evidence corpus (CONTROL never_fabricate_titles).
+_TITLE_TOKEN_RE = re.compile(
+    r"(?ix)\b("
+    r"chief|cto|ceo|cio|vp|vice\s+president|director|head\s+of|principal|staff|"
+    r"senior|sr\.?|lead|architect|founder|co-?founder|manager|executive"
+    r")\b"
+)
+
+# A company-like proper noun anchored by an explicit company suffix. Anchoring on
+# the suffix keeps the employer check PRECISE: it fires on "Globex Corporation" /
+# "Initech Inc." but never on ordinary prose like "with Python" or "for Docker",
+# which a bare "<preposition> <Capitalized word>" heuristic would wrongly flag.
+_EMPLOYER_RE = re.compile(
+    r"\b([A-Z][A-Za-z0-9&.\-]*(?:\s+[A-Z][A-Za-z0-9&.\-]*)*"
+    r"\s+(?:Inc|Incorporated|LLC|Ltd|Limited|Corp|Corporation|Co|Company|GmbH|PLC|AG|SA|NV|LLP)\.?)\b"
+)
+
+# Tokens that, while numeric, never imply a fabricated claim on their own — they
+# are structural/sequence numbers, not metrics. Kept tiny and explicit.
+_BENIGN_NUMERIC_TOKENS = frozenset({"0", "1", "2", "3", "4", "5", "24", "365", "24/7"})
+
+
+def _normalize(text: str) -> str:
+    """Lowercase + collapse insignificant whitespace; preserve word content.
+
+    Case and whitespace are normalised so a quote that differs only in spacing or
+    capitalisation still matches; punctuation and digits are preserved so the
+    check stays a genuine literal-containment test.
+    """
+    return _WHITESPACE_RE.sub(" ", text).strip().lower()
+
+
+def _digits_only(token: str) -> str:
+    """Return only the digit/decimal characters of a token for numeric matching.
+
+    A bullet may render ``$1.2M`` while the evidence records ``1,200,000`` or
+    ``1.2 million``; matching on the run of significant digits avoids false
+    fabrication flags on equivalent renderings while still catching invented
+    numbers (whose digits appear nowhere in the corpus).
+    """
+    return re.sub(r"[^\d.]", "", token).strip(".")
+
+
+@dataclass(frozen=True)
+class EvidenceCorpus:
+    """The canonical profile text + token sets a real bullet fact must trace to.
+
+    Assembled once from profile data via :func:`build_evidence_corpus`. ``text``
+    is the normalised concatenation of every grounded source; the token sets are
+    pre-extracted so per-bullet checks are cheap and deterministic.
+    """
+
+    text: str
+    numeric_digits: frozenset[str]
+    title_tokens: frozenset[str]
+    date_tokens: frozenset[str]
+
+    def contains_phrase(self, phrase: str) -> bool:
+        normalized = _normalize(phrase)
+        return bool(normalized) and normalized in self.text
+
+    def has_numeric(self, token: str) -> bool:
+        digits = _digits_only(token)
+        if not digits:
+            return True  # nothing numeric to ground
+        return digits in self.numeric_digits
+
+    def has_title(self, token: str) -> bool:
+        return _normalize(token) in self.title_tokens
+
+    def has_date(self, token: str) -> bool:
+        return _digits_only(token) in self.date_tokens or _normalize(token) in self.text
+
+
+def _collect(value: Any, into: list[str]) -> None:
+    """Flatten arbitrary nested profile values into a list of non-empty strings."""
+    if value is None:
+        return
+    if isinstance(value, str):
+        if value.strip():
+            into.append(value)
+    elif isinstance(value, dict):
+        for child in value.values():
+            _collect(child, into)
+    elif isinstance(value, (list, tuple, set)):
+        for child in value:
+            _collect(child, into)
+    else:
+        text = str(value).strip()
+        if text:
+            into.append(text)
+
+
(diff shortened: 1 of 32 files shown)


workers/automation/src/jobhunter/domain/materials/provenance_builder.py
```diff
@@ -0,0 +1,406 @@
...
+    sources = _sources(plan, analysis)
+    claim_mode = get_claim_mode(profile)
+    resume = get_resume_master(profile)
+    rows: list[BulletProvenance] = []
+
+    # ---- Executive profile (single line) --------------------------------
+    baseline_summary = _normalize_space(
+        str(resume.get("executive_profile", {}).get("baseline_text", ""))
+    )
+    tailored_summary = _normalize_space(str(tailored_payload.get("executive_profile") or ""))
```
ebarti: PR Reviewer - claude-opus-pr-reviewer

Severity: High
Gate: Fail
Evidence: `provenance_builder.py:243` records the executive-profile row's `generated_text` (and at :251/:266 the matched keywords + coverage anchor) from `tailored_payload.get("executive_profile")` unconditionally. The actual assembler gates the shipped summary on policy: `services.py:515-518` uses `data.get("executive_profile")` only when `tailoring_policy["allow_summary_rewrite"]` is true, otherwise it ships `resume.executive_profile.baseline_text`. I reproduced the divergence with `allow_summary_rewrite=False`: shipped resume line = `'Senior engineer (baseline).'` but the persisted provenance `generated_text` = `'Backend engineer who delivered 99.99% uptime.'` (the model's proposed summary that never ships). The fabrication detector in `use_cases.py:1786` also scans this non-shipped text.
Reproduction: Reproduced with an inline `build_bullet_provenance` + `ResumeAssembler.assemble_resume_text` harness over a profile whose `tailoring_policy.allow_summary_rewrite=False`.

This breaks the central Phase 2 invariant (ROADMAP success criterion 1 / Pattern 2 / GROUND-06): provenance `generated_text` is documented as "byte-identical to what ResumeAssembler renders — the row is a true coverage anchor", but for the executive-profile section it is not when summary rewrite is disabled. Consequences: (1) the inspector (Phase 5) will display provenance, matched keywords, and a rationale for a summary line the user never received; (2) coverage/`requirement_ids` for that row are computed against text that is not on the resume; (3) the deterministic detector scans text that won't ship — conservative for fabrication (false-positive reject of a clean resume), but it means the gate is not anchored to the rendered line. Experience and skills rows are unaffected because they go through the same `tailored_experience_*`/`tailored_skill_items` helpers the assembler uses; only the executive profile bypasses the policy gate. There is no fixture covering `allow_summary_rewrite=False`.

Required fix: Render the executive-profile `generated_text` through the same policy-gated path the assembler uses (honor `tailoring_policy.allow_summary_rewrite`: emit the baseline summary when rewrite is off, the tailored summary when on), so the provenance anchor and the detector input equal the shipped line. Add a regression fixture with `allow_summary_rewrite=False` asserting the executive-profile provenance `generated_text` equals the assembled resume's summary line.

workers/automation/src/jobhunter/domain/materials/fabrication_detector.py
```diff
@@ -0,0 +1,364 @@
...
+
+    def contains_phrase(self, phrase: str) -> bool:
+        normalized = _normalize(phrase)
+        return bool(normalized) and normalized in self.text
+
+    def has_numeric(self, token: str) -> bool:
+        digits = _digits_only(token)
+        if not digits:
+            return True  # nothing numeric to ground
+        return digits in self.numeric_digits
```
ebarti: PR Reviewer - claude-opus-pr-reviewer

Severity: Medium
Gate: Pass
Evidence: `has_numeric` (line 128-132) grounds a numeric token if its significant-digit run is present ANYWHERE in the corpus-wide `numeric_digits` set (built globally in `build_evidence_corpus`, lines 218-220). Membership is by digit-run, not by value/unit/context. I reproduced: with a profile whose only number is "35% latency reduction" (`numeric_digits={'35','2020','2015'}`), the detector passes a fabricated "Drove $35M in revenue." and "Scaled to 35 million users." — both reuse the digit-run "35" so `has_numeric('$35M')`/`has_numeric('35')` return True and no finding is emitted. "Handled 150 incidents." (digits absent) is correctly flagged.
Reproduction: Reproduced with an inline `find_fabricated_tokens` harness over a single-number profile.

This deviates from the stated guarantee (ROADMAP success criterion 4: "every numeric/date/title token traces to recorded profile evidence"; PITFALLS Pitfall 3: "require each [numeric/currency/date] to be present in (or derivable from) canonical evidence ... any unsourced metric is a hard fail"). A fabricated dollar/percentage/magnitude that coincidentally shares a digit-run with an unrelated profile number ("35% latency" grounding "$35M revenue") ships unflagged — exactly the costly case the detector exists to stop. Note: the specific criterion-4 fixture (`test_metrics_hungry_job_with_numberless_profile_yields_zero_unsourced_numerics`) passes only because a numberless profile yields an empty digit set so every number is caught; the gap is invisible to that fixture. The LLM judge + adversarial review are a second layer and Phase 3 re-validates against rendered text, so this is defense-in-depth weakening rather than a total bypass.
Reproduction: Reasoned from code + reproduced as above.

Required fix: Tighten numeric grounding beyond bare digit-run membership — e.g., match the full normalized numeric token (digits + unit/magnitude, so "$35M"/"35 million" must match a corpus token of the same magnitude, not just "35"), and/or scope the match so a percentage only grounds against a percentage and a currency only against a currency. Add a fixture: a profile containing one number + a fabricated different-magnitude/different-unit number sharing its digits must be flagged.

workers/automation/src/jobhunter/domain/materials/fabrication_detector.py
```diff
@@ -0,0 +1,364 @@
...
+)
+
+# A company-like proper noun anchored by an explicit company suffix. Anchoring on
+# the suffix keeps the employer check PRECISE: it fires on "Globex Corporation" /
+# "Initech Inc." but never on ordinary prose like "with Python" or "for Docker",
+# which a bare "<preposition> <Capitalized word>" heuristic would wrongly flag.
+_EMPLOYER_RE = re.compile(
+    r"\b([A-Z][A-Za-z0-9&.\-]*(?:\s+[A-Z][A-Za-z0-9&.\-]*)*"
+    r"\s+(?:Inc|Incorporated|LLC|Ltd|Limited|Corp|Corporation|Co|Company|GmbH|PLC|AG|SA|NV|LLP)\.?)\b"
+)
```
ebarti, line 112: PR Reviewer - claude-opus-pr-reviewer

Severity: Medium
Gate: Pass
Evidence: `_EMPLOYER_RE` (lines 79-82) only matches a proper noun that ends in an explicit corporate suffix (Inc/Corp/LLC/Ltd/GmbH/PLC/...). A bare-name fabricated employer is not matched. I reproduced: "Owned the API at Netflix." → no finding (the employer arm never fires, and "netflix" is absent from the title/numeric/date arms too, so the whole bullet passes). "Globex Corporation" is correctly flagged because of the suffix.
Reproduction: Reproduced with an inline `find_fabricated_tokens` harness.

Assessment of the suffix-anchor tradeoff vs CONTROL-01 ("never fabricate ... employers"): I judge this an ACCEPTABLE precision/recall tradeoff at the deterministic layer, NOT a High gap, for three concrete reasons: (1) the resume's per-entry company is code-injected from the master resume in the assembler (`services.py:526`, `entry.get('company')`), not model-authored, so the structured employer field cannot be fabricated through this path; the model only authors bullet/summary/title prose. (2) A bare capitalized token is genuinely ambiguous with tools/products/methodologies ("Python", "Docker", "Kafka"), so a non-suffix heuristic would produce frequent false-positive rejections — suffix-anchoring is a defensible precision choice. (3) The LLM judge prompt explicitly fails "unsupported ... employers" (`use_cases.py:647`) as a second gate. The residual risk is a model inventing a *new* employer name inside bullet prose without a suffix; that survives the deterministic detector and relies on the judge.

The concern that keeps this at Medium rather than Low: the codepath documents itself as the prompt-independent never-fabricate gate, and `test_detector_flags_fabricated_title_and_employer` only exercises the suffixed case — there is no test or doc asserting the bare-name limitation is intentional, so a future reader may assume coverage that does not exist.
Reproduction: Reproduced as above + reasoned from the assembler injection path.

Required fix (low-cost): Document the suffix-anchored employer limitation explicitly here as a known, intentional precision tradeoff (and note the company field is code-injected so the structured employer is not model-fabricable), and add a regression test that pins the intended behavior (suffixed employer flagged; bare-name employer deliberately not flagged at this layer; covered by the judge). If broader recall is wanted later, consider grounding capitalized multi-word proper nouns that follow "at/with/for" against the corpus while excluding the known tool/skill vocabulary.

workers/automation/src/jobhunter/domain/materials/provenance_builder.py
ebarti: PR Fixer - claude-opus-pr-fixer

Fixed in 8ae44e5. The executive-profile provenance row now selects its `generated_text` through the SAME policy gate the assembler uses (`services.py` `assemble_resume_text`): `shipped_summary = proposed_summary if allow_summary_rewrite else baseline_summary`, via `get_tailoring_policy(profile)["allow_summary_rewrite"]`. When `allow_summary_rewrite=False` the row anchors to the baseline (the line the resume actually ships), the transform resolves to VERBATIM (baseline == shipped) instead of a phantom REFRAME, and `_served_requirements` / the detector now scan only the shipped text. Regression fixture `test_executive_provenance_anchors_to_shipped_summary_when_rewrite_disabled` asserts `provenance.generated_text == _assembled_summary_line(...)` (extracted from the real `ResumeAssembler` output) and that the dropped rewrite (`99.99%`) appears in no scanned bullet; `test_executive_provenance_anchors_to_rewrite_when_rewrite_enabled` pins the rewrite-on arm.

workers/automation/src/jobhunter/domain/materials/fabrication_detector.py
ebarti: PR Fixer - claude-opus-pr-fixer

Fixed in 8ae44e5. `has_numeric` no longer grounds on corpus-wide digit-run membership. New `_normalize_numeric(token)` canonicalises each numeric token to a `kind:value` key, where `kind` is one of `pct` / `money` / `mult` / `bare` and `value` is magnitude-resolved (k/m/b/million/billion applied; commas treated as thousands separators). The corpus stores `numeric_keys` (full normalized tokens) instead of `numeric_digits`, and a bullet number grounds only when a profile number of the SAME kind+magnitude exists. Verified directly: with a profile whose only number is `35% latency reduction` (`numeric_keys={pct:35}`), `$35M` -> `money:35000000` and `35 million` -> `bare:35000000` are now flagged, while the real `35%` (`pct:35`) still grounds. Equivalent renderings of one quantity still collapse (`$1.2M` / `$1.2 million` / `$1,200,000` -> `money:1200000`) so clean resumes are not over-flagged. Fixtures: `test_detector_flags_digit_colliding_fabrication_with_different_unit` (the $35M-vs-35% case) and `test_detector_grounds_equivalent_money_renderings`.

workers/automation/src/jobhunter/domain/materials/fabrication_detector.py
ebarti: PR Fixer - claude-opus-pr-fixer

Fixed in 8ae44e5. Documented the suffix-anchored employer check as a KNOWN, INTENTIONAL precision-over-recall limitation directly above `_EMPLOYER_RE`, with the three concrete reasons from your assessment: (1) the per-entry employer is code-injected from the master resume in the assembler (`entry.get("company")`), never model-authored, so the structured field cannot be fabricated through this path; (2) bare capitalised tokens are ambiguous with tools/products ("Python", "Docker", "Kafka"), so a non-suffix heuristic would over-flag prose; (3) the LLM judge is a second prose-aware gate that fails unsupported employers. Added a regression test `test_detector_defers_bare_name_employer_to_judge` pinning the contract: a suffixed fabricated employer ("Globex Corporation") is flagged here, while a bare-name one ("Owned the API at Netflix.") is deliberately NOT flagged at this layer (deferred to the judge). Did not attempt to flag all bare names, per your guidance.

PR #146, opened 2026-06-09
fix(materials): make Codex analysis leg work live (strict schema + enum status)

ebarti:
## What
Two live-only bugs in the Codex ensemble leg, found by a real end-to-end smoke test (mocked unit tests passed but never hit the real SDK):

1. **Strict JSON Schema** — Codex/OpenAI `output_schema` rejected Pydantic's `model_json_schema()` with `400 invalid_json_schema: 'additionalProperties' is required to be supplied and to be false`. Added `infrastructure/analysis/strict_schema.py::strict_json_schema` (ported from the mestre reference) that recursively sets `additionalProperties:false`, lists all props in `required`, and makes optionals nullable; applied it in the Codex adapter.
2. **TurnStatus enum** — the adapter compared `str(result.status) != "completed"`, but the SDK returns a `TurnStatus` enum whose `str()` is `"TurnStatus.completed"`, so every successful turn was wrongly rejected. Now compares `status.value`.

## Why
The 2-SDK ensemble (Claude + Codex) is the flagship Phase-1 path; the Codex leg was silently failing live (degraded to Claude-only).

## How validated
- New unit tests: `test_strict_schema.py` (strict invariants, optional→nullable, idempotent) and `test_codex_adapter_status.py` (enum status accepted/rejected — the mock now uses an enum-like status to pin the regression).
- **Live end-to-end** (real Claude + Codex sessions): both legs now produce grounded drafts; canonical analysis has every evidence span as a literal JD substring (grounding PASS), EEO PASS, both-legs PASS.
- Full worker suite: 1283 passed, 1 pre-existing unrelated failure (`test_suppress_backfilled_legacy_job_*`); ruff clean.

## Docs
No doc update warranted — bug fix restoring the documented 2-SDK behavior; no public contract/CLI/UI change.

### DIFF
--- a/workers/automation/src/jobhunter/infrastructure/analysis/codex_analysis_adapter.py
+++ b/workers/automation/src/jobhunter/infrastructure/analysis/codex_analysis_adapter.py
@@ -24,6 +24,7 @@ from jobhunter.domain.materials.analysis import (
     JobAnalysis,
     JobAnalysisDraft,
 )
+from jobhunter.infrastructure.analysis.strict_schema import strict_json_schema
 
 AsyncCodexFactory = Callable[[], Any]
 
@@ -79,10 +80,17 @@ class CodexAnalysisAdapter:
             )
             result = await thread.run(
                 prompt,
-                output_schema=JobAnalysis.model_json_schema(),
+                # Codex/OpenAI strict structured output requires every object to
+                # set additionalProperties:false and list all props in required;
+                # Pydantic's model_json_schema() emits neither (live 400 otherwise).
+                output_schema=strict_json_schema(JobAnalysis.model_json_schema()),
                 effort="high",
             )
-        status = str(getattr(result, "status", "") or "")
+        # ``status`` is a ``TurnStatus`` enum whose ``str()`` is
+        # "TurnStatus.completed" — compare its ``.value`` ("completed"), not the
+        # enum repr, or a successful turn is wrongly rejected (live-caught bug).
+        status_obj = getattr(result, "status", None)
+        status = str(getattr(status_obj, "value", status_obj) or "")
         final_response = getattr(result, "final_response", None)
         if (status and status != "completed") or not final_response:
             error = getattr(result, "error", None)
--- a/workers/automation/src/jobhunter/infrastructure/analysis/strict_schema.py
+++ b/workers/automation/src/jobhunter/infrastructure/analysis/strict_schema.py
@@ -7,17 +7,27 @@ plain Pydantic ``model_json_schema()`` with::
     'additionalProperties' is required to be supplied and to be false.
 
 OpenAI strict structured output requires every object node to set
-``additionalProperties: false`` AND to list *every* property in ``required``.
-Pydantic v2 emits neither by default. :func:`strict_json_schema` rewrites the
-schema to satisfy strict mode while preserving "optional" semantics by making
-non-required properties accept ``null`` (so the parsed object still validates
-against the original Pydantic model, where those fields are ``X | None`` /
-defaulted).
+``additionalProperties: false`` AND to list *every* property in ``required``
+(strict mode has no notion of an absent optional property). :func:`strict_json_schema`
+rewrites the schema to satisfy that.
 
-Mirrors the canonical helper in mestre (``mestre/structured_output.py::
-strict_json_schema``). Pure dict manipulation — recurses into ``$defs`` and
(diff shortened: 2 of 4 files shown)


workers/automation/src/jobhunter/infrastructure/analysis/codex_analysis_adapter.py
```diff
@@ -79,10 +80,17 @@ async def draft(self, system_prompt: str, jd_snapshot: str) -> JobAnalysisDraft:
             )
             result = await thread.run(
                 prompt,
-                output_schema=JobAnalysis.model_json_schema(),
+                # Codex/OpenAI strict structured output requires every object to
+                # set additionalProperties:false and list all props in required;
+                # Pydantic's model_json_schema() emits neither (live 400 otherwise).
+                output_schema=strict_json_schema(JobAnalysis.model_json_schema()),
```
ebarti, line 86: PR Reviewer - codex-strict-schema-reviewer

Severity: High
Gate: Fail
Evidence: `strict_json_schema` makes every optional property nullable + required. But four `JobAnalysis` optional fields are non-nullable in the model: `ReasonedKeyword.rationale: str = ""`, `ReasonedKeyword.is_orphan: bool = False`, `JobAnalysis.requirements: list = default_factory`, `JobAnalysis.keywords: list = default_factory` (`domain/materials/analysis.py:84-85,135-136`). The strict schema sent to Codex therefore advertises `rationale/is_orphan: null` and `requirements/keywords: null` as valid, but `JobAnalysis.model_validate_json(final_response)` (line 98, uncaught) rejects those nulls.

Reproduction: Reproduced with the worktree code via `uv ... python`:
- Strict-schema validation of `{... "keywords":[{"keyword":"Go","evidence_span":"Go","requirement_ref":null,"rationale":null,"is_orphan":null}], "requirements":[...]}` -> `NONE (payload is schema-valid)`.
- Same payload through `JobAnalysis.model_validate_json(...)` -> `ValidationError`: `keywords.0.rationale Input should be a valid string`, `keywords.0.is_orphan Input should be a valid boolean`. (`requirements: null` / `keywords: null` likewise fail with `list_type`.)

The strict_schema.py docstring (and PR body) explicitly claim the optional->nullable rewrite means "the parsed object still validates against the original Pydantic model" — that invariant is false for this model. The mestre reference works only because its hand-authored schemas are co-designed with genuinely-nullable model fields (see `mestre/structured_output.py` PLANNER_SCHEMA etc.); porting the transform onto a model whose optionals are non-nullable + defaulted breaks the round-trip.

Impact: This is the same live-only failure class this PR exists to fix. When Codex emits `null` for any of these strict-permitted fields (plausible for `rationale` free text, and for `is_orphan`, which the model has no business setting — it's computed by `_flag_orphans`), `_draft_with_retry` (ensemble.py:77) catches the `ValidationError`, retries up to 2x against the same permissive schema, then records an `AnalysisFailure` and the ensemble degrades to Claude-only (1/2) — silently dropping the Codex leg. The single live smoke run cannot rule this out because the output is stochastic; the new `test_strict_schema.py` only asserts schema shape and never round-trips a strict-valid null payload through `model_validate_json`, so the mocked suite misses it too.

Required fix: Make the schema the adapter sends and the model it parses agree on nullability. Lowest-risk option: in the adapter, before `model_validate_json`, drop top-level/nested keys whose value is `None` so Pydantic applies the field defaults (json.loads -> recursively strip None -> `model_validate`). Alternatively, make the four optional fields genuinely nullable + normalize via a `mode="before"` validator (coerce `None` -> default), or have `strict_json_schema` only nullable-ize properties whose model field actually accepts null. Add a regression test that feeds a strict-schema-valid payload with `null` for `rationale`/`is_orphan`/`requirements`/`keywords` through the Codex adapter (or `model_validate_json`) and asserts it parses to the defaulted values.

ebarti:
Addressed the High (strict-schema/model nullability mismatch): `strict_json_schema` no longer force-nullables non-required fields — it only adds all props to `required` + `additionalProperties:false`, preserving each property's declared type. `requirement_ref` (str|None) stays nullable; `rationale`/`is_orphan`/lists stay non-null, so Codex is never told it may emit null for them and the payload round-trips through `JobAnalysis`. Added `test_defaulted_nonnull_field_is_required_but_not_nullable` + `test_strict_valid_payload_round_trips_through_model`. Re-verified live: both legs ground (grounding PASS, EEO PASS).

PR #151, opened 2026-06-10
feat(web): add shadcn token foundation

ebarti:
## Summary

Adds the shadcn standard token foundation for the web app and migrates the shared primitive token base onto the CSS-first semantic token contract.

## Changes

- Configures shadcn for `apps/web` with local aliases and standard semantic token usage.
- Replaces the Tailwind config bridge with CSS-first tokens in the web styles.
- Migrates core shared primitives, overlays, menus, and Storybook support away from legacy token utilities.
- Adds token contract regression coverage and a browser smoke test for light, dark, and density behavior.
- Records Phase 6 planning, review, and verification artifacts.

## Validation

- `corepack pnpm --filter @jobhunter/web test src/styles/token-contract.test.ts`
- `corepack pnpm web:check`
- `corepack pnpm web:build`
- `corepack pnpm dlx shadcn@latest info -c apps/web`
- `corepack pnpm --filter @jobhunter/web e2e -- tests/token-foundation.spec.ts`
- shared UI legacy-token scan: passed with 0 matches
- generated CSS proof for semantic token utilities: passed
- `git diff --check`

Note: `web:build` completed with the existing Vite large chunk warning.


ebarti (commented):
Token foundation migration is thorough and verifies clean locally (`web:check`, `web:build`, and `token-contract.test.ts` all pass; no dangling references to the removed `--bg/--paper/--ink/--rule/--soft/--danger/--warn/--ok/--info` vars or their old Tailwind utilities anywhere in `apps/web`). One visual regression on the pipeline progress bar is noted inline.

apps/web/src/styles/globals.css
```diff
@@ -521,29 +526,29 @@ textarea {
...
 }
 
 .seg-running {
-  background: var(--info);
+  background: var(--status-info);
 }
 
 .seg-pending {
-  background: var(--soft);
+  background: var(--muted);
```
ebarti, line 551: `.seg-pending` now resolves to `var(--muted)`, which is the exact same value as the `.bar` track background (`.bar { background: var(--muted) }`, line 529). The "pending" segment of `SegmentBar` (rendered in production by `views/dashboard/Funnel.tsx` as `["pending", stage.pending]`) therefore becomes indistinguishable from the unfilled track — it visually disappears between the colored done/failed/blocked/running segments.

On `main` this segment used `--soft` (a distinct mid-gray), so pending was visible. The legacy `--soft` token was a mid-tone used both as a subtle surface and as a foreground swatch; everywhere it was used as a *foreground/swatch* in this PR it was correctly migrated to `--muted-foreground` (e.g. `.kpi-lbl` color), and only the *background-surface* usages went to `--muted`. This swatch case looks like it took the surface mapping by mistake — `var(--muted-foreground)` would restore a visible, distinct pending segment. (The token-contract test and `token-foundation.spec.ts` assert token presence/density/theme but don't exercise the segment bar, so this isn't caught by the suite.)

PR #152, opened 2026-06-10
feat(web): harden shared primitives for standard tokens

ebarti:
## Summary

Stacked on #151 (`phase-06-token-foundation-shadcn-preset`).

- Hardens shared shadcn/Radix primitives for the standard semantic token baseline, including overlay/menu stories, form/action/layout states, and shared primitive QA documentation.
- Repairs shared table activation semantics by replacing focusable structural rows with named native row activation buttons while preserving table/row roles.
- Keeps `shared/ui` domain-agnostic by moving the markdown description helper into `shared/lib` and recording zero forbidden boundary imports.
- Reconciles Phase 7 planning, review, and verification artifacts after the review/fix loop.

## Validation

- `corepack pnpm web:check` - passed
- `corepack pnpm --filter @jobhunter/web test src/shared/ui/data-table.test.tsx src/shared/ui/toast.a11y.test.tsx src/shared/ui/filterable-data-grid.test.tsx src/shared/ui/table-pager.test.tsx src/shared/lib/job-description-blocks.test.ts` - passed, 5 files / 19 tests
- `corepack pnpm web:storybook:build` - passed
- `corepack pnpm web:storybook:test` - sandboxed run failed before suites with Chromium MachPort permission; browser-launch escalation passed 89 suites / 320 tests
- `corepack pnpm dlx shadcn@latest info -c apps/web` - passed
- Corrected legacy-token scanner over shared primitives/story/config surfaces - 0 matches
- Shared/ui boundary scan for context/view/API/query/SSE/local-storage/platform imports - 0 matches
- Storybook a11y deferral scan - 10 tracked story deferrals
- `git diff --check` - passed


apps/web/src/shared/ui/filterable-data-grid.tsx
```diff
@@ -622,36 +628,50 @@ export function FilterableDataGrid<TData>({
...
+              const baseRowClassName = rowClassName?.(row);
+              const effectiveRowClassName = onRowActivate
+                ? [baseRowClassName, "data-grid-row-activatable"].filter(Boolean).join(" ")
+                : baseRowClassName;
+              const activationLabel = rowActivationLabel?.(row) ?? `Open row ${rowId}`;
               return (
                 <tr
                   key={rowId}
-                  className={rowClassName?.(row)}
+                  className={effectiveRowClassName}
```
ebarti, line 639: [High] This removes the row-level `onClick`/`onKeyDown` (base branch had `onClick={() => onRowActivate?.(row)}` on the `<tr>`); activation now fires only from the small left-aligned `.data-grid-row-activation-button` (`min-width:30px`, `margin-left:8px`). Two existing Playwright specs still open drawers by clicking the row body and were NOT updated, so they regress:

- `apps/web/e2e/tests/jobs-drawer.spec.ts:13` — `await row.click()` on `table.jobs-data-grid-table tbody tr`, then expects `/jobs/.+` + the Job details dialog. Playwright clicks the row center, which is not over the left-aligned Open button, so navigation no longer fires.
- `apps/web/e2e/tests/materials.spec.ts:68` — same `row.click()` to open the Job details drawer for the Generate-materials flow.

The unit tests here were switched to `getByRole('button', { name: 'Open ...' })` (and assert ArrowDown no longer activates), confirming the row-click affordance was intentionally dropped — but the e2e specs that exercise the real product path must be updated to click the named Open button (or whole-row click restored). The PR validation only ran 5 scoped Vitest files; `web:check` won't catch this, and the new Shared Primitive QA Gate in `docs/local-reliability-qa.md` waives only the broad Vitest run, not the e2e suite. Separately, `.data-grid-row-activatable:hover` / `.table-row-activatable:hover` still highlight the entire row on hover, which now falsely signals the whole row is clickable.

ebarti (commented):
Reviewed the shared-primitive token hardening relative to phase-06. The token/role/a11y work and doc reconciliation look solid (a11y deferral count correctly drops 13->10, QA gate documented). One real regression: moving row activation from the whole-row onClick to a small inline 'Open' button breaks two existing Playwright specs (jobs-drawer, materials) that still click the row body, and these specs were not updated. Details inline.

apps/web/src/shared/ui/filterable-data-grid.tsx
ebarti: Fixed in 2504638.

**e2e regression (the actionable part): fixed.** Both specs now open the Job details drawer via the named per-row activation button rather than a whole-row click:

```ts
await row
  .getByRole("button", { name: /^Open job Director of Platform Engineering/ })
  .click();
```

- `apps/web/e2e/tests/jobs-drawer.spec.ts`
- `apps/web/e2e/tests/materials.spec.ts`

This matches the deliberate CR-01 design (commit `cc77e36`, documented in `07-REVIEW.md` / `07-PRIMITIVE-AUDIT.md` / `07-VERIFICATION.md`): activation moved off the focusable structural `<tr>` onto a named native `<button>`, and the unit test asserts the row is non-activatable except via that button (ArrowDown does not activate). Restoring whole-row `onClick` would regress that accessibility fix, so updating the specs is the correct direction. The button's accessible name is `Open job <title> at <company>`; the regex is anchored on the unique seeded title and is the only `Open job ...` button in the row, so the selector is unambiguous.

Verified: `pnpm web:check` passes (its tsconfig includes `e2e/**/*.ts`, so the new Playwright API usage type-checks), the scoped Vitest set passes (5 files / 19 tests), and `playwright test --list` collects both specs. I could not execute the full Playwright run here — it needs the seeded dev stack (API + Vite + worker + SQLite global-setup) which is not available in this sandboxed worktree.

**Re the "Separately, row hover still highlights the whole row" point — intentionally left as-is, not a regression to fix here.** The hover uses `background: var(--muted)` (neutral), which `07-UI-SPEC.md` explicitly permits for list/row affordances ("neutral `accent` hover is allowed for menu/list affordances"). A subtle full-row tint is a standard row-scanning affordance for dense data grids even when a sub-element is the action target, and it is token-compliant. Removing or re-scoping it is a separable interaction-design decision that exceeds this token-migration PR's scope; if you'd like the hover affordance constrained to the activation button, I'd track that as its own UI change rather than fold it in here.

ebarti:
Addressed the review feedback in 2504638.

The one actionable finding (the jobs-drawer / materials Playwright specs still opening the Job details drawer with a whole-row click, which no longer activates after the CR-01 move to a named per-row Open button) is fixed: both specs now click `getByRole('button', { name: /^Open job .../ })` within the row, matching the intended activation affordance. Verified with `pnpm web:check` (e2e specs are in its tsconfig include), the scoped Vitest set (5 files / 19 tests), and `playwright test --list` (both specs collect). The full Playwright run was not executed here — it requires the seeded dev stack (API + Vite + worker + SQLite global-setup) unavailable in this sandboxed worktree.

The secondary 'whole-row hover still highlights' observation is left as-is with rationale in the resolved inline thread: the neutral `--muted` row hover is explicitly permitted for list affordances by 07-UI-SPEC.md and re-scoping it is a separable interaction-design change beyond this token-migration PR.

PR #195, opened 2026-06-26
docs: add synthetic screenshot workflow

ebarti:
## Summary
- Add `pnpm docs:screenshots` as a repeatable Playwright screenshot workflow.
- Extend the synthetic QA seed with current material artifact schema, HTML resume preview data, and layout anchors so Apply Review screenshots render without touching real local data.
- Add public PNG screenshots, a user screenshot guide, and a developer screenshot playbook.
- Hide TanStack devtools for screenshot-mode Vite runs while leaving normal development unchanged.

## Stack
Base: #194
Next: #196

## Validation
- `corepack pnpm api:check`
- `corepack pnpm web:check`
- `corepack pnpm api:test`
- `corepack pnpm docs:screenshots`


ebarti (commented):
Reviewed the screenshot workflow against its base `worktree/8143`. The substantive code/config changes are correct:

- **Devtools gating is safe.** `showDevtools = import.meta.env.DEV && import.meta.env.VITE_JOBHUNTER_HIDE_DEVTOOLS !== "1"` (`apps/web/src/routes/__root.tsx`) leaves normal `pnpm dev`/`web:dev` unchanged (var unset → renders devtools as before), hides them only when the screenshot web server sets the var to `"1"`, and constant-folds to `false` in production (`DEV` is statically false), so devtools stay tree-shaken out of prod. No throw when unset.
- **No real-data risk.** `docs:screenshots` runs the existing Playwright harness, whose `global-setup.ts` wipes+recreates the disposable `JOBHUNTER_E2E_APP_DIR` and seeds it via `qa-seed.ts`; the API is launched with `JOBHUNTER_DIR`/`JOBHUNTER_DB_PATH` pointed at that disposable dir, never `~/.jobhunter`. Seed data is synthetic (`QA Candidate`, `qa@example.local`, fabricated greenhouse URL).
- **Seed schema matches production.** The rewritten `job_materials`, `job_materials_artifacts` (PK `(job_url, generation, artifact_type)`, `render_format`, `superseded_at`) and new `job_material_layout_boxes` DDL match the canonical DDL in `apps/api/src/resume-review-drafts.ts` column-for-column, and the projection/read-model SELECTs. The `resume_pdf` row uses `render_format: "html_pdf"` + `metadata_json.html_path` pointing at the same-basename sibling `gitlab-platform-resume.html`, which satisfies the strict `metadataHtmlPath === siblingHtmlPath` invariant in `server.ts htmlPreviewForArtifact`, so the Apply Review HTML preview resolves.
- Screenshot spec selectors/routes all exist (`apply-review` route + `aria-label="Application review queue"`, the named headings, `jobs-data-grid-table`/`runs-data-grid-table`). `docs:screenshots` filter resolves to exactly the one new spec.

One minor (non-blocking) consistency note below; nothing else actionable.

Low — `apps/web/src/vite-env.d.ts`: the new `VITE_JOBHUNTER_HIDE_DEVTOOLS` env var consumed in `__root.tsx` isn't added to the `ImportMetaEnv` interface, unlike the other two `VITE_*` vars declared there. It won't break `web:check` (Vite's `ImportMetaEnv` has a `[key: string]: any` index signature), but declaring `readonly VITE_JOBHUNTER_HIDE_DEVTOOLS?: string;` would keep the typed env contract consistent and self-documenting. (Couldn't anchor inline — `vite-env.d.ts` isn't in this diff.)

PR #205, opened 2026-07-01
fix(materials): require word-boundary match in employer-analysis grounding gate

ebarti:
## What

Tighten the deterministic evidence-span **grounding gate** (`workers/automation/src/jobhunter/domain/materials/analysis_grounding.py`) so a located match must sit on **whole-token boundaries**. A hallucinated short keyword can no longer ground *inside* an unrelated word.

## Why (root cause)

The grounding validator is the cardinal no-fabrication gate — the one invariant JSON Schema cannot enforce: every `evidence_span` MUST be locatable in the persisted JD snapshot. The locator built a per-character tolerant regex and ran `re.search` with **no token boundary**, so a short span matched inside a longer word:

- `"Go"` grounded against `"goals"` (snapped to the mid-word fragment `"go"`)
- `"AI"` grounded against `"email"` / `"detail"`
- `"R"` / `"C"` grounded against almost anything

A fabricated keyword thus passed the gate, and snap-to-source then rewrote the stored `evidence_span` to a mid-word fragment — persisting/displaying it as "verbatim proof" and feeding the spurious keyword into tailoring's ATS targets. Correctly-absent tokens like `"ML"`/`"SRE"` were rejected, masking the hole.

## How

Bracket the tolerant locator pattern with zero-width **token-boundary lookaround** at the OUTER edges only:

```
(?<![^\W_]) <tolerant-pattern> (?![^\W_])
```

- `[^\W_]` is exactly `str.isalnum()` (Unicode-aware). Deliberately **not** `\b`, which treats `_` as a word char; the invariant is "the match is not glued to another alphanumeric".
- Zero-width + outer-edge only: **multi-word spans still match across internal whitespace**; only the first/last real character must border a non-alphanumeric (or the string edge).
- Applied against the ORIGINAL JD, so **all formatting tolerance** (whitespace runs, Unicode hyphen/dash, smart quotes, case, Markdown-escape) and **snap-to-source** are preserved.
- Because the assertions live in the pattern, `re.search` **skips a sub-word occurrence and keeps scanning** for a boundary-aligned one (e.g. `"goals … Go"` still grounds and snaps to the standalone `"Go"`, never `"go"`).

Hyphenated compounds still ground correctly (`"AI"` in `"AI-native"` — the hyphen is a non-alphanumeric boundary).

## Tests

New regression class `TestTokenBoundaryGrounding` in `workers/automation/tests/test_employer_analysis_model.py` (10 cases) pinning the exact invariant: sub-word rejection (`"Go"` vs `"goals"`, `"AI"` vs `"email"/"detail"`, single-letter `"R"`/`"C"`), standalone-token grounding + case-preserving snap, hyphen-boundary grounding, multi-word span snap across a whitespace run, sub-word-before-standalone snap-to-real-token, and an end-to-end `ground_and_snap` rejection of a keyword grounded only by a sub-word.

## Validation

- `pytest workers/automation/tests/test_employer_analysis_ensemble.py test_analyze_job_use_case.py test_employer_analysis_model.py` → **68 passed**
- `pytest` (full worker suite) → **1496 passed** (Temporal gRPC connection warnings are unrelated env noise; no test failed)
- `ruff check` on the changed source and test files → **All checks passed**

## Docs

Updated `docs/architecture.md` grounding-gate description to state the whole-token-boundary requirement — the doc explicitly claims spans whose WORDS are absent are rejected, which the sub-word hole silently violated. No other doc owns this behavior; no README/CLI/API surface changed.

### DIFF
--- a/workers/automation/tests/test_employer_analysis_model.py
+++ b/workers/automation/tests/test_employer_analysis_model.py
@@ -420,6 +420,43 @@ class TestTokenBoundaryGrounding:
             ground_and_snap(bad, jd)
         assert [(v.kind, v.ref_id) for v in exc.value.violations] == [("keyword", "Go")]
 
+    # --- Hardening: lock in the exact boundary semantics the fix relies on so a
+    # future switch to ``\b`` (treats "_" as a word char) or adding ``re.ASCII``
+    # (would stop counting non-ASCII letters/digits as alphanumeric) cannot
+    # silently reopen the sub-word hole. All pass against the current code. --- #
+
+    def test_digit_span_does_not_ground_inside_a_larger_number(self) -> None:
+        # A number is a token too: "5" must not ground inside "15", nor "24"
+        # inside "2024". Digits are alphanumeric boundaries just like letters.
+        assert is_grounded("5", "We had 15 incidents last year.") is False
+        assert is_grounded("24", "Founded in 2024 by two engineers.") is False
+
+    def test_digit_span_grounds_as_standalone_token(self) -> None:
+        jd = "We need 5 years of experience."
+        assert is_grounded("5", jd) is True
+        assert locate_grounded_span("5", jd) == "5"
+
+    def test_span_grounds_at_absolute_start_and_end_of_snapshot(self) -> None:
+        # The lookaround must treat the string edge as a boundary: a span flush
+        # against position 0 or against the final character still grounds.
+        start_jd = "Go is our primary backend language."
+        end_jd = "Our primary language is Rust"  # no trailing punctuation
+        assert is_grounded("Go", start_jd) is True
+        assert locate_grounded_span("Go", start_jd) == "Go"
+        assert is_grounded("Rust", end_jd) is True
+        assert locate_grounded_span("Rust", end_jd) == "Rust"
+
+    def test_letter_glued_span_is_rejected_including_non_ascii_glue(self) -> None:
+        # Glued to another letter -> part of a larger word -> rejected; bordered
+        # by a hyphen (non-alphanumeric) -> a whole token -> grounds. The
+        # non-ASCII cases pin the Unicode-aware boundary (str.isalnum), so an
+        # ASCII-only class would treat the accented letter as a non-boundary and
+        # wrongly ground the sub-word here.
+        assert is_grounded("AI", "We prefer naiveAI approaches here.") is False
+        assert is_grounded("AI", "We invest in AI-native tooling.") is True
+        assert is_grounded("AI", "Legacy stack AIür module.") is False  # U+00FC glue
+        assert is_grounded("Go", "The Goéland project.") is False  # U+00E9 glue
+
 
 # --------------------------------------------------------------------------- #
 # Cache key (D-11/D-12)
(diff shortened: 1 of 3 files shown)


workers/automation/tests/test_employer_analysis_model.py
```diff
@@ -334,6 +334,93 @@ def test_ground_and_snap_preserves_draft_model_id(self) -> None:
...
+# boundary, so a short fabricated keyword grounded INSIDE an unrelated word
+# ("Go" in "goals", "AI" in "email"/"detail", "R"/"C" in almost anything) and
+# snap-to-source then rewrote the stored evidence to a mid-word fragment ("go"),
+# slipping a hallucinated keyword past the one gate JSON Schema cannot enforce.
+# The fix requires the located match to border a non-alphanumeric (or the string
+# edge) at both OUTER edges of the whole (possibly multi-word) span.
+# --------------------------------------------------------------------------- #
+
+
+class TestTokenBoundaryGrounding:
```
ebarti, line 350: PR Reviewer - pr-reviewer

Severity: Low
Gate: Pass
Evidence: `TestTokenBoundaryGrounding` pins the letter cases (Go/goals, AI/email, single-letter R/C) but not the other members of the boundary class `[^\W_]`. The invariant now hinges on `[^\W_]` matching digits and Unicode letters (Unicode-aware because no `re.ASCII`), not just ASCII letters. None of that is pinned by a test, so a future switch to `\b` (which also treats `_` as a word char) or adding `re.ASCII` would silently regress the cardinal no-fabrication gate with a green suite. Verified correct against this PR: `is_grounded("5","15 years")`→False, `is_grounded("24","in 2024 we grew")`→False, `is_grounded("Go","Go is our language")`→True (absolute start), `is_grounded("R","R")`→True (whole string), `is_grounded("AI","naiveAI here")`→False (unicode-word glue).
Reproduction: Reproduced with a Python probe importing is_grounded/locate_grounded_span from a worktree at the PR head.

Optional hardening (does not block merge): the behavior is already correct; the gap is only in the regression net for the single most important gate.

Required fix: Add regression cases for (1) digit sub-word rejection ("5" in "15", "24" in "2024") plus a digit standalone, (2) a span at the absolute string start and at the absolute string end (exercises the lookbehind at pos 0 and the lookahead at end-of-string), and optionally (3) a Unicode-letter glue case.

ebarti: Addressed in eedf08c6. Added four hardening cases to `TestTokenBoundaryGrounding` that pin the full `[^\W_]` boundary class beyond ASCII letters:

- digit sub-word rejection (`"5"` in `"15"`, `"24"` in `"2024"`) + a standalone digit (`"5"` in `"5 years"`);
- string-edge boundaries — a span flush at position 0 (`"Go ..."`) and at end-of-string (`"... Rust"`), exercising the lookbehind at pos 0 and the lookahead at EOS;
- letter-glue rejection including **non-ASCII** glue (`"AI"` vs `"naiveAI"` and `"AIür"` U+00FC; `"Go"` vs `"Goéland"` U+00E9) vs hyphen-bordered grounding — this pins the Unicode-awareness so a future `re.ASCII` (or `\b`, which also treats `_` as a word char) would fail here instead of silently regressing.

`pytest workers/automation/tests/test_employer_analysis_model.py` -> 50 passed; ruff clean. The production fix is unchanged (test-only commit).

PR #206, opened 2026-07-01
feat(db): add SQLite backup command and schema-version guard

ebarti:
## What

Adds two additive, non-destructive data-durability safety nets to the local SQLite store (`~/.jobhunter/jobhunter.db`). No schema rewrite, no behavior change to existing paths.

1. **Schema-version guard** — `database.py` now defines `SCHEMA_VERSION` and stamps `PRAGMA user_version` on `init_db`:
   - Pre-guard databases report `user_version == 0` and are **adopted cleanly** (stamped to `SCHEMA_VERSION`).
   - A DB whose version is **newer than the running build** is left untouched (never silently downgraded) and logs a clear warning.
   - The existing idempotent `ensure_*` migrations remain the schema source of truth — this is a lightweight forward-incompatibility guard, not a migration framework.
2. **`jobhunter backup` command** (+ `backup_database` helper) — writes a consistent snapshot with SQLite `VACUUM INTO` to a timestamped file under `~/.jobhunter/backups/jobhunter-<timestamp>.db`. Safe to run while the app is running (WAL-consistent), produces a standalone single-file DB (no `-wal`/`-shm` sidecars), supports `--output <file|dir>`, prints the path, and never deletes anything.

## Why

The entire product state — canonical event log, all job/profile data, artifact pointers — lives in one WAL-mode SQLite file with **no backup/restore/export path** and **no `user_version` marker**. A stale build or an interrupted schema change could touch the DB undetected, the highest-probability real-world data-loss vector for this local-first app. These changes make snapshots trivial and make a version mismatch loud instead of silent.

## Docs

- `README.md`: new "Back Up And Restore" section (backup command + WAL-safe restore steps), and `backups/` added to the local files list.
- `docs/local-reliability-qa.md`: regression row mapping the invariant to `test_database_backup.py`.

## How validated

Run from the worktree.

**New tests + existing DB regression (`test_job_repository.py`):**
```
    workers/automation/tests/test_database_backup.py \
    workers/automation/tests/test_job_repository.py
.....................                                                    [100%]
21 passed in 0.51s
```
Also ran `test_dashboard_projection.py` + `test_projection_builder.py` (init_db-heavy): 32 passed.

**Lint:**
```
    workers/automation/src/jobhunter/database.py workers/automation/src/jobhunter/cli.py
All checks passed!
$ uv --project workers/automation run --extra dev ruff check .   # whole worker package
All checks passed!
$ git diff --check
clean
```

**CLI smoke** (throwaway `JOBHUNTER_DIR`, never touches real `~/.jobhunter`): created a DB with one job row, then:
- `jobhunter backup` → wrote `backups/jobhunter-<ts>.db` (749,568 bytes), printed the path.
- `jobhunter backup --output <file>` → wrote the explicit file.
- Both backups reopened with `user_version=1` and the seeded job row present.
- `backup` shows in `jobhunter --help` and `jobhunter backup --help`.

## Safety

No auto-apply/browser/destructive actions were run. Backup is read-only against the DB; the restore procedure is documented (manual, app stopped) and not automated.

### DIFF
--- a/workers/automation/src/jobhunter/database.py
+++ b/workers/automation/src/jobhunter/database.py
@@ -9,7 +9,6 @@ persistent run and event telemetry.
 """
 
 import json
-import logging
 import sqlite3
 import threading
 from datetime import datetime, timezone
@@ -18,13 +17,16 @@ from typing import Any
 
 from jobhunter.config import DB_PATH, DEFAULTS
 
-log = logging.getLogger(__name__)
-
 # Schema version stamped into the SQLite ``user_version`` header. The ensure_*
 # helpers below are additive and idempotent, so this is a lightweight
-# forward-incompatibility guard (detect a DB written by a newer build), not a
+# forward-incompatibility guard (refuse a DB written by a newer build), not a
 # migration framework. Bump it only when the schema shape changes.
 SCHEMA_VERSION = 1
+
+
+class IncompatibleSchemaVersionError(RuntimeError):
+    """Raised when the database was written by a newer build than this code."""
+
 
 # Thread-local connection storage — each thread gets its own connection
 # (required for SQLite thread safety with parallel workers)
@@ -105,6 +107,9 @@ def backup_database(
 
     src = sqlite3.connect(source)
     try:
+        # Match the app's lock-wait budget so a backup taken during transient
+        # DDL waits briefly instead of failing on "database is locked".
+        src.execute("PRAGMA busy_timeout=10000")
         src.execute("VACUUM INTO ?", (str(destination),))
     finally:
         src.close()
@@ -119,7 +124,7 @@ def _resolve_backup_destination(source: Path, output: Path | str | None) -> Path
         target_dir = candidate
     else:
         target_dir = source.parent / "backups"
-    timestamp = datetime.now().strftime("%Y%m%d-%H%M%S")
+    timestamp = datetime.now(timezone.utc).strftime("%Y%m%d-%H%M%S")
     return target_dir / f"jobhunter-{timestamp}.db"
 
 
@@ -127,20 +132,20 @@ def _ensure_schema_version(conn: sqlite3.Connection) -> int:
     """Stamp ``PRAGMA user_version`` as a lightweight schema guard.
 
     Databases created before this guard report ``user_version == 0`` and are
-    adopted by stamping ``SCHEMA_VERSION``. A database whose version is newer
-    than this build is left untouched -- never silently downgraded -- and a
-    warning is logged so a stale build does not corrupt a forward-migrated file.
+    adopted by stamping ``SCHEMA_VERSION``; an equal version is a no-op. A
+    database whose version is newer than this build fails closed with
+    ``IncompatibleSchemaVersionError`` -- the caller (``init_db``) invokes this
+    before any table creation or migration, so a stale build never runs its
+    potentially destructive ``ensure_*`` migrations against a forward-migrated
+    file, and is never silently downgraded.
     """
     current = int(conn.execute("PRAGMA user_version").fetchone()[0])
     if current > SCHEMA_VERSION:
-        log.warning(
-            "Database schema version %s is newer than this build's schema version %s; "
-            "the database was written by a newer JobHunter. Upgrade JobHunter before "
-            "writing to it, and keep a backup ('jobhunter backup').",
-            current,
-            SCHEMA_VERSION,
+        raise IncompatibleSchemaVersionError(
+            f"database was written by a newer JobHunter build "
+            f"(schema version {current} > code schema version {SCHEMA_VERSION}); "
+            f"upgrade JobHunter or restore a compatible backup ('jobhunter backup')."
         )
-        return current
     if current < SCHEMA_VERSION:
         conn.execute(f"PRAGMA user_version = {SCHEMA_VERSION}")
         conn.commit()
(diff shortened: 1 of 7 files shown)


workers/automation/src/jobhunter/database.py
```diff
@@ -64,6 +73,80 @@ def close_connection(db_path: Path | str | None = None) -> None:
...
+    current = int(conn.execute("PRAGMA user_version").fetchone()[0])
+    if current > SCHEMA_VERSION:
+        log.warning(
+            "Database schema version %s is newer than this build's schema version %s; "
+            "the database was written by a newer JobHunter. Upgrade JobHunter before "
+            "writing to it, and keep a backup ('jobhunter backup').",
+            current,
+            SCHEMA_VERSION,
+        )
+        return current
```
ebarti: PR Reviewer - pr-reviewer

Severity: Medium
Gate: Pass
Evidence: `_ensure_schema_version` (database.py:135-143) logs a warning and `return current` when the on-disk version is newer than this build, but `init_db` (database.py:178) discards that return value and unconditionally proceeds to run every `ensure_*` migration and CREATE TABLE. Some of those are shape-detecting and destructive on the newer DB path -- e.g. `ensure_market_compensation_tables` (database.py:379-393) RENAMEs/DROPs `job_market_compensation_estimates` when its detection substrings ("total_compensation", "market_baseline_fallback") are absent. So a stale build does not just observe a forward-migrated file, it writes to and can restructure it within the same `init_db` call.
Reproduction: Reasoned from code + confirmed the newer-version test only asserts the version number is preserved (test_database_backup.py::test_newer_schema_version_is_not_downgraded_and_warns); it does not assert init_db aborts. Verified init_db continues (all init_db-heavy tests pass with the guard in place).

The docstring claims the warning exists "so a stale build does not corrupt a forward-migrated file," but the warning is emitted mid-`init_db`, after which the migrations run anyway -- the user cannot act on it in time to prevent the write. Today SCHEMA_VERSION==1 so no real "newer" DB exists and there is no present-day harm; this is a forward-facing correctness/expectation gap in a durability-focused change.

Required fix: Either (a) fail closed -- when `current > SCHEMA_VERSION`, raise from `init_db` before any CREATE TABLE / ensure_* runs so the stale build refuses to touch the newer DB (recommended for the stated "does not corrupt" guarantee); or (b) if warn-and-continue is the intended single-user behavior, soften the docstring/PR wording to "warns" and stop claiming corruption prevention. A test should pin whichever contract is chosen.

ebarti: Fixed in 3ebdfce. The guard now fails closed: `_ensure_schema_version` raises `IncompatibleSchemaVersionError` when the on-disk `user_version` exceeds `SCHEMA_VERSION`, and `init_db` calls it before any `CREATE TABLE`/`ensure_*` migration, so no potentially destructive migration runs against a forward-migrated DB, and the version is never downgraded. `jobhunter backup` does not call `init_db`, so it remains available to snapshot a newer DB before upgrading. Added two regression tests: an empty newer-stamped DB raises before any table is created, and a populated newer-stamped DB raises without downgrade or data loss.

ebarti:
### Appended fix: B5b (Low) — TS API now enforces the schema-version guard

**Finding:** `_ensure_schema_version` only ran in the Python worker's `init_db`. The TypeScript API never read `PRAGMA user_version`, so a forward-migrated DB (`user_version` > the code's supported version) was refused only by the worker — the API would happily open and additively write a newer schema it doesn't understand.

**Fix (`apps/api/src/db.ts`):**
- Added `SUPPORTED_SCHEMA_VERSION = 1`, mirroring the worker's `SCHEMA_VERSION` with a lockstep comment (bump both together).
- Both `openDatabase` and `openReadOnlyDatabase` now read `PRAGMA user_version` and throw `IncompatibleSchemaVersionError` (fail closed) when it exceeds the supported version. `user_version` 0 (pre-guard) and an exact match open exactly as before.
- The API **never stamps** `user_version` — the Python worker stays the single writer of the schema marker (stated in a comment and asserted by a test).

**Test (reproduce-then-fix): `apps/api/test/schema-version-guard.test.ts`**
- Written first; confirmed red on the prior tip (`3ebdfce`): a `user_version = 999` DB opened without throwing (`AssertionError: expected [Function] to throw an error`).
- After the fix it is green. Cases: `SUPPORTED_SCHEMA_VERSION + 1` throws for both open helpers; `user_version` 0 and `== supported` open fine; the API leaves `user_version` untouched (still 0 after open).

**Docs:** narrowed the `docs/local-reliability-qa.md` regression-matrix row (the only doc referencing the guard) to note the API path and cite the new test; it previously implied worker-only coverage.

**Verification (from the worktree root):**
- `pnpm api:check` → pass (tsc, exit 0).
- `pnpm api:test` → pass (18 files, 289 tests).

Python is untouched; PR base unchanged.

ebarti:
### Docs follow-up: restore-path hardening (closes the #207-review cross-PR note)

The #207 re-review flagged a Low cross-PR interaction: #207 makes the projection watermark monotonic (MAX semantics), which is only provably safe with #206's restore when restore is a **full-file** replacement. A partial reconstruction (hand-importing tables, or resetting `sqlite_sequence` below a preserved watermark) could leave the watermark ahead of the restored `job_events` and wedge projection refresh.

Verified the restore procedure #206 documents is already a full-file replacement (`cp` a backup over `jobhunter.db`). Tightened the README "Back Up And Restore" section to make the invariant explicit:
- Restore must replace the entire database file — never hand-import individual tables from a backup.
- If the DB is ever rebuilt piecemeal, delete the `operations_projections` row from `event_watermarks` so projections rebuild from scratch (exact `sqlite3` command included).

Docs-only. `28968891`. `git diff --check` clean; no suite run needed for a docs change.

PR #207, opened 2026-07-01
fix(projections): Python builder writes score criteria/trace/correction audit columns

ebarti:
## What

Make the Python `ProjectionBuilder` write the three score-audit columns
`score_criteria_json`, `score_trace_json`, and `score_correction_json` into the
`job_list_projections` and `job_detail_projections` tables, mirroring the TS
`refreshProjections` builder exactly, AND backfill those columns for jobs that
were scored before this fix. TS side is unchanged.

Changed:
- `SCORE_EVIDENCE_COLUMNS` registry + `CREATE TABLE` statements for both
  projection tables (`sqlite_projection_store.py`), plus a new tiny
  `projection_backfills` marker table.
- `JobListProjection` / `JobDetailProjection` dataclasses (`projections.py`).
- Both upserts in `SqliteProjectionStore` (column list, placeholders, `ON
  CONFLICT` set, params).
- `ProjectionBuilder._load_latest_score` now selects `criteria_json`,
  `trace_json`, `correction_json` from the latest `job_scores` version and
  `_rebuild_job` threads them into both projections.
- `ProjectionBuilder._refresh_impl` runs a one-time, marker-guarded backfill of
  already-projected scored jobs whose audit columns are still NULL.

## Why (auditability data-loss bug, high severity)

Two projection builders write the SAME projection tables: the TS
`refreshProjections` and the Python `ProjectionBuilder`. The TS builder writes
the three score-audit columns; the Python builder omitted them.

Because Python's in-process event handler consumes each worker score event
**first** and advances the shared `event_watermarks` row, the TS refresher
early-returns (watermark already advanced) and never rebuilds worker-scored
jobs. The read model serves `scoreCriteria` / `scoreTrace` / `scoreCorrection`
straight from the projection column with no canonical fallback
(`read-model.ts` ~:1953-1955), so the score-audit surface (rubric, prompt/model/
version trace, self-correction history) was **NULL for every normally-scored
job** even though the data exists in `job_scores`.

## Two parts to the fix

1. **Write path (fresh + future scores).** `_load_latest_score` reads
   `criteria_json` / `trace_json` / `correction_json` from the latest
   `job_scores` version and both upserts persist them verbatim into the two
   projections.

2. **Backfill path (existing backlog).** Simply adding the columns to
   `SCORE_EVIDENCE_COLUMNS` is NOT enough on real user DBs: those columns were
   already added months ago by the TS builder's `ensureProjectionColumn`, so
   `_ensure_column` returns False, the `_reset_score_evidence_projections`
   schema-migration reset never fires, and the pre-existing Python-first rows
   (NULL audit + already-advanced watermark) would stay NULL until a brand-new
   `JobScored` event arrived. A one-time, marker-guarded backfill in
   `_refresh_impl` fixes the backlog: it marks every already-projected scored
   job whose `score_criteria_json` is still NULL as dirty and rebuilds it from
   canonical `job_scores`, then records a per-tenant marker row in the new
   `projection_backfills` table so the O(jobs) scan runs **at most once per DB**
   and every later refresh costs a single marker lookup. The marker write and
   the rebuilds share the refresh transaction, honouring the existing
   `defer_commit` contract; it is intentionally NOT coupled to PR #206's
   `SCHEMA_VERSION`.

## Parity guarantees

- Same source and ordering as TS `loadLatestScore`: latest row by `version DESC`
  from `job_scores`.
- Same correction resolution as TS: the latest score row's own `correction_json`
  column verbatim (NULL when absent). No separate corrected-vs-latest query.
- Byte-for-byte values: both runtimes pass the raw column string through without
  re-serialising. The canonical writer stores `criteria_json`/`trace_json` as
  non-empty `json.dumps(..., sort_keys=True)` (`NOT NULL DEFAULT '{}'`) and
  `correction_json` as NULL or `json.dumps(..., sort_keys=True)`, so the
  empty-string coercion in `_row_nullable_str` never triggers for these columns.

## How validated

Regression tests in `tests/test_projection_builder.py` (4 new):
- `test_projects_score_audit_columns_from_job_scores` — write path, no
  correction: audit columns projected verbatim into both projections;
  `score_correction_json` NULL.
- `test_projects_score_correction_json_when_correction_exists` — write path with
  a self-correction projected verbatim.
- `test_score_audit_backfill_repopulates_existing_null_rows` — reproduces the
  existing-DB backlog state (NULL-audit projection row + canonical `job_scores`
  row + watermark advanced past the score event) and asserts `refresh()`
  repopulates all three columns in both projections with no new event.
  **Verified to FAIL without the backfill** (temporarily disabled the backfill:
  `assert None == '{"formula_version": "score-v3"}'`) and pass with it.
- `test_score_audit_backfill_runs_at_most_once` — a NULL-audit row appearing
  after the marker is set is not re-backfilled (steady state stays cheap).

Commands:

```
  workers/automation/tests/test_projection_builder.py \
  workers/automation/tests/test_audit_projection_parity.py \
  workers/automation/tests/test_job_list_projection.py
# 25 passed

  workers/automation/src/jobhunter/infrastructure/projections/sqlite_projection_store.py \
  workers/automation/src/jobhunter/infrastructure/projections/projection_builder.py \
  workers/automation/src/jobhunter/domain/operations/projections.py \
  workers/automation/tests/test_projection_builder.py
# All checks passed!

# Safety net (full worker suite)
uv --project workers/automation run --extra dev pytest -q
# 1490 passed
```

### DIFF
--- a/workers/automation/src/jobhunter/infrastructure/projections/sqlite_projection_store.py
+++ b/workers/automation/src/jobhunter/infrastructure/projections/sqlite_projection_store.py
@@ -311,6 +311,14 @@ def ensure_projection_tables(conn: sqlite3.Connection) -> list[str]:
         ON operational_attempt_metrics(tenant_id, source_id, occurred_at DESC, metric_id DESC)
         """
     )
+    conn.execute(
+        """
+        CREATE TABLE IF NOT EXISTS projection_backfills (
+            name         TEXT PRIMARY KEY,
+            completed_at TEXT NOT NULL
+        )
+        """
+    )
     score_evidence_schema_changed = False
     for table_name, column_name, definition in SCORE_EVIDENCE_COLUMNS:
         score_evidence_schema_changed = (
--- a/workers/automation/src/jobhunter/infrastructure/projections/sqlite_projection_store.py
+++ b/workers/automation/src/jobhunter/infrastructure/projections/sqlite_projection_store.py
@@ -55,6 +55,9 @@ SCORE_EVIDENCE_COLUMNS: tuple[tuple[str, str, str], ...] = (
     ("job_list_projections", "score_keywords_json", "TEXT NOT NULL DEFAULT '[]'"),
     ("job_list_projections", "score_version", "INTEGER"),
     ("job_list_projections", "scored_at", "TEXT"),
+    ("job_list_projections", "score_criteria_json", "TEXT"),
+    ("job_list_projections", "score_trace_json", "TEXT"),
+    ("job_list_projections", "score_correction_json", "TEXT"),
     ("job_list_projections", "current_substage", "TEXT NOT NULL DEFAULT 'discover'"),
     ("job_detail_projections", "score_breakdown_json", "TEXT"),
     ("job_detail_projections", "compensation_summary_json", "TEXT"),
@@ -62,6 +65,9 @@ SCORE_EVIDENCE_COLUMNS: tuple[tuple[str, str, str], ...] = (
     ("job_detail_projections", "score_keywords_json", "TEXT NOT NULL DEFAULT '[]'"),
     ("job_detail_projections", "score_version", "INTEGER"),
     ("job_detail_projections", "scored_at", "TEXT"),
+    ("job_detail_projections", "score_criteria_json", "TEXT"),
+    ("job_detail_projections", "score_trace_json", "TEXT"),
+    ("job_detail_projections", "score_correction_json", "TEXT"),
     ("job_detail_projections", "employer_analysis_json", "TEXT"),
     ("job_detail_projections", "requirement_fit_report_json", "TEXT"),
     ("artifact_list_projections", "metadata_json", "TEXT"),
@@ -97,6 +103,9 @@ def ensure_projection_tables(conn: sqlite3.Connection) -> list[str]:
             score_reasoning        TEXT NOT NULL DEFAULT '',
             score_version          INTEGER,
             scored_at              TEXT,
+            score_criteria_json    TEXT,
+            score_trace_json       TEXT,
+            score_correction_json  TEXT,
             current_stage          TEXT NOT NULL DEFAULT 'discover',
             current_substage       TEXT NOT NULL DEFAULT 'discover',
             current_state          TEXT NOT NULL DEFAULT 'pending',
@@ -151,6 +160,9 @@ def ensure_projection_tables(conn: sqlite3.Connection) -> list[str]:
             score_reasoning        TEXT NOT NULL DEFAULT '',
             score_version          INTEGER,
             scored_at              TEXT,
+            score_criteria_json    TEXT,
+            score_trace_json       TEXT,
+            score_correction_json  TEXT,
             stages_json            TEXT NOT NULL DEFAULT '[]',
             employer_analysis_json TEXT,
             requirement_fit_report_json TEXT,
@@ -352,6 +364,7 @@ class SqliteProjectionStore:
                 full_description, fit_score, compensation_summary_json,
                 score_breakdown_json, score_keywords_json,
                 score_reasoning, score_version, scored_at,
+                score_criteria_json, score_trace_json, score_correction_json,
                 current_stage, current_substage, current_state, current_error_code,
                 current_error_message, current_next_action, has_resume,
                 has_cover_letter, has_pdf, apply_status, applied_at,
@@ -359,7 +372,7 @@ class SqliteProjectionStore:
                 last_updated_at
             ) VALUES (
                 ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?,
-                ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?
+                ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?
             )
             ON CONFLICT(tenant_id, job_id) DO UPDATE SET
                 title                 = excluded.title,
@@ -379,6 +392,9 @@ class SqliteProjectionStore:
                 score_reasoning       = excluded.score_reasoning,
                 score_version         = excluded.score_version,
                 scored_at             = excluded.scored_at,
+                score_criteria_json   = excluded.score_criteria_json,
+                score_trace_json      = excluded.score_trace_json,
+                score_correction_json = excluded.score_correction_json,
                 current_stage         = excluded.current_stage,
                 current_substage      = excluded.current_substage,
                 current_state         = excluded.current_state,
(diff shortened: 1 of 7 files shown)


workers/automation/src/jobhunter/infrastructure/projections/sqlite_projection_store.py
```diff
@@ -55,13 +55,19 @@
     ("job_list_projections", "score_keywords_json", "TEXT NOT NULL DEFAULT '[]'"),
     ("job_list_projections", "score_version", "INTEGER"),
     ("job_list_projections", "scored_at", "TEXT"),
+    ("job_list_projections", "score_criteria_json", "TEXT"),
```
ebarti, line 58: PR Reviewer - pr-reviewer

Severity: High
Gate: Fail
Evidence: The TS projection columns score_criteria_json/score_trace_json/score_correction_json have been on main since 733a29d9 (2026-05-14), and TS `ensureProjectionColumn` runs on every read-model query, so existing user DBs already have these three columns. Consequently `_ensure_column` returns False here, `score_evidence_schema_changed` stays False, and `_reset_score_evidence_projections()` (line ~323) never runs. The buggy rows (Python-inserted projection rows with NULL score-audit columns + an already-advanced event_watermarks row) are never re-marked dirty, so `_refresh_impl` early-returns without rebuilding them.
Reproduction: Reproduced with the actual PR code (worktree @ cbd7c7c). Seeded a job + a job_scores row (criteria/trace), inserted list+detail projection rows with NULL audit columns, set the watermark past the JobScored event, then ran ProjectionBuilder(conn_factory=lambda: conn).refresh() -> 0 dirty jobs processed; list.score_criteria_json and detail.score_criteria_json stayed None. A follow-up new JobScored event DID repopulate them, confirming only re-touched jobs recover.

The PR description claims "_ensure_column migrates existing DBs and triggers the standard _reset_score_evidence_projections rebuild on next refresh." That is only true when these columns are ABSENT (fresh DB). On a real upgrade the columns already exist, so no reset/backfill fires and the auditability bug this PR targets ("NULL for every normally-scored job") persists for the entire existing backlog until each job is independently re-scored / re-enriched / tailored / applied. The two new tests use a fresh init_db with empty projection tables, so they only cover the first-run backfill write path, not this existing-NULL-row + advanced-watermark state.

Required fix: Add a one-time, idempotent backfill that re-marks jobs whose projection audit columns are NULL but which have a canonical job_scores row (dirty those job_urls, or gate on a projection rebuild token / version bump), independent of _ensure_column returning True. Add a regression test that reproduces the bad state from canonical data (existing projection row with NULL audit + advanced watermark) and asserts refresh repopulates it. If backfilling existing data is intentionally out of scope, correct the PR description and document the manual recovery step in docs/local-reliability-qa.md.

ebarti: Fixed in 2935344d. You're right — on existing DBs the columns were already added by the TS `ensureProjectionColumn`, so `_ensure_column` returns False and the `_reset_score_evidence_projections` schema-migration reset never fires; the pre-existing Python-first rows (NULL audit + advanced watermark) stayed NULL. Correct root cause.

The fix now has two parts:
- Write path: `_load_latest_score` + both upserts persist criteria/trace/correction (original commit).
- Backfill path (new): a one-time, marker-guarded pass in `_refresh_impl`, INDEPENDENT of `_ensure_column`. It marks every already-projected scored job whose `score_criteria_json` is still NULL as dirty and rebuilds it from canonical `job_scores`, then records a per-tenant marker in a new `projection_backfills` table so the O(jobs) scan runs at most once per DB and later refreshes cost a single marker lookup. Marker write + rebuilds share the refresh transaction (honours `defer_commit`). Not coupled to PR #206's `SCHEMA_VERSION`.

Regression: `test_score_audit_backfill_repopulates_existing_null_rows` seeds exactly the existing-NULL state (NULL-audit projection row + canonical job_scores + watermark advanced past the JobScored event), runs `refresh()`, and asserts all three audit columns repopulate in both projections with no new event. I verified it FAILS with the backfill temporarily disabled (`assert None == '{"formula_version": "score-v3"}'`) and passes with it. Added `test_score_audit_backfill_runs_at_most_once` for the marker guard.

Targeted suite now 25 passed (4 new); full worker suite 1490 passed; ruff clean on all changed files. The PR description's "triggers the standard reset rebuild on next refresh" claim has been corrected.

ebarti: PR Reviewer - pr-reviewer -- RESOLVED by 2935344d.

Verified the marker-guarded one-time backfill in `_refresh_impl`: it dirties every already-projected scored job whose `score_criteria_json IS NULL` (JOIN job_scores) and rebuilds both list+detail from canonical, independent of `_ensure_column`. Reproduced the exact existing-NULL + advanced-watermark state against the fixed code -- all three audit columns (criteria/trace/correction) repopulate on the next refresh with NO new score event, in BOTH projections (this same scenario stayed NULL on cbd7c7c). At-most-once confirmed: the per-tenant `projection_backfills` marker prevents re-scan (a post-marker NULL row is left untouched). The mark shares the refresh transaction (honors defer_commit) and runs after the rebuilds, so a failed refresh re-runs cleanly. New regression test genuinely reproduces the bad state (manual NULL-audit projection rows + watermark advanced past the event, not a fresh DB) plus a separate at-most-once guard test. pytest 14 passed (incl. parity) + ruff clean. High resolved.

ebarti:
Appended B3 (watermark hardening, defense-in-depth) — commit `febfa66f` `fix(projections): monotonic watermark honoring defer_commit`.

**What changed (Python only)**
- `SqliteEventWatermarkRepository.set()` now takes `commit: bool = True`. The `ProjectionBuilder` passes `commit=not defer_commit` at both watermark-advance sites, so when a refresh runs inside the caller's open transaction (e.g. `record_job_event` during `acquire_job`'s `BEGIN IMMEDIATE`), the watermark write no longer flushes the caller's uncommitted writes / releases the row lock. The builder's transaction owns the commit and the watermark bump stays atomic with the projection rebuild.
- `set()` is now monotonic via `MAX(...)`: a stale/out-of-order write with a lower `last_event_id` is a no-op and can never rewind the watermark. `updated_at` moves only when the value advances.
- Per-writer watermark separation was **deliberately skipped**: the TS runtime reads the same `event_watermarks` table, so a per-writer column/row would ripple cross-runtime. The shared watermark stays correct for both writers under MAX semantics (noted in the commit body).
- Port docstring (`domain/ports/watermark.py`) updated to document the monotonic contract + the `commit` flag.

**Test evidence (reproduce-then-fix)**
- 3 new regression tests in `test_event_watermark.py`: `test_set_is_monotonic_never_regresses`, `test_deferred_set_does_not_autocommit` (asserts the deferred write stays pending and a rollback reverts it), `test_default_set_commits_immediately`. Confirmed the first two FAIL on the prior tip and pass after the fix.
- `uv --project workers/automation run --extra dev pytest -q` → **1493 passed**.
- `uv --project workers/automation run --extra dev ruff check .` → clean.

No TS changes in this PR.

PR #208, opened 2026-07-01
fix(scoring): period-aware posted-pay parsing to stop false hard blockers

ebarti:
## What

The deterministic scoring `ConstraintChecker` fabricated hard eligibility blockers from brittle text heuristics, silently dropping truthful, high-fit jobs from the tailoring/apply funnel. A hard blocker zeroes the downstream eligibility gate (`_SCORE_ELIGIBLE_FOR_DOWNSTREAM` / `js_hard_blocker_count` in `database.py`).

Fixes in `workers/automation/src/jobhunter/domain/scoring/services.py`:

- **Compensation is now period-aware.** It reuses the existing posted-compensation parser (`jobhunter.domain.compensation.parse_posted_compensation`) instead of blindly multiplying any salary amount `< 1000` by `1000`. The old `_salary_amounts`/`_parse_amount`/`_salary_max` helpers (and the `_SALARY_AMOUNT` regex) are removed.
- **Below-minimum is a hard blocker only when confidently annual.** A below-minimum reading is a hard blocker only when the structured `jobs.salary` field yields a genuinely annual figure (period `year`, or a salary-field range whose amounts are unambiguously annual-scale). Hourly/monthly figures (annualized via a fixed-hours/x12 assumption) and description-derived figures become **warnings**, so the job stays in the funnel for the user to judge.
- **Remote/work-model mismatch is a warning, not a hard blocker.** A remote-preference vs onsite-posting mismatch is a soft preference, not a hard eligibility constraint.
- **Sponsorship stays a hard blocker (unchanged message).** It is corroborated by the structured `work_authorization.require_sponsorship` flag plus an explicit denial phrase — a genuine visa constraint the task asked to keep working.
- **Auditability.** Every compensation blocker/warning now records its source field (`jobs.salary`/`jobs.description`), the parsed annual ceiling, the period, and the annualization assumption, so `breakdown.eligibility` explains why a job was gated.

## Why

`$50/hourly` from jobspy was mis-annualized to `$50,000` (via `amount < 1000 -> *1000`), tripping a false "below profile minimum" hard blocker on a role actually paying ~`$104k/yr` (`$50 x 2080`). Remote/sponsorship substring checks were similarly fragile. The result was a truthful, high-fit job silently removed from the funnel with the reason buried. Per the repo's Root-Cause & Auditability discipline, a hard blocker must reflect genuine ineligibility from an explicit, auditable source.

## How validated

- `uv --project workers/automation run --extra dev pytest -q workers/automation/tests/test_scorer.py workers/automation/tests/test_score_use_cases.py workers/automation/tests/test_posted_compensation_parser.py` -> **59 passed**
- Full worker suite: `uv --project workers/automation run --extra dev pytest -q workers/automation/tests` -> **1491 passed**
- `uv --project workers/automation run --extra dev ruff check workers/automation/src/jobhunter/domain/scoring/services.py` -> **All checks passed**

### Regression tests added/updated (`workers/automation/tests/test_score_use_cases.py`)

- `test_constraint_checker_hourly_pay_annualizes_and_does_not_false_block` — `$50/hourly` with a `$80k` desired min annualizes to `$104k` and produces **no** blocker/warning.
- `test_constraint_checker_hourly_below_minimum_is_warning_with_annualization_audit` — `$50/hr` with a `$150k` min is a **warning** (not a hard blocker); message cites `$104,000`, `jobs.salary`, and the `2,080`-hour assumption.
- `test_constraint_checker_confident_annual_below_minimum_is_hard_blocker_with_source` — `$40,000 per year` below a `$120k` min stays a **hard blocker** with source + parsed figure in the audit trail.
- `test_constraint_checker_monthly_pay_below_minimum_is_warning_not_hard_blocker` — monthly pay below floor is a warning.
- `test_constraint_checker_remote_work_model_mismatch_is_warning_not_hard_blocker` — remote-vs-onsite is a warning.
- Updated `test_score_job_keeps_hard_blockers_separate_from_high_score` (remote is now a warning; sponsorship stays a blocker) and `test_score_job_blocks_when_explicit_posted_compensation_is_below_minimum` (asserts the enriched audit message).

## Docs

Added a regression-matrix row to `docs/local-reliability-qa.md` capturing the eligibility-classification invariant (no false hard blockers from mis-annualized pay or preference mismatches; source/reason preserved in the audit trail).

## Safety

No auto-apply, browser, or destructive actions were run. Scoring rubric weights and fit-band thresholds are unchanged.


workers/automation/src/jobhunter/domain/scoring/services.py
```diff
@@ -641,17 +645,90 @@ def _first_number(value: Any) -> int | None:
...
+        if ceiling is None:
+            return None
+        return _PostedCompensationSignal(
+            annual_ceiling=ceiling,
+            source_field=fact.source_field,
+            period=fact.period,
+            assumption=fact.annualization_assumption or "",
+            reliable_annual=salary_field and fact.period == "year",
+        )
+    ceiling = fact.maximum_amount
```
ebarti, line 716: PR Reviewer - pr-reviewer

Severity: Low
Gate: Pass
Evidence: `parse_posted_compensation`/`_detect_period` recognize only hour/month/year — never week or day. In this unknown-period branch an undetected-period figure from the structured salary field is stamped `reliable_annual=salary_field` (True) and read as annual. Verified against this PR head: `ConstraintChecker().evaluate(job={"salary": "$15,000/week"}, min $120k)` → status "blocked" (a ~$780k/yr role false-hard-blocked); `"$2000/week"` and `"$800/day"` fall below `_ANNUAL_SALARY_FLOOR` (12,000) and fail open → "eligible"; a bare `"$13,000"` from the salary field is likewise read as annual → hard blocker.
Reproduction: Reproduced with a Python probe importing ConstraintChecker/_posted_compensation_signal from a worktree at the PR head.

This is a narrow residual of the exact class the new regression-matrix row claims to close ("no false hard blockers from mis-annualized posted pay"): a weekly/daily rate at or above the $12,000 floor is mis-annualized into a false below-minimum HARD blocker, because the "/week" token is discarded and the amount is then treated as confidently annual. Realistic weekly/daily rates sit below the floor and fail open safely, and this is strictly better than main, so impact is Low and it does not gate merge.

Required fix (optional, non-gating): prefer extending `_detect_period`/`_annualize` in `domain/compensation/posted.py` to recognize week/day (×52 / ×260) so those readings are classified and annualized explicitly. Note: naively flipping period=="unknown" to `reliable_annual=False` here would downgrade the intended k-suffixed/large-bare annual hard-block (e.g. "$80k-$95k", which parses as period unknown) to a warning and break `test_score_job_blocks_when_explicit_posted_compensation_is_below_minimum`, so the guard must distinguish a residual unrecognized per-period marker in the source text from a genuinely bare annual figure. Alternatively, narrow the regression-matrix claim to state weekly/daily are out of scope.

ebarti: Addressed in d4d192b7. The eligibility gate now detects week/day pay markers that the shared posted-comp parser leaves as `period=unknown` and annualizes them (week x52, day x260) as **warnings** — never hard blockers, since they rely on a fixed-schedule assumption. `$15,000/week` now annualizes to ~$780k -> eligible; a below-min `$300/day` (~$78k) is a warning, not a hard blocker.

Per your note, the guard distinguishes a residual per-period marker in the source text from a genuinely bare annual figure: a bare `$13,000` (and `$80k-$95k`, which the parser reports as `period=unknown`) carries no week/day marker, so it is still read as annual and `test_score_job_blocks_when_explicit_posted_compensation_is_below_minimum` still passes.

I kept the fix in the scoring layer rather than extending `_detect_period`/`_annualize` in `domain/compensation/posted.py`: that parser's period + warning enums are mirrored in `packages/contracts/src/schemas.ts` (`POSTED_COMPENSATION_PERIODS`, warning codes), consumed by the projection builder, and persisted via SQLite — extending them would diverge the cross-language contract/persistence, beyond this bug's scope and untested by the given verify set.

New tests: `test_constraint_checker_weekly_pay_annualizes_and_does_not_false_block`, `test_constraint_checker_daily_rate_below_minimum_is_warning_not_hard_blocker`, `test_constraint_checker_bare_amount_without_period_is_still_read_as_annual`. 62 passed (three-file set), 1494 (full worker suite), ruff clean.

PR #209, opened 2026-07-01
fix(discovery): don't hard-reject engineering titles containing business-function tokens

ebarti:
## What

The discovery title filter's `_has_business_function_false_positive` check hard-rejected **any** title containing a standalone business-function token (`account`, `accounts`, `commercial`, `pmo`, `pricing`, `sales`) — unless that token was in the user's query — *before* verbatim/alias/window matching and the LLM role adjudicator ever ran (`workers/automation/src/jobhunter/discovery/title_filter.py`, called at the top of `title_matches_query`).

Result: legitimate engineering roles were dropped as **pure false negatives** the LLM precision filter could never rescue: `Staff Engineer, Pricing Platform`, `Engineer, Commercial Systems`, `Engineering Manager, Accounts`, `Software Engineer, Sales Platform`. For a job hunter, dropping a relevant role is the worst failure mode.

## How it works (recall-first, precision via the LLM only)

Two layers, mirroring the existing position-aware `_has_excluded_engineering_specialty` precedent:

1. **Deterministic hard-reject (precision, unchanged shape).** `_has_engineering_role_head` recognizes a genuine engineering head (`architect`, `developer`, `engineer`, `engineering`, `sre`, `swe`). A title is still hard-rejected only when a business token is present with **no** genuine engineering head — i.e. the business token *is* the role head or its immediate modifier: `Sales Engineer`, `Account Executive`, `Sales Manager`, `Commercial Director`, `Pricing Analyst`. This matches the locked `PMO Engineering Manager` recall invariant.

2. **LLM adjudication for the ambiguous case (recall-safe).** When a title carries a business token beside a **non-adjacent** engineering head, deterministic rules cannot tell an engineering-primary role (`Staff Engineer, Pricing Platform`) from a business-primary one (`Sales Director, Engineering`). `_business_function_needs_adjudication` flags this case and suppresses the verbatim short-circuit in both the strict and recall paths, so the title routes through `_adjudicate_loose_match` to the role adjudicator:
   - **Adjudicator available** -> it decides (rejects `Sales Director, Engineering`, accepts `Staff Engineer, Pricing Platform`).
   - **Adjudicator unavailable/disabled** (no API key) -> the loose-match path **accepts**, preserving recall. There is never a new deterministic hard-reject for the ambiguous case.

The multi-word business-function phrase check (`business transformation`, `customer success`, `product management`) is intentionally untouched, so locked invariants like `Director of Customer Success for Engineering Platforms` (rejected) still hold. Clean verbatim matches with no business token still bypass the LLM (no new adjudicator calls for ordinary titles).

## Fix history

- `f63edd9a` — position-aware rescue so business-function tokens no longer hard-reject engineering titles.
- `403a00fc` — follow-up for reviewer's Medium (#209 (comment)): the adjacency-only head check let business-primary titles with a *non-adjacent* business token (`Sales Director, Engineering`, `Pre-Sales Solutions Engineer`, `Sales Development Engineer`) pass the hard-reject and then verbatim-short-circuit to ACCEPT with the adjudicator never consulted. Now routed to the adjudicator (accept when it is absent).

## How validated

```
    workers/automation/tests/test_title_filter.py \
    workers/automation/tests/test_role_title_matcher.py \
    workers/automation/tests/test_discovery_title_filter.py
19 passed

    workers/automation/src/jobhunter/discovery/title_filter.py
All checks passed!

    workers/automation/tests/test_discovery_production_wiring.py \
    workers/automation/tests/test_workday_discovery.py \
    workers/automation/tests/test_smartextract_discovery.py \
    workers/automation/tests/test_target_search_preferences.py \
    workers/automation/tests/test_discovery_limits.py
85 passed
```

### Regression tests (`test_title_filter.py`)

- `test_business_and_engineering_titles_default_to_accept_without_adjudicator` — must-survive titles **and** the reviewer's five leak shapes are all accepted when the adjudicator is absent (`role_matcher=None`): recall preserved.
- `test_business_and_engineering_titles_are_routed_to_the_adjudicator` — the same titles consult the adjudicator (`calls == 1`) instead of verbatim-short-circuiting.
- `test_business_primary_engineering_titles_are_rejected_when_adjudicator_declines` — the five leak shapes are rejected when the adjudicator declines (`calls == 1`), proving precision runs.
- `test_business_function_primary_roles_are_deterministically_rejected` — `Sales Engineer`, `Account Executive`, `Sales Manager`, `Commercial Director`, `Pricing Analyst` are hard-rejected before adjudication (accepting matcher never consulted).

## Docs

`docs/local-reliability-qa.md` regression-matrix row states the recall-first-with-LLM-precision behavior: ambiguous engineering titles route to the adjudicator and default to accept when it is unavailable; only business-primary titles are deterministically rejected.


workers/automation/src/jobhunter/discovery/title_filter.py
```diff
@@ -382,14 +390,31 @@ def _has_business_function_false_positive(
...
+        title_sequence, excluded_tokens
+    ):
         return True
     return any(
         phrase.issubset(title_tokens) and not phrase.intersection(query_token_set)
         for phrase in _EXCLUDED_BUSINESS_FALSE_POSITIVE_PHRASES
     )
 
 
+def _has_engineering_role_head(
```
ebarti, line 406: PR Reviewer - pr-reviewer

Severity: Medium
Gate: Pass
Evidence: `_has_engineering_role_head` only disqualifies the engineering head when the *immediately preceding* token is a business token (index-1 check). A business token elsewhere in the title still counts the head as "genuine," so several business-primary roles now fall through and, because they verbatim-match a short query, are accepted with the LLM adjudicator NEVER consulted. Reproduced (role_matcher that returns False, i.e. LLM says "no"; calls=0 in every case, so verbatim short-circuits `_adjudicate_loose_match` before adjudication):
- "Sales Director, Engineering" vs "Engineering" -> ACCEPTED (was hard-rejected before)
- "Head of Sales, Platform Engineering" vs "Engineering" -> ACCEPTED
- "Pre-Sales Solutions Engineer" vs "Engineer" -> ACCEPTED
- "Solutions Engineer, Sales" vs "Engineer" -> ACCEPTED
- "Sales Development Engineer" vs "Engineer" -> ACCEPTED
Reproduction: Reproduced with a probe script calling title_matches_query(title, query, role_matcher=_Fake(False)) in the PR worktree; each returned True with matcher.calls==0.

Two concrete problems: (1) The PR states these titles "fall through to ... the LLM precision filter," but for short/broad queries they verbatim-match and skip the adjudicator entirely, so the stated safety net does not engage. (2) The new `test_business_function_primary_roles_remain_rejected` only pins no-head titles (Account Executive, Sales Manager, ...) and the adjacent-head case (Sales Engineer); none of the leak shapes above are covered, so the suite implies broader rejection coverage than actually exists.

This is bounded (only single-token "Engineer"/"Engineering" queries leak; realistic multi-word queries like "Software Engineer"/"Engineering Manager" correctly reject these), it is the deliberately-accepted side of a recall-over-precision trade, and scoring is a downstream backstop -- hence Medium, not a blocker. But it is a real precision regression the team-lead asked about.

Required fix (pick one): (a) accept the trade explicitly -- document it and add regression tests pinning which business-primary+eng-head titles are intentionally allowed through on short queries; or (b) tighten -- disqualify the engineering head when ANY business token precedes it in the sequence (catches "Sales Director, Engineering" / "Pre-Sales Solutions Engineer" while keeping "<Eng Role>, <Business Domain>" survivors where the business token trails the head), and for the residual trailing-business cases ("Solutions Engineer, Sales") force adjudication instead of the verbatim short-circuit so the LLM precision filter actually runs as claimed.

ebarti: Fixed in 403a00fc.

Confirmed the gap: `_has_engineering_role_head` only checked the token *immediately* before the engineering head, so a business-primary title with a **non-adjacent** business token passed the hard-reject and then verbatim-short-circuited to ACCEPT with the adjudicator never consulted.

New behavior — recall-first, precision only via the LLM:
- A new `_business_function_needs_adjudication` check flags the ambiguous case (business token + non-adjacent genuine engineering head) and suppresses the verbatim short-circuit in both the strict and recall paths, so the title routes through `_adjudicate_loose_match` -> the role adjudicator.
- Adjudicator unavailable/disabled (no API key) -> the loose-match path returns accept, preserving recall. Never a new deterministic hard-reject.
- Adjudicator available -> it can reject business-primary titles.
- Adjacent / head-less business-primary titles ("Sales Engineer", "Account Executive", "Sales Manager", "Commercial Director", "Pricing Analyst") stay deterministically rejected as before.

All five leak shapes you reproduced ("Sales Director, Engineering", "Head of Sales, Platform Engineering", "Pre-Sales Solutions Engineer", "Solutions Engineer, Sales", "Sales Development Engineer") are now covered:
- consult the LLM and are rejected when it declines -> `test_business_primary_engineering_titles_are_rejected_when_adjudicator_declines` (asserts calls == 1);
- default to ACCEPT when the LLM is absent -> `test_business_and_engineering_titles_default_to_accept_without_adjudicator` (`role_matcher=None`).

Also added `test_business_and_engineering_titles_are_routed_to_the_adjudicator` (must-survive + leak shapes all consult the adjudicator, calls == 1) and reframed the overstated test as `test_business_function_primary_roles_are_deterministically_rejected`. The QA-doc row now states the recall-first-with-LLM-precision behavior honestly.

Verification (all green):
- `pytest -q test_title_filter.py test_role_title_matcher.py test_discovery_title_filter.py` -> 19 passed
- `ruff check title_filter.py` -> All checks passed
- `pytest -q test_discovery_production_wiring.py test_workday_discovery.py test_smartextract_discovery.py test_target_search_preferences.py test_discovery_limits.py` -> 85 passed

PR #210, opened 2026-07-01
fix(api): render resume PDF via real HTML renderer instead of truncated hand-rolled writer

ebarti:
## What changed

Both TypeScript resume `resume_pdf` producers stopped hand-rolling a single-page text PDF and now render the resume HTML they already build through the real Playwright HTML→PDF path.

- **New shared seam** `apps/api/src/resume-pdf-render.ts`: `ResumeHtmlPdfRenderer` type + `defaultResumeHtmlPdfRenderer`, a synchronous `execFileSync("uv … python -c …")` call that renders an HTML file to a PDF file via the worker's Playwright adapter. Synchronous so it runs inside the existing better-sqlite3 write transactions; injectable so tests don't spawn a browser. It throws (with captured stderr) on failure.
- **Python** `html_resume_pdf.py`: added a public `render_resume_html_to_pdf(html_content, output_path)` wrapper over the existing `_render_resume_pdf_playwright` primitive (registered in `__all__`).
- **`resume-review-drafts.ts`**: `persistRenderedDraftArtifacts` now calls `renderPdf({ htmlPath, pdfPath })` on the full `htmlForEditedResume(...)` HTML (previously thrown away) instead of the truncating writer. Threaded through `renderResumeReviewDraft` / `createOrLoadResumeReviewDraft`.
- **`resume-templates.ts`**: `persistRenderOnlyRefresh` renders `htmlForTemplateRefresh(...)` the same way; threaded through `ensureCurrentResumeTemplateMaterials` / `resolveCurrentResumeArtifactIdForOpen`. Added `data-resume-page="1"` to the refresh HTML so the shared layout extractor has a page ancestor (the edited-resume HTML already had one).
- **`server.ts`**: `BuildAppOptions.resumePdfRenderer` (default `defaultResumeHtmlPdfRenderer`), wired into the render, ensure-current, create-draft, and artifact-open routes.
- Deleted the hand-rolled `pdfBufferForEditedResume` / `pdfBufferForText` and their `escapePdfText` helpers. `render_format` stays `html_pdf`. Existing heuristic layout boxes are unchanged (out of scope; not flagged).
- Docs: `docs/local-ts-api.md` (render + ensure-current now use the real renderer, multi-page, fail-safe) and `docs/local-reliability-qa.md` (new no-truncation regression-matrix row).

## Why

`pdfBufferForEditedResume` / `pdfBufferForText` kept only the first 58 non-empty lines, cut each line to 110 chars, and drew from `y=750` at 14pt on one 792pt page with no wrapping. Lines past ~54 fell off-page and line 59+ were dropped. This buffer was persisted as the highest-generation `resume_pdf` with status `approved` and served as the current, submit-ready resume (`preview.pdf` / apply join). The document the user reviewed, opened, and submitted was truncated and layout-broken. The correct HTML→PDF path already existed (worker `HtmlResumePdfAdapter`, `local-actions.ts` profile preview) and the full-page resume HTML was already built and then discarded.

## Failure behavior

Render runs before any DB insert. In the review flow a render failure propagates out of the write transaction (rollback → 500) and no new generation is written; in the template-refresh flow it is caught, recorded as a `failed` attempt, and the last approved generation is preserved. No truncated/broken PDF is ever persisted.

## Validation

- `pnpm api:check` — clean (tsc, no errors).
- `pnpm api:test` — **17 files, 289 tests passed** (includes 4 new tests: no-truncation + render-failure-preservation for both the review-draft and template-refresh flows, asserting the full untruncated HTML/content reaches the renderer and the output is not the hand-rolled `72 750 Td` buffer).
- `uv --project workers/automation run --extra dev ruff check .` — all checks passed.
- `uv --project workers/automation run --extra dev pytest -q` — **1487 passed** (includes a new wrapper test; Temporal gRPC "connection refused" warnings are pre-existing, no dev server running, tests pass).

Unit tests mock the Playwright subprocess via the injectable renderer; no browser is spawned in the suite.


apps/api/src/resume-pdf-render.ts
```diff
@@ -0,0 +1,61 @@
...
+render_resume_html_to_pdf(html_path.read_text(encoding="utf-8"), str(output_path))
+`;
+
+// Renders pre-built resume HTML to a full, paginated PDF via the same Playwright
+// adapter the Python worker uses, so the reviewed/submitted resume matches the
+// edited HTML instead of a truncated single-page fallback. Synchronous so it can
+// run inside the better-sqlite3 write transactions that persist the artifact.
+export const defaultResumeHtmlPdfRenderer: ResumeHtmlPdfRenderer = ({ htmlPath, pdfPath }) => {
+  try {
+    execFileSync(
```
ebarti, line 41: PR Reviewer - pr-reviewer

Severity: Low
Gate: Pass
Evidence: `execFileSync` is called with no `timeout` option. It is synchronous by design (to run inside the better-sqlite3 write transaction), so it blocks the entire Node event loop for the full render duration — `uv` project resolution + Python import + Chromium launch + render, several seconds cold. During that window the whole API is unresponsive to every request, not just the caller. This now also fires on the `/v1/artifacts/:artifactId/open` route (via ensureCurrentResumeTemplateMaterials), so merely opening an artifact that needs a refresh can freeze the server. Worse, a hung render (pathological HTML that never satisfies Playwright `wait_until="load"`, or a stuck Chromium) blocks the event loop indefinitely with no recovery but a process restart.
Reproduction: Reasoned from code (unit tests inject a mock renderer, so the real subprocess is never exercised).

The preservation invariant is unaffected — a hang writes nothing, so the prior approved generation is safe. This is availability/latency only, and acceptable for the current single-user local tool, hence Low. Flagging because it becomes serious under the documented hosted-future seam (one tenant’s render would freeze all tenants).

Required fix (optional, non-gating): pass a bounded `timeout` (and consider `maxBuffer`) to `execFileSync`. A timeout throws, which the existing `catch` already turns into a render failure → the flow aborts before any DB insert and preserves the prior generation, so it composes cleanly with the fail-safe path. Longer term, consider whether the render must hold the write transaction open for its whole duration.

ebarti: Fixed in fd87203. The synchronous `execFileSync` now passes `timeout: 120_000` (120s — generous enough for a cold Chromium start) plus a generous `maxBuffer` (64MB), so a hung `uv` resolve / Chromium launch / render can no longer freeze the Node event loop. On timeout `execFileSync` throws, which the existing catch wraps into a render-failure Error, so the fail-safe path is unchanged (review-draft render rolls back the write tx; template refresh records a `failed` attempt) and the last approved generation is preserved. render_format stays `html_pdf`. Added `apps/api/test/resume-pdf-render.test.ts` (2 tests) asserting the timeout/maxBuffer options are passed and that an ETIMEDOUT-style throw surfaces as a render failure. Verified: `pnpm api:check` clean; `pnpm api:test` 18 files / 291 passed.

PR #211, opened 2026-07-01
fix(materials): render resume PDF before superseding prior approved generation

ebarti:
## What changed

On **re-tailor**, `TailorResumeUseCase.execute` superseded the prior approved generation (status -> `superseded`) and published `ResumeApproved` **as soon as the new resume passed validation/judge** — before the PDF existed. The runner (`_tailor_one_job`) then rendered the PDF inline and, on **any** render error, demoted the just-approved new generation to `rejected`.

Net result of a transient render failure: `prior = superseded`, `new = rejected`, so `load_current_approved` returned `None` — the job **permanently lost its last accepted resume**, a `ResumeApproved` event was left dangling with no artifact, and provenance rows orphaned against a now-rejected generation. This violated the preservation invariant (`docs/architecture.md`, CLAUDE.md "Root-Cause And Auditability Discipline").

This PR moves resume PDF rendering **into the use case's transaction boundary**, reordered so the artifact is durable before any destructive state change:

- `TailorResumeUseCase` gains an optional `pdf_renderer: PdfRendererPort`. When an approved candidate is produced, the PDF is rendered **before** superseding the prior generation and **before** publishing `ResumeApproved`.
- On render **success**: attach the PDF, supersede the prior, save, record provenance, publish `ResumeApproved` (unchanged happy path) — now with `TailorOutcome.pdf_path` populated.
- On render **failure**: the prior approved generation is left **intact** (never superseded), the new generation is demoted to a `rejected` attempt (audit history), `ResumeFailed` is published, and **no provenance** is written for the unshipped generation. `load_current_approved` therefore still returns the prior generation with its artifact + provenance.
- `_tailor_one_job` no longer renders/saves the PDF; it only surfaces the use case outcome. Dead `_selected_candidate_payload` and the now-unused `ValidationResult` import were removed.

**First-time tailoring behaviour is unchanged**: with no prior generation, a PDF failure still rejects the new (and only) generation — nothing was destroyed because nothing existed. Consistent with PR #204 (last accepted materials must survive failed refreshes).

## Why

A re-tailor must never leave a job with no current approved resume. Rendering after the supersede/publish made a transient PDF failure destroy canonical, user-visible material and produced an audit trail (dangling `ResumeApproved`, orphaned provenance) that no longer matched reality.

## How it was validated

Added a regression test that reproduces the bad state from canonical data (`test_retailor_pdf_render_failure_preserves_previous_approved_generation`): gen 1 is a real accepted resume with a durable PDF + provenance; a gen-2 re-tailor whose PDF render fails leaves gen 1 as the current approved resume (`load_current_approved` returns it with PDF + provenance), gen 2 is rejected audit history, no second `ResumeApproved` fires, and no gen-2 provenance is orphaned. The in-memory materials fake was made faithful to the real repo's per-generation overwrite so the test actually catches the bug (verified: it fails when the supersede is reordered before the render, passes with the fix). Also added use-case-level tests for the moved render (happy path attaches PDF + surfaces `pdf_path`; first-generation render failure rejects and leaves nothing current) and updated the two `_tailor_one_job` runner tests to assert outcome propagation. Added a high-risk regression-matrix row in `docs/local-reliability-qa.md`.

Commands:

```
    workers/automation/tests/test_tailor_retailor.py \
    workers/automation/tests/test_materials_use_cases.py \
    workers/automation/tests/test_tailor_provenance_integration.py
63 passed in 0.60s

    workers/automation/src/jobhunter/domain/materials/use_cases.py \
    workers/automation/src/jobhunter/scoring/tailor.py
All checks passed!

$ uv --project workers/automation run --extra dev pytest -q   # full worker suite
1489 passed in 39.18s
```

## Docs

No behavioral doc change is warranted — this is a bug fix that restores the already-documented preservation invariant rather than changing public behavior. Added one regression-matrix entry in `docs/local-reliability-qa.md`, the doc that owns high-risk workflows.


workers/automation/src/jobhunter/domain/materials/use_cases.py
```diff
@@ -1565,6 +1574,45 @@ def execute(
...
+                    report=report,
+                    error=render,
+                    final_payload=final_payload,
+                )
+            pdf_artifact, pdf_path = render
+            materials = materials.with_resume_pdf(pdf_artifact, updated_at=_utc_now())
+
         if materials.is_resume_approved and prior_generation is not None:
             self._repository.save(prior_generation)
         self._repository.save(materials)
```
ebarti: PR Reviewer - pr-reviewer

Severity: Low
Gate: Pass
Evidence: The PDF render is now correctly ordered BEFORE this supersede — on render failure `save(prior_generation)` is skipped, gen N stays `approved`, and `load_current_approved` returns it (verified). However `save(prior_generation)` (line 1617) and `save(materials)` (line 1618) are two separate transactions: `SqliteMaterialsRepository.save` ends each call with `self._conn.commit()`. A process crash between these two commits (after a successful render) would leave gen N superseded and gen N+1 unpersisted, so `load_current_approved` transiently returns None.
Reproduction: Reasoned from code (infrastructure/materials/sqlite_repository.py `save` commits per call). The render-ordering fix itself was verified by reproducing the original bug (reordered supersede before render -> regression test fails with `load_current_approved == None`).

The supersede of the prior generation and the persist of the new approved generation are not atomic. This is pre-existing (the PR did not introduce this ordering) and the PR materially narrows the window by moving the failure-prone PDF render ahead of the supersede, so this does not block merge.

Required fix: None required for this PR (out of scope). Future hardening: wrap `save(prior_generation)` + `save(materials)` in a single transaction / unit-of-work so the supersede and the new approved generation commit together.

ebarti: Tracked as follow-up **A9 (materials persist unit-of-work)** in the session backlog: wrap `save(prior_generation)` + `save(materials)` in a single transaction so the supersede and the new approved generation commit together. Out of scope here per this review (pre-existing ordering; this PR narrows the window by rendering before superseding). Resolving.

ebarti:
### Appended fix: A9 — atomic unit-of-work for the generation flip

Follow-up landed on this PR (commit `e6166ac2`). #211 fixed the *ordering* (render the replacement resume PDF before superseding the prior approved generation). A9 makes the *persistence itself* atomic.

**The gap.** The tailor persist path does a three-write "generation flip": supersede the prior approved generation → save the new generation → record the new generation's provenance/coverage. Each SQLite repository committed per call, so this was three separate transactions:
- A crash **between the supersede and the new-generation save** committed the supersede alone, leaving `load_current_approved` transiently (or, on a hard crash, permanently) empty — the job lost its last accepted resume.
- A **provenance write failure after the artifact committed** left an approved generation with no provenance, breaking the audit invariant that every committed approved generation has its provenance (the old `_record_provenance` deliberately swallowed that failure).

**The fix.**
- New `UnitOfWork` port (`domain/ports/materials.py`) + `SqliteUnitOfWork` adapter (`infrastructure/materials/unit_of_work.py`) over the shared connection. The materials + provenance SQLite repos defer their per-call commit while a unit of work is active; the boundary commits once on clean exit and rolls the whole block back on any exception.
- `TailorResumeUseCase` opens the unit of work around the flip, so supersede + new-generation save + provenance/coverage save commit atomically. The provenance write moved **inside** the flip and no longer swallows failures — it propagates to trigger rollback.
- `BulletProvenanceRecorded` / `ResumeApproved` events and best-effort requirement-fit coverage enrichment now run **after** the flip commits, so they never announce or enrich a rolled-back state.
- Wired in `scoring/tailor.py::_build_use_case` for the shared-connection default repos; injected-repo callers keep per-call commits (no false atomicity guarantee).

**Crash-window guarantees now:** any failure mid-flip rolls the whole block back — the prior approved generation stays current and no new artifact is ever committed without its provenance. #211's render-before-supersede ordering is preserved: text + PDF are written to disk before the flip, so on rollback the DB (source of truth) holds no row referencing the new bytes (orphaned bytes on disk are acceptable). Behaviour on success is identical.

**Reproduce-then-fix tests.**
- `test_materials_unit_of_work.py` — drives the **real** SQLite repos + `SqliteUnitOfWork`: rollback keeps generation 1 current on both a crash between supersede/save and a provenance write failure. Control tests **without** the unit of work reproduce the pre-A9 bad state (current resume lost / approved generation orphaned without provenance), proving the unit of work is load-bearing.
- `test_tailor_provenance_integration.py` — use-case orchestration: the flip runs inside the injected unit of work and commits once on success; a provenance failure rolls the boundary back and propagates (no dangling `ResumeApproved` / `BulletProvenanceRecorded`).
- Added a companion QA regression-matrix row in `docs/local-reliability-qa.md`.

**Verification.** `uv --project workers/automation run --extra dev pytest -q` → **1496 passed**. `uv --project workers/automation run --extra dev ruff check .` → **clean**. #211's existing render-failure preservation tests remain green.

workers/automation/src/jobhunter/domain/materials/use_cases.py
```diff
@@ -1565,23 +1585,79 @@ def execute(
...
+        # new artifact is ever committed without its provenance (the audit
+        # invariant: every approved generation has its provenance). A failed
+        # re-tailor still writes no provenance rows, so the prior generation's
+        # rows survive untouched (Anti-Pattern 4 / success criterion 5). The
+        # Phase-3 generation-time coverage (GROUND-06) + voice audit (VOICE-02)
+        # ride on the same set. Event publication + best-effort requirement-fit
+        # coverage enrichment run only AFTER the flip commits so they never
+        # announce or enrich a rolled-back state.
+        record_provenance = materials.is_resume_approved and bool(provenance_rows)
+        with self._unit_of_work if self._unit_of_work is not None else nullcontext():
```
ebarti, line 1647: PR Reviewer - pr-reviewer-a8

Severity: Medium
Gate: Pass
Evidence: `SqliteUnitOfWork` (infrastructure/materials/unit_of_work.py:44-56) manages no transaction of its own — it relies on sqlite3's default deferred autocommit and just calls `conn.commit()` / `conn.rollback()` on the outermost `__exit__`. The flip's all-or-nothing guarantee therefore depends on an *unenforced* invariant: no other writer on the shared thread-local connection may issue `commit()` / `rollback()` / `BEGIN` between the first staged write (use_cases.py:1642 supersede) and the boundary commit (unit_of_work.py:53). Today that holds — I traced it: `SqliteTailoringPolicyRepository.resolve_current` (sqlite_repository.py:889, `BEGIN IMMEDIATE` + `commit()` on the *same* connection) runs at use_cases.py:1514, *before* the block, and the in-process EventPublisher's synchronous projection commits run *after* (use_cases.py:1657/1665, both swallow). Full worker suite: 1496 passed. So this is NOT a current defect.
Reproduction: Reasoned from code + full-suite run (no competing writer currently touches the connection inside 1640-1651).

The risk is latent/regression, and it is exactly concern (6) of this re-review. Nothing prevents a future edit from silently re-opening the crash window this commit closed: a stray `conn.commit()` from any non-enrolled repo sharing the connection (policy repo, a projection handler, a watermark write) placed inside the flip would commit the *partial* flip (prior superseded, new generation possibly missing), and a stray `BEGIN`/`BEGIN IMMEDIATE` would raise mid-flip. Because the guarantee is implicit, such a change would land with the entire suite green: `test_materials_unit_of_work.py` drives only the two enrolled repos, and the `_RecordingUnitOfWork` integration double does not simulate a real transaction, so neither exercises a competing writer on the real connection during the flip.

Required fix (pick one, in order of strength): (a) have `SqliteUnitOfWork.__enter__` open an explicit transaction/`SAVEPOINT` so a non-enrolled `BEGIN` surfaces loudly ("cannot start a transaction within a transaction") instead of silently splitting the flip; or (b) add a regression test that drives the REAL `SqliteUnitOfWork` + real repos and asserts a competing `conn.commit()` mid-flip is prevented/detected; at minimum (c) add a prominent invariant comment at use_cases.py:1640 naming the "no other commit on this connection inside the flip" constraint so future edits don't reintroduce the bug.

ebarti: Fixed in dc57dc0c.

`SqliteUnitOfWork.__enter__` now opens an explicit `BEGIN IMMEDIATE` at the outermost depth (the re-entrant depth counter and `__exit__` commit/rollback are unchanged), mirroring the existing `SqliteTailoringPolicyRepository.resolve_current` idiom. The flip is now one explicit transaction on the shared connection rather than resting on the driver's implicit deferred `BEGIN` plus the unenforced assumption that nothing else commits/rolls back/BEGINs mid-block:

- **(a)** Explicit boundary: a stray `BEGIN` from another writer on the shared connection now fails loudly instead of silently splitting the flip, and the write lock is taken upfront for the whole block (shrinking the `SQLITE_BUSY` race when workers > 1).
- **(b)** New regression test `test_flip_holds_one_explicit_transaction_and_locks_out_competitors` drives the REAL `SqliteUnitOfWork` + real repos on the real connection: asserts `conn.in_transaction is True` across every staged write, and that a competing writer's `BEGIN IMMEDIATE` on a second connection is refused (busy_timeout-bounded) while the flip is open. Existing rollback tests stay green.
- **(c)** Prominent invariant comment at the flip site (`use_cases.py`) naming the rule: the block is one transaction on the shared connection; no code reachable from it may commit, roll back, or open its own transaction, and event/projection writes stay OUTSIDE the block.

No `docs/local-reliability-qa.md` row added — the existing A9 atomic-flip regression row already covers this hardening.

Verified: `uv --project workers/automation run --extra dev pytest -q` -> 1497 passed; `ruff check .` -> clean.

PR #212, opened 2026-07-01
fix(discovery): dedup postings across all sources by content identity, not just jobspy

ebarti:
## What

Adds content-based deduplication to the shared discovery write boundary (`DiscoverJobsUseCase`) so a posting re-discovered on a different source collapses onto the existing `Job` aggregate instead of creating a second one.

- **`SqliteJobRepository.find_content_owner`** (new): resolves an existing `Job` by the SAME content identity JobSpy already uses — normalized **title + company + description fingerprint**, or a substantial-description shingle match — both gated on exact title+company equality. It reuses `domain/job_content_identity.py` (no duplicated normalization) and coalesces `jobs.company` (populated by JobSpy) with `jobs.site` (the board, which is the employer for ATS-owned rows), so a posting collapses regardless of which source discovered it first.
- **`DiscoverJobsUseCase`**: after `find_canonical_owner` misses (native-id / exact-URL / normalized-observation checks), it now calls `find_content_owner`. A content match is an authoritative merge — it bypasses the low-confidence identity quarantine and records a `DuplicateJobLink` with reason `content_fingerprint_match` at confidence `0.95`, matching the JobSpy path.
- Added `find_content_owner` to the `JobRepository` port.

## Why

Only the JobSpy adapter ran content-fingerprint/description dedup against existing jobs. Workday, ATS-API, and manual-capture ingest go solely through `DiscoverJobsUseCase.find_canonical_owner`, which matches only on `(source_id, source_native_id)` or an exact/normalized canonical URL. Because JobSpy runs first, the same posting later found on the company's Workday/ATS board resolves to a **different canonical URL** and became a **second `Job` aggregate** — so one real job was scored twice, tailored twice (double LLM spend), and showed a double apply surface.

**Avoiding over-merge:** strictness matches JobSpy exactly, so genuinely different roles at the same company (different title, or same title with a divergent description) stay separate. The scoring-only repost/title-variant helpers (`role_titles_match_as_repost`) are intentionally NOT enabled here.

**Scope note:** `smartextract._store_jobs_filtered` persists rows via direct SQL and does **not** go through `DiscoverJobsUseCase`, so it is not covered by this change. The task framing listed smart-extract as a use-case intake; in the current code it is a separate legacy path. Routing it through the shared intake is a larger follow-up and is intentionally out of scope for this fix.

## How validated

New regression tests in `test_discovery_identity.py`:
1. JobSpy job re-discovered via a canonical source (Workday) with the same title+company+description collapses to ONE `Job` (second ingest adds an observation + `content_fingerprint_match` duplicate link, no new job).
2. Reworded cross-source description (ATS-first, Greenhouse → Lever) collapses via the shingle path.
3. Two distinct roles at the same company (different title) stay separate — no over-merge.
4. Same title+company with divergent descriptions stay separate — no over-merge.
5. Native-id identity match still wins over content (records `canonical_url_match`, not `content_fingerprint_match`).

Commands (run from repo root):

```
  workers/automation/tests/test_discovery_identity.py \
  workers/automation/tests/test_workday_discovery.py \
  workers/automation/tests/test_ats_adapters.py \
  workers/automation/tests/test_smartextract_discovery.py
# 53 passed

  workers/automation/src/jobhunter/domain/discovery/use_cases.py \
  workers/automation/src/jobhunter/infrastructure/discovery/sqlite_repository.py
# All checks passed!

# Broader regression coverage:
  workers/automation/tests/test_discovery_limits.py \
  workers/automation/tests/test_discovery_production_wiring.py \
  workers/automation/tests/test_discovery_scheduler_pr4.py
# 92 passed (with test_discovery_preparation_orchestration + test_content_validator)

# Full worker suite:
uv --project workers/automation run --extra dev pytest -q
# 1491 passed
```

## Docs

No doc update warranted: this is an internal discovery bug fix with no change to CLI commands, API surface, generated artifacts, or runtime requirements. `docs/job-pipeline-architecture.md` already describes Discover as owning "dedupe against existing jobs" (now more accurate across sources).

---

## Update — review round 2 (fix HIGH: cross-employer over-merge)

Reviewer reproduced a HIGH: the round-1 match key used `posting.source.board` (incoming) and `COALESCE(NULLIF(company,''), site)` (stored). Use-case-created Jobs save `employer="Unknown"` so `jobs.company` is empty → the stored gate collapsed to `site` = board. For CONSTANT platform/sentinel boards (`Source(board="User-mediated capture")` for manual capture; the `"Workday"` fallback; JobSpy platform boards) the employer discriminator vanished and two DISTINCT employers' jobs could merge on title + description alone (silent data loss).

**Fix (commit `1a0774ad`):** content dedup now requires a genuine employer on BOTH sides.
- New `jobhunter.domain.job_content_identity.is_genuine_employer_identity` rejects empty, `"Unknown"`, and non-employer platform/board labels (JobSpy boards `linkedin/indeed/zip_recruiter/glassdoor/google/bayt/naukri/bdjobs`, `"User-mediated capture"`, `"Workday"`).
- `DiscoverJobsUseCase` attempts content dedup only when the incoming `source.board` is a genuine employer.
- `SqliteJobRepository.find_content_owner` rejects a non-genuine incoming company and skips any candidate whose resolved employer (`company` else `site`) is not genuine — a sentinel board can never be the merge key. Either side lacking a genuine employer falls through to a distinct Job (safe under-merge). JobSpy-company→Workday/ATS-real-employer and ATS per-employer boards still dedup.
- Medium (b): the SQL prefilter now compares title/employer via a registered `normalize_identity_text` SQL function instead of `lower(trim())`, so unicode/punctuation-only duplicates are no longer silently excluded.

**New tests (round 2):** cross-employer over-merge regressions for the manual-capture and Workday-fallback sentinel boards (verified FAIL-before / PASS-after by reverting only the two behavior files), a boilerplate-heavy same-employer/same-title separation test (0.83 threshold), and a `is_genuine_employer_identity` denylist unit test. `test_discovery_identity.py` now has 23 tests.

**Re-verified:** required pytest set → 57 passed; ruff (use_cases.py, sqlite_repository.py, job_content_identity.py, ports/discovery.py) → All checks passed!; broader discovery suite → 110 passed; full worker suite → 1495 passed.


workers/automation/src/jobhunter/domain/discovery/use_cases.py
```diff
@@ -256,6 +267,17 @@ def _ingest_one(
             source_native_id=identity.source_native_id,
             canonical_url=identity.canonical_url,
         )
+        content_matched = False
+        if owner_id is None:
+            content_owner_id = self._repository.find_content_owner(
+                tenant_id,
+                title=posting.metadata.title,
+                company=posting.source.board,
```
ebarti, line 293: PR Reviewer - pr-reviewer

Severity: High
Gate: Fail
Evidence: The incoming content key is `posting.source.board`, not the employer. Combined with `find_content_owner`'s `COALESCE(NULLIF(j.company,''), j.site)` gate (sqlite_repository.py:476) and the fact that every use-case-created Job is saved with `employer="Unknown"` -> `jobs.company` empty (use_cases.py:351 `_create_new_job` + save() writing `None if job.employer.is_unknown()`), the stored gate key collapses to `jobs.site` = `Source.board`. When `board` is a shared CONSTANT rather than the employer, the employer discriminator disappears and two DISTINCT jobs at different employers can merge on title + description alone. Constant boards exist today: manual capture -> `Source(board="User-mediated capture")` (production_wiring.py:1866, ALWAYS constant) and Workday -> `board = employer_name or "Workday"` (workday.py:372, constant when employer_name missing). (ATS boards are per-employer, so ATS is safe.)
Reproduction: Reproduced in the PR worktree against init_db + SqliteJobRepository. Two distinct rows (jobA, jobB) with company='' , site='User-mediated capture', title='Software Engineer', and boilerplate-heavy descriptions (~0.9 shingle Jaccard) -> `find_content_owner(title='Software Engineer', company='User-mediated capture', description=<jobB>)` returned jobA => jobB collapses into jobA. Control with distinct boards ('Acme' vs 'Globex') correctly kept them separate, proving the board string is the sole cross-employer guard.

Why this is High: the merge is silent and lossy. Because `content_matched=True` bypasses the low-confidence quarantine branch (use_cases.py:432), the wrongly-absorbed posting attaches as an observation on the wrong Job with a 0.95 `DuplicateJobLinked` and NO rejection event -- a real captured job disappears with no signal. This is NEW behavior: JobSpy's dedup keys on the real scraped company (never on `site`/board), so it cannot collide across employers this way; the `company=board` + `COALESCE(company, site)` design introduces the collision.

Required fix: Use an employer-derived key for the incoming side that matches the stored semantics (e.g., the posting's employer when known), and do NOT let a shared/constant/platform board act as the company key. Concretely: skip the content lookup (or the site-coalesce) when the board is a non-employer sentinel (manual-capture constant, Workday fallback), or gate the coalesce so `jobs.site` is only used as an employer proxy for sources where board == employer. Add a regression test seeding two distinct-employer rows that share a constant board and asserting they stay separate.

workers/automation/src/jobhunter/infrastructure/discovery/sqlite_repository.py
```diff
@@ -428,6 +433,63 @@ def find_canonical_owner(
...
+                   COALESCE(je.full_description, j.full_description, j.description)
+                       AS description,
+                   CASE WHEN d.job_url IS NULL THEN 0 ELSE 1 END AS is_deleted
+            FROM jobs j
+            LEFT JOIN job_enrichments je ON je.job_url = j.url
+            LEFT JOIN jobhunter_deleted_jobs d
+              ON d.job_url = j.url
+             AND (d.restored_at IS NULL OR julianday(d.restored_at) <= julianday(d.deleted_at))
+            WHERE lower(trim(COALESCE(j.title, ''))) = ?
+              AND lower(trim(COALESCE(NULLIF(j.company, ''), j.site, ''))) = ?
```
ebarti: PR Reviewer - pr-reviewer

Severity: Medium
Gate: Pass
Evidence: Two coverage/strictness gaps in the shingle path, both untested: (1) Same employer + same title + heavy shared boilerplate. The 0.83 shingle threshold is inherited from JobSpy, but the new test `test_..._divergent_descriptions_separate` only exercises deliberately divergent text (~0.04 Jaccard). Real same-company postings for two DIFFERENT roles under an identical title routinely share a large boilerplate block (about-us / benefits / EEO / legal); if that boilerplate dominates the token count, two distinct roles can exceed 0.83 and merge. This risk is inherited from JobSpy but this PR newly extends it to Workday/ATS. (2) The SQL prefilter compares `lower(trim(...))` against a fully `normalize_identity_text`-normalized bound param (NFKC + smart-quote/dash folding + internal-whitespace collapse + casefold), so a stored title/company with those variations is excluded from candidates -> a real cross-board duplicate is missed. This is fail-safe (under-merge) and matches JobSpy exactly, so it is not a regression, but it undercuts the fix's own goal for punctuation/whitespace-divergent cross-board titles.
Reproduction: Reasoned from code; (1) follows from the token-fraction math on `_description_shingles` (5-grams, >=80 tokens, Jaccard>=0.83); (2) from the `lower(trim())` vs `normalize_identity_text` asymmetry.

Required fix: Add a regression test for same-title/same-company boilerplate-heavy-but-distinct roles asserting they stay separate (and tighten if it fails -- e.g. weight role-specific vs shared shingles, or require a minimum count of NON-shared shingles). Optionally normalize the SQL prefilter consistently (or drop it to a pure title+company equality on already-normalized stored columns) so whitespace/punctuation variants are not silently excluded.

workers/automation/src/jobhunter/domain/discovery/use_cases.py
ebarti: Fixed in 1a0774ad. The incoming side no longer passes the raw board as the employer key unconditionally: `_ingest_one` now calls `find_content_owner` only when `is_genuine_employer_identity(posting.source.board)` holds. A platform/sentinel board ("User-mediated capture", the "Workday" fallback, JobSpy boards, or "Unknown") is not a genuine employer, so the posting falls through to a distinct Job (safe under-merge). Regression tests `test_discover_jobs_use_case_does_not_merge_distinct_employers_behind_manual_capture_board` and `...behind_workday_fallback_board` reproduce the cross-employer over-merge and now pass (verified FAIL-before / PASS-after).

workers/automation/src/jobhunter/infrastructure/discovery/sqlite_repository.py
ebarti: Fixed in 1a0774ad. The stored-side coalesce no longer blindly trusts `site`: `find_content_owner` rejects a non-genuine incoming employer up front, and for each candidate it resolves the stored employer (`company` else `site`) and skips it unless `is_genuine_employer_identity` holds, so a sentinel board (empty company falling back to "User-mediated capture" / "Workday" / a JobSpy board) can never be the merge key. Per-employer ATS/Workday-real boards still dedup. Also addressed the prefilter normalization mismatch (Medium b): title/employer are now compared via a registered `normalize_identity_text` SQL function instead of `lower(trim())`, so unicode/punctuation-only duplicate variants are no longer silently excluded.

workers/automation/src/jobhunter/domain/discovery/use_cases.py
ebarti: PR Reviewer - pr-reviewer

Severity: Resolved
Gate: Pass
Verification: Re-reviewed at 1a0774ad. The `is_genuine_employer_identity` gate on BOTH sides fixes this. Re-ran my original reproduction against the fixed code: incoming board "User-mediated capture" now returns None (two distinct jobs stay separate) instead of merging. Confirmed the denylist is complete vs the actual JobSpy Site enum {bayt,bdjobs,glassdoor,google,indeed,linkedin,naukri,zip_recruiter} plus the "Workday" fallback and "User-mediated capture". Targeted dedup preserved: JobSpy(company=Acme)+Workday(board=Acme) still merges; distinct genuine employers (Acme vs Globex) with similar descriptions stay separate. The two new regression tests genuinely reproduce the bad state (manual-capture and Workday-fallback, ~0.93 Jaccard, distinct employers) and assert 2 jobs — fail-before/pass-after confirmed by construction. The gate check only narrows what can merge, so no new over-merge is introduced. Resolved.

workers/automation/src/jobhunter/infrastructure/discovery/sqlite_repository.py
ebarti: PR Reviewer - pr-reviewer

Severity: Medium (partially addressed)
Gate: Pass
Verification: (b) normalization asymmetry is FIXED — the SQL prefilter now calls the registered `jh_normalize_identity` (== Python `normalize_identity_text`); verified a stored double-space title now matches a normalized incoming title. (a) same-title boilerplate: a guard test (`test_..._keeps_boilerplate_heavy_distinct_roles_separate`, 75 boilerplate + 25 distinct tokens -> ~0.59 Jaccard) now pins that the common case stays separate. Residual (non-blocking): two distinct roles at the SAME employer with the SAME exact title and >=~90% shared boilerplate (<=~8 distinct tokens) can still exceed the 0.83 threshold and merge. This is inherited JobSpy strictness, now partially tested. Acceptable as a documented follow-up; not a gate blocker.

ebarti:
## Fix: HIGH cross-PR composition defect — rejected content duplicate deleted the accepted owner

Pushed `e0c2dd60` to `fix/discovery-cross-source-content-dedup`.

### Defect (reproduced deterministically)
This PR's cross-source content dedup resolves an incoming posting to an existing owner via `find_canonical_owner` → `find_content_owner`, then routes a content-owner match into `_observe_existing_job`. That function's **pre-existing** policy-rejection branch soft-deletes the resolved job. Because the content fingerprint is title + company + description (**location is not in the fingerprint**), a DISTINCT posting of the same role from another source content-matches the accepted owner — and if current policy rejects it (e.g. `location_mismatch`), the branch **soft-deleted the accepted owner**, emptying `list_recent`. Final state was order-dependent per run (an accepted re-scrape restores it; the rejected one re-deletes it).

Scenario: an employer posts the same role as "Remote – US" (accepted, already in the funnel) and "Bengaluru, India" (arrives from another source). The India copy content-matches and is rejected for location → the Remote–US owner got deleted.

Confirmed with the new regression test on the pre-fix code: owner comes back `is_deleted=True`, `delete_reason='current_policy_mismatch: location_mismatch'`.

### Fix (minimal, root-cause)
- Thread the "resolved via content match" fact explicitly from the resolution site into `_observe_existing_job` (via a new `ContentOwnerMatch` return from `find_content_owner`), instead of re-deriving it.
- When the owner was resolved by **content identity**, a policy rejection **no longer soft-deletes the owner** and **no longer attaches the rejected posting as an owner observation** — attaching it would let `find_canonical_owner` resurface it as a same-identity re-observation and delete the owner on a *later* run. The declined duplicate is recorded as `DuplicateJobLinkRejected` audit; the owner is left untouched.
- **Genuine same-URL re-observations that fail policy still soft-delete exactly as before** (existing test `test_discover_jobs_use_case_soft_deletes_active_job_rejected_by_current_policy` unchanged and green).

### Also fixed (LOW, audit truthfulness)
`find_content_owner` now returns the match **basis**, so a shingle-similarity match records `reason=content_shingle_match` (confidence `0.85`) instead of overstating it as `content_fingerprint_match` (reserved for exact fingerprint matches, `0.95`).

### Tests
- New: `test_discover_jobs_use_case_keeps_accepted_owner_when_content_duplicate_rejected` — reproduce-then-fix; asserts owner stays active + in `list_recent`, no `JobDeleted`, no India observation attached, `DuplicateJobLinkRejected` audit emitted, and stability on a second rejected run.
- Updated: `test_discover_jobs_use_case_collapses_reworded_cross_source_description` now asserts `content_shingle_match` @ `0.85`.
- Extended the `docs/local-reliability-qa.md` delete/hide regression-matrix row with this invariant.

### Verification
- Targeted: `pytest workers/automation/tests/ -k "dedup or discovery or content or repository or identity"` → **315 passed**.
- Full: `pytest -q` → **1496 passed**; `ruff check .` → **All checks passed!**

ebarti:
## Appended fixes: closed the four documented cross-source dedup gaps (E2b + E2c)

Five commits on top of `e0c2dd60`, one per gap plus a docs row. No thresholds changed (0.95 fingerprint / 0.85 shingle audit weights, 0.83 Jaccard preserved); the anti-over-merge and accepted-owner-preservation guard-rail tests stay green.

### E2b-a — Smart Extract bypass (`31a0320b`)
`smartextract._store_jobs_filtered` wrote rows with direct SQL, never content-deduping. It now calls the shared `SqliteJobRepository.find_content_owner` before every insert and, on a genuine content match to a different URL, records the duplicate link + a source observation against the surviving owner, publishes `DuplicateJobLinked`, resurfaces the owner, and skips the insert.
Tests (`test_smartextract_discovery.py`): ATS-first then Smart Extract same content -> single aggregate + `content_fingerprint_match` link; and the reverse direction.

### E2b-b — JobSpy-second direction (`b7c9dbda`)
JobSpy's pre-use-case dedup filtered on `jobs.company` alone, so an ATS-first owner (use-case rows leave `company` NULL, employer in `site`) was missed. The lookup now coalesces `company` -> `site` with the genuine-employer guard, mirroring `find_content_owner`.
Test (`test_discovery_limits.py::test_jobspy_dedups_against_ats_first_null_company_owner`) — verified FAILS on the pre-fix tip (stored a 2nd aggregate, `(1,0)`), passes after.

### E2b-c — post-enrichment miss (`37a7984c`)
`find_content_owner` compared the incoming LISTING against `COALESCE(je.full_description, ...)`; once the owner is enriched, listing-vs-enriched similarity drops below threshold. A shared `content_match_basis` helper now compares against BOTH the stored listing (`jobs.description`) and the enriched full text (strongest basis wins); both `find_content_owner` and the JobSpy lookup use it.
Test (`test_discovery_identity.py::test_discover_jobs_use_case_matches_fresh_listing_against_enriched_owner`) — reproduced the miss on a real fixture pair (reworded-vs-enriched Jaccard **0.657 < 0.83**, reworded-vs-listing **0.945**); verified FAILS on the pre-fix tip (`new_jobs=1`), passes after.

### E2c — rejected-link audit attribution + idempotency (`08fcec73`)
`DuplicateJobLinkRejectedPayload.candidate_ids` (no owner key, no consumer) is replaced with `job_id` (surviving owner — the key `_event_job_url` and the audit read model already key on) + `candidate_job_id` (declined posting). The event now lands in the owner's audit history and the TS "Candidate" row populates; the dead confidence row is dropped from the read-model mapping. Rejected links are persisted in a new `job_rejected_duplicate_links` table keyed on (tenant, owner, candidate URL) and the event is published only on first record, so repeat observations append no rows.
Tests: Python event shape + idempotency (`test_discovery_identity.py`, both rejection paths); TS read-model mapping with a fixture event (`apps/api/test/server.test.ts`).

### Verification (all green on the final tip)
- `uv --project workers/automation run --extra dev pytest -q` -> **1500 passed**
- `uv --project workers/automation run --extra dev ruff check .` -> **All checks passed**
- `pnpm api:check` -> clean; `pnpm api:test` -> **286 passed**

### Residual (out of scope, noted for follow-up)
`DuplicateJobLinked` (the accepted-merge event) is still attributed with `job_url` NULL because `_event_job_url` does not recognise `surviving_job_id`; only the rejected event was in scope here. Attributing the linked event to its owner would be a small, separate follow-up.

ebarti:
### Follow-up: accepted-merge `DuplicateJobLinked` now attributes to its owner (`617a623f`)

Closes the residual flagged in the GAP4 comment. The rejected event was fixed there; the accepted-merge event had the same invisibility bug and is now fixed symmetrically.

**Root cause.** `DuplicateJobLinkedPayload` carries the surviving owner in `surviving_job_id` (both call sites: `use_cases.py` and `smartextract.py`). `DurableJobEventPublisher._event_job_url` only recognised `job_id`/`jobId`/`job_url`/`jobUrl`, so the durable publisher stored every merge with a NULL `job_url` — invisible in the owner's audit history. `DuplicateJobLinked` flows *only* through this publisher path (no direct `record_job_event` with an explicit URL), so this was the single source of truth.

**Fix (two sides).**
- **Publisher** (`production_wiring.py`): `_event_job_url` now recognises `surviving_job_id` (appended last, so a direct owner key still wins). New merges persist with `job_url` = owner. Only `DuplicateJobLinkedPayload` carries this key, so no other event type is affected.
- **Read model** (`read-model.ts`): `buildJobAuditHistory` matched only `job_url` OR `JSON_EXTRACT` `jobId`/`job_id`/`jobKey`/`job_key`, so merges already persisted with a NULL `job_url` stayed invisible even after the publisher fix. Added `$.surviving_job_id` to the audit query so both new and pre-existing merges surface. Only `DuplicateJobLinked` carries that key and it is always the owner, so there is no mis-attribution risk. The display (`Reason` + `Confidence`) already aligned with the persisted snake_case keys, so no display change was needed (unlike the rejected event which needed `candidate_job_id` alignment).

**Reproduce-then-fix evidence.**
- Python: strengthened `test_smart_extract_dedups_against_ats_first_content_owner` to assert the `DuplicateJobLinked` row's `job_url == owner_url`. Pre-fix: `AssertionError: assert None == 'https://boards.greenhouse.io/acme/jobs/staff-eng'`. Post-fix: passes.
- TS: new `server.test.ts` test "attributes accepted duplicate links to the surviving owner's audit history" inserts a `DuplicateJobLinked` event with `job_url` NULL and `surviving_job_id` set (the exact legacy state). Stashing only the `read-model.ts` clause reproduced the miss (`auditHistory` was `[]`, entry `undefined`); with the clause it renders "Duplicate linked" / `Content Fingerprint Match` / `95%`.

**Verification (worktree tip `617a623f`).**
- `uv --project workers/automation run --extra dev pytest -q` -> 1500 passed
- `uv --project workers/automation run --extra dev ruff check .` -> All checks passed
- `pnpm api:check` -> clean
- `pnpm api:test` -> 287 passed (17 files)

QA regression-matrix row extended narrowly to name the accepted-merge attribution invariant. Guard rails intact (thresholds 0.95/0.85 unchanged; audit bases preserved).

workers/automation/src/jobhunter/domain/events/discovery.py
```diff
@@ -272,13 +272,16 @@ def create_duplicate_job_linked(
...
+    matched) so the rejected link shows in that job's audit history — ``job_id``
+    is the standard attribution key both the durable event publisher and the API
+    audit read model key on. ``candidate_job_id`` is the distinct posting that
+    was declined.
     """
 
     duplicate_link_id: str
-    candidate_ids: tuple[str, ...]
+    job_id: str
+    candidate_job_id: str
```
ebarti, line 284: PR Reviewer - pr-reviewer

Severity: Medium
Gate: Pass
Evidence: This commit renamed `DuplicateJobLinkRejectedPayload` from `candidate_ids` to `job_id` + `candidate_job_id` (discovery.py:283-284), but the canonical shared TS contract was not updated: `packages/domain-types/src/events/discovery.ts:321-326` still declares `readonly candidateIds: readonly string[]` with no `jobId`/`candidateJobId`. `workers/automation/tests/test_domain_event_parity.py` only compares event-type NAMES, not payload fields, so nothing catches the divergence; the domain-types tests (`packages/domain-types/test/events.test.ts:186,802`) and the web fixture (`apps/web/src/test/fixtures/events.ts:146`) still build the event with `candidateIds` and pass.
Reproduction: Reasoned from code; confirmed `pnpm api:test` (287 passed) and discovery pytest (91 passed) are both green, i.e. the drift is silent.

The runtime read path is correct (read-model.ts reads snake_case `candidate_job_id` and matches `$.job_id`), so there is no current user-visible break. But the canonical event contract now misdescribes the payload Python actually emits: a TS consumer trusting the type reads `event.payload.candidateIds` (always `undefined`) and cannot type-safely reach the owner id (`jobId`) or the candidate. This also blocks the correct fix for the frontend `duplicateJobLinkRejectedHandler`, which now needs `event.payload.jobId` to invalidate the newly-attributed owner's job detail on SSE.

Required fix: Update `DuplicateJobLinkRejectedPayload` in `packages/domain-types/src/events/discovery.ts` to `{ duplicateLinkId; jobId; candidateJobId; reason; rejectedAt }`, and update the domain-types tests + `apps/web/src/test/fixtures/events.ts` to match. Consider extending `test_domain_event_parity.py` to also compare payload field names so future renames cannot drift silently.

workers/automation/src/jobhunter/discovery/smartextract.py
```diff
@@ -122,6 +203,24 @@ def _store_jobs_filtered(
         if not description:
             missing_description += 1
             continue
+        content_owner = repository.find_content_owner(
+            LOCAL_TENANT,
+            title=str(job.get("title") or ""),
+            company=str(job.get("company") or ""),
+            description=description,
+        )
+        if content_owner is not None and str(content_owner.job_id) != url:
```
ebarti, line 212: PR Reviewer - pr-reviewer

Severity: Medium
Gate: Pass
Evidence: Smart Extract now SKIPS the INSERT on a content match (smartextract.py:212-223). The anti-over-merge guards (`is_genuine_employer_identity` on both sides + exact normalized title/employer prefilter + `content_match_basis` 0.83 shingle threshold) are shared with `DiscoverJobsUseCase`, which HAS negative tests (`test_discover_jobs_use_case_keeps_distinct_roles_at_same_company_separate` and the same-title/divergent-description pair). But the two new Smart Extract tests (`test_smart_extract_dedups_against_ats_first_content_owner`, `test_ats_dedups_against_smart_extract_first_content_owner`) are BOTH positive (should-merge), and the new JobSpy company->site path (`test_jobspy_dedups_against_ats_first_null_company_owner`) is likewise positive-only.
Reproduction: Reasoned from code + test inventory (grep of the discovery test files); targeted discovery pytest = 91 passed.

The route that now suppresses inserts has no regression test proving two genuinely-distinct roles at the same employer with >=0.83-similar text stay SEPARATE via this route. Because the guard is shared and tested on the use-case path, current production risk is bounded — but a future change to the Smart Extract wiring (wrong company/description arg passed to `find_content_owner`, or a guard/threshold change) would silently collapse two distinct jobs into ONE `Job` on this intake (silent data loss the user never sees), with nothing red on the Smart Extract or JobSpy-null-company paths.

Required fix: Add a Smart Extract negative test — two distinct-title (or same-title/divergent-description) postings at the same genuine employer must yield TWO `jobs` rows via `_store_jobs_filtered` (assert no `job_duplicate_links` row, `COUNT(*) FROM jobs == 2`). Add the mirror negative on the JobSpy company-NULL / site-coalesce path.

workers/automation/src/jobhunter/domain/events/discovery.py
ebarti: Fixed in b205b592 (contract + parity) and 03e34b22 (the handler this unblocks).

- Contract: `DuplicateJobLinkRejectedPayload` in `packages/domain-types/src/events/discovery.ts` is now `{ duplicateLinkId; jobId; candidateJobId; reason; rejectedAt }` (snake->camel of the Python payload), with a doc comment noting `jobId` = surviving owner. Updated both domain-types tests (`events.test.ts:183,800`) and the web fixture (`apps/web/src/test/fixtures/events.ts:144`).
- Parity: `test_domain_event_parity.py` now asserts payload FIELD parity for both discovery duplicate-link events, driven from the Python dataclasses via `dataclasses.fields()` (camelised) against the TS interface fields. Proven to catch this exact drift -- reverting the TS type to `candidateIds` fails with `python=[candidateJobId,duplicateLinkId,jobId,reason,rejectedAt] ts=[candidateIds,duplicateLinkId,reason,rejectedAt]`. Scoped to the duplicate-link pair (documented in the test): a generic sweep over all events would need a real TS AST parse (nested/shared sub-payloads, unions) to stay safe.
- Unblocked handler (03e34b22): `duplicateJobLinkRejectedHandler` now invalidates `jobsKeys.detail(event.payload.jobId)` so an open owner drawer live-refreshes over SSE.

Verified: domain-types check + 170 tests; pnpm web:check / api:check clean; web 846 / api 287; full pytest 1503; ruff clean.

workers/automation/src/jobhunter/discovery/smartextract.py
ebarti: Fixed in 82d5ed5a. Added both negatives:

- Smart Extract: `test_smart_extract_keeps_distinct_roles_at_same_employer_separate` -- seeds an ATS owner (`Staff Platform Engineer`), then `_store_jobs_filtered` a `Staff Data Scientist` at the same employer with a near-identical, shingle-eligible description (168 tokens, Jaccard 1.0). Asserts `_store_jobs_filtered == (1, 0)`, `COUNT(*) FROM jobs == 2`, `COUNT(*) FROM job_duplicate_links == 0`.
- JobSpy company->site null-company: `test_jobspy_keeps_distinct_roles_at_same_null_company_owner_separate` -- same two roles, owner seeded with NULL company (employer in `site`), JobSpy frame carrying the distinct title + shared description. Asserts `store_jobspy_results == (1, 0)`, 2 jobs, 0 links.

Each negative differs from its sibling positive test ONLY in the candidate title, isolating the normalized-title+employer identity gate as the reason for non-merge. Confirmed the shared description is genuinely in the merge regime (`descriptions_substantially_match(shared, shared) == True`, 168 tokens), so if the title gate stopped short-circuiting, the shared text would fingerprint- or shingle-match and the test would fail by merging.

Verified: full pytest 1503 passed; ruff clean.

ebarti:
### Re-review round: 2 Mediums + 1 unblocked Low fixed (`b205b592`, `82d5ed5a`, `03e34b22`)

Both inline Medium threads replied-to and resolved. Three commits on top of `617a623f`:

- **M1 (contract drift + parity) -- `b205b592`**: `DuplicateJobLinkRejectedPayload` in `packages/domain-types/src/events/discovery.ts` now matches the Python payload: `{ duplicateLinkId; jobId; candidateJobId; reason; rejectedAt }`. Updated the two domain-types tests and the web fixture. Extended `test_domain_event_parity.py` to assert payload FIELD parity for both discovery duplicate-link events (driven from the Python dataclasses); proven to fail on the old `candidateIds` shape. Scoped to the duplicate-link pair -- a generic sweep would need a real TS AST parse to stay safe.
- **M2 (over-merge negatives) -- `82d5ed5a`**: added distinct-roles negatives for both new skip-insert routes -- `test_smart_extract_keeps_distinct_roles_at_same_employer_separate` and `test_jobspy_keeps_distinct_roles_at_same_null_company_owner_separate`. Each differs from its sibling positive test only in the candidate title (near-identical 168-token, shingle-eligible descriptions), isolating the title+employer identity gate; both assert 2 `jobs` rows and 0 `job_duplicate_links`.
- **L (unblocked by M1) -- `03e34b22`**: `duplicateJobLinkRejectedHandler` now invalidates `jobsKeys.detail(owner jobId)` so an open owner drawer live-refreshes over SSE; router parity test extended; `every-event-has-handler` stays green.

**Verification**: `domain-types` check + 170 tests; `pnpm web:check` / `api:check` clean; web 846 / api 287; full `pytest` 1503 passed; `ruff` clean.

PR #213, opened 2026-07-02
feat(observability): trace ensemble legs, synthesizer, and voice pass as Langfuse generation spans

ebarti:
## What

Wraps the five previously-untraced frontier-model calls in the system in the existing `llm_generation_span` so they emit `langfuse.observation.type=generation` spans:

- **Employer-analysis ensemble** — the three parallel draft legs (Claude Agent SDK, Codex SDK, Antigravity/Gemini SDK) and the Claude synthesizer pass.
- **Resume voice pass** — the Claude voice adapter.

Each span carries the model name, prompt/completion, latency (span duration), and — when the SDK surfaces usage — input/output token counts, mirrored into both the Langfuse `langfuse.observation.*` and the GenAI-semconv `gen_ai.*` attributes.

## Why

Each employer-analysis run makes 4 frontier-model calls plus the voice pass makes a 5th. None of the agent-SDK adapters wrapped their calls in a span — only the legacy httpx `LLMClient.chat` was traced. So the single most expensive AI in the system reported **no** model, prompt, completion, token counts, cost, or latency to Langfuse, and there was no cost aggregation anywhere. `docs/architecture.md` admitted folding these legs into the span pipeline was still "next".

## How

- Token usage is read from each SDK's native shape and normalised: Claude `ResultMessage.usage` (fresh + cache-creation + cache-read = total input), Codex `TurnResult.usage.total`, Antigravity `response.usage_metadata` (`prompt_token_count` in, `candidates + thoughts` out). Extraction is best-effort: **absent usage → the span omits tokens rather than fabricating them.**
- Distinct instrumentation scopes (`jobhunter.analysis.{claude,codex,antigravity,synthesizer}`, `jobhunter.materials.voice`) keep the legs, synthesizer, and voice pass separable under the shared `llm.<model>` span name.
- **Control flow is unchanged.** An SDK error is recorded on the span (`StatusCode.ERROR`) and re-raised into the existing per-leg retry / partial-failure path; the ensemble's "no wall-clock timeout / partial-failure safe" behaviour is untouched.
- **Per-analysis rollup is free.** The legs run inside the enclosing pipeline-stage / JSON-RPC span (OTel context propagates through `asyncio.run` + `asyncio.gather`), so Langfuse aggregates their token usage and cost onto the surrounding analysis trace with no extra plumbing.
- No new key management, no new dependencies, no control-flow or product-behaviour change.

## Tests

New span assertions (in-memory OTLP exporter, SDK boundary mocked as the existing tests do) prove each leg + synthesizer + voice pass opens a generation span with the model attribute and token attributes when the mocked SDK returns usage:

- `test_employer_analysis_ensemble.py`: Claude draft, Codex draft, Claude synthesizer spans; graceful token omission when the SDK reports no usage; SDK error recorded on span + re-raised.
- `test_antigravity_adapter.py`: Antigravity draft span + graceful token omission.
- `test_voice_adapter.py`: voice-pass span + graceful token omission.

## Docs

`docs/architecture.md` observability section: added the ensemble legs / synthesizer / voice rows to the spans table and replaced the "next observability step" caveat with the implemented behaviour (generation spans, distinct scopes, trace-level cost rollup, non-fatal instrumentation).

## Validation

\`\`\`
    workers/automation/tests/test_llm_spans.py \
    workers/automation/tests/test_employer_analysis_ensemble.py \
    workers/automation/tests/test_voice_adapter.py \
    workers/automation/tests/test_otel_init.py \
    workers/automation/tests/test_antigravity_adapter.py
43 passed in 0.47s

    workers/automation/src/jobhunter/infrastructure/analysis \
    workers/automation/src/jobhunter/infrastructure/materials/voice_adapter.py
All checks passed!

$ uv --project workers/automation run --extra dev pytest -q   # full worker suite
1495 passed in 38.50s
\`\`\`

### DIFF
--- a/workers/automation/src/jobhunter/infrastructure/analysis/claude_analysis_adapter.py
+++ b/workers/automation/src/jobhunter/infrastructure/analysis/claude_analysis_adapter.py
@@ -92,14 +92,25 @@ def _structured_output_from_messages(messages: list[Any]) -> dict[str, Any]:
     return structured
 
 
+def _optional_int(value: Any) -> int | None:
+    """Coerce an SDK usage field to ``int``, or ``None`` when absent/unparseable."""
+    if value is None:
+        return None
+    try:
+        return int(value)
+    except (TypeError, ValueError):
+        return None
+
+
 def _usage_from_messages(messages: list[Any]) -> tuple[int | None, int | None]:
     """Best-effort ``(input_tokens, output_tokens)`` from the final ResultMessage.
 
     The Claude Agent SDK reports authoritative cumulative usage on the terminal
     ``ResultMessage``; the true input is fresh + cache-creation + cache-read
     tokens (reading only ``input_tokens`` under-reports the cached system prompt).
-    Returns ``(None, None)`` when the SDK surfaced no usage so the span omits
-    token counts rather than fabricating them.
+    Every field is coerced defensively so a drifted/non-int usage shape yields
+    omitted counts rather than raising inside the span (telemetry must never fail
+    the leg). Returns ``(None, None)`` when the SDK surfaced no usage.
     """
     for message in reversed(messages):
         if type(message).__name__ != "ResultMessage":
@@ -108,11 +119,11 @@ def _usage_from_messages(messages: list[Any]) -> tuple[int | None, int | None]:
         if not isinstance(usage, Mapping):
             return None, None
         input_tokens = (
-            int(usage.get("input_tokens") or 0)
-            + int(usage.get("cache_creation_input_tokens") or 0)
-            + int(usage.get("cache_read_input_tokens") or 0)
+            (_optional_int(usage.get("input_tokens")) or 0)
+            + (_optional_int(usage.get("cache_creation_input_tokens")) or 0)
+            + (_optional_int(usage.get("cache_read_input_tokens")) or 0)
         )
-        output_tokens = int(usage.get("output_tokens") or 0)
+        output_tokens = _optional_int(usage.get("output_tokens")) or 0
         return (input_tokens or None, output_tokens or None)
     return None, None
 
--- a/workers/automation/src/jobhunter/infrastructure/analysis/claude_analysis_adapter.py
+++ b/workers/automation/src/jobhunter/infrastructure/analysis/claude_analysis_adapter.py
@@ -18,7 +18,7 @@ cancellation of the wrapping asyncio task.
 from __future__ import annotations
 
 import json
-from collections.abc import AsyncIterator, Awaitable, Callable
+from collections.abc import AsyncIterator, Awaitable, Callable, Mapping
 from typing import Any
 
 from jobhunter.domain.materials.analysis import (
@@ -26,6 +26,7 @@ from jobhunter.domain.materials.analysis import (
     JobAnalysisDraft,
 )
 from jobhunter.infrastructure.analysis.prompts import build_synthesizer_user_prompt
+from jobhunter.infrastructure.observability.llm_spans import llm_generation_span
 
 # A query callable: (prompt, options) -> async iterator of SDK messages.
 QueryFn = Callable[..., AsyncIterator[Any] | Awaitable[AsyncIterator[Any]]]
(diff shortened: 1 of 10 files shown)


workers/automation/src/jobhunter/infrastructure/analysis/claude_analysis_adapter.py
```diff
@@ -86,6 +92,31 @@ def _structured_output_from_messages(messages: list[Any]) -> dict[str, Any]:
...
+    token counts rather than fabricating them.
+    """
+    for message in reversed(messages):
+        if type(message).__name__ != "ResultMessage":
+            continue
+        usage = getattr(message, "usage", None)
+        if not isinstance(usage, Mapping):
+            return None, None
+        input_tokens = (
+            int(usage.get("input_tokens") or 0)
```
ebarti: PR Reviewer - pr-reviewer

Severity: Medium
Gate: Pass
Evidence: `_usage_from_messages` (here, lines 111-115) and its verbatim twin in `infrastructure/materials/voice_adapter.py:105-109` parse SDK usage with bare `int(usage.get(...) or 0)`. This runs INSIDE the `llm_generation_span` `with` block, whose handler is `except Exception: span.set_status(ERROR); span.record_exception(exc); raise` — i.e. it re-raises into the leg. So a non-int usage value fails the leg (then retries via `_draft_with_retry`; if it persists across the three Claude-based calls — draft, synthesizer, voice pass — it degrades the whole analysis/voice). By contrast `codex_analysis_adapter._usage_from_result` and `antigravity_analysis_adapter._usage_from_metadata` coerce every field through a defensive `_optional_int` and never raise.
Reproduction: Reproduced with a script against the PR head: `_usage_from_messages([ResultMessage(usage={"input_tokens":"n/a","output_tokens":10})])` raises `ValueError: invalid literal for int() with base 10: 'n/a'`; the Codex/Antigravity extractors return `(None, 10)` / `(None, 6)` on the same malformed shape.

This is the one place the PR's stated primary invariant — "a span/telemetry error can never fail or skip a leg" — does not hold. Token extraction is not isolated from the leg on the Claude/voice path, so a malformed/drifted SDK usage shape can fail a leg through the span block. The repo explicitly treats SDK shape drift as expected (see the repeated "model ids drift / re-confirm against the SDK at impl time" comments), and the docstrings only handle ABSENT usage (returns None), not MALFORMED usage.

Required fix: make `_usage_from_messages` defensive like the Codex/Antigravity extractors (coerce each field via a try/except int helper), OR wrap the token-extraction + `record(...)` in a non-fatal `try/except` that logs and continues so no telemetry operation can propagate into the leg. Prefer the latter to guarantee the invariant for all future telemetry. Add a test with a malformed usage value asserting the leg still succeeds and the span omits tokens.

workers/automation/tests/test_antigravity_adapter.py
```diff
@@ -26,6 +31,23 @@
 pytestmark = pytest.mark.asyncio
 
 
+@pytest.fixture
+def in_memory_exporter(monkeypatch):
```
ebarti: PR Reviewer - pr-reviewer

Severity: Low
Gate: Pass
Evidence: This `in_memory_exporter` fixture is duplicated verbatim (identical `_TRACER_PROVIDER_SET_ONCE`/`Once()` monkeypatch + `InMemorySpanExporter` body) in `tests/test_employer_analysis_ensemble.py:43` and `tests/test_voice_adapter.py:30`, and an identical fixture already exists in `tests/test_llm_spans.py:15`. That is four copies of the same fixture.
Reproduction: Reasoned from code (four identical fixture definitions).

Duplicated test fixture — DRY/maintainability. Not a correctness issue; the fixture works and the full suite passes.

Required fix: promote the fixture to a shared `tests/conftest.py` and remove the duplicated copies. Non-blocking.

workers/automation/src/jobhunter/infrastructure/analysis/claude_analysis_adapter.py
ebarti: Fixed in cdce81ef. On both the Claude draft/synthesizer path (`claude_analysis_adapter._usage_from_messages`) and the voice path (`voice_adapter._usage_from_messages`), every usage field is now coerced through the same defensive `_optional_int` helper the Codex/Antigravity legs already use, so a drifted / non-int SDK usage shape yields omitted token counts instead of raising inside the re-raising `llm_generation_span` block. Added regression tests `test_draft_span_survives_malformed_sdk_usage` and `test_rewrite_span_survives_malformed_sdk_usage` that call the leg with `usage={"input_tokens":"n/a","output_tokens":10}` and assert it still succeeds (unparseable input omitted, well-formed output still recorded) — malformed usage no longer fails a leg. Full worker suite: 1497 passed.

workers/automation/tests/test_antigravity_adapter.py
ebarti: Fixed in cdce81ef. Promoted `in_memory_exporter` to `workers/automation/tests/conftest.py` and removed the four in-PR copies (`test_llm_spans.py`, `test_employer_analysis_ensemble.py`, `test_voice_adapter.py`, `test_antigravity_adapter.py`) plus their now-unused OTel imports.

PR #214, opened 2026-07-02
fix(enrichment): preserve enriched description on LinkedIn apply-URL retry

ebarti:
## What

Fixes a data-loss bug in the LinkedIn authenticated enrichment retry path: an already-enriched LinkedIn job that was merely missing its external apply URL could lose its canonical `full_description` when the authenticated re-scrape failed.

## Why (root cause)

**Invariant:** a re-tailor/retry must not destroy the current reviewable material (CLAUDE.md "Root-Cause And Auditability Discipline"). The enriched `full_description` feeds scoring and tailoring.

On every enrich run, `_reset_authenticated_linkedin_retry_candidates` (`workers/automation/src/jobhunter/enrichment/detail.py`) called `aggregate.reset()` for LinkedIn rows that were **already enriched** but missing an apply URL. `JobEnrichment.reset()` nulls `full_description` **and** `application_url`, and `SqliteEnrichmentRepository.save` commits eagerly (`sqlite_repository.py:175`) **before** the pending-row re-scrape runs. When the authenticated re-scrape then failed or returned worse text (common on LinkedIn login walls / logged-out profiles), the description was already committed as `NULL` and permanently lost.

The destructive `reset()` is load-bearing for re-queuing genuinely non-enriched / failed rows, so it could not simply be removed.

## How

- New domain method `JobEnrichment.backfill_application_url(application_url, updated_at)` — updates **only** `application_url` (+ `updated_at`), requires an `enriched` aggregate, and preserves `full_description`, the `enriched` status, `enriched_at`, extraction tier, and attempt history.
- `_reset_authenticated_linkedin_retry_candidates` now routes candidates:
  - **Enriched but missing apply URL** -> non-destructive backfill. It resolves the apply URL via the existing `_apply_authenticated_linkedin_apply_url` (pageless, `page=None` -> `resolver.resolve(url)`); on a hit it calls `backfill_application_url`, on a miss / raise it leaves the aggregate untouched (description preserved). The enrich stage stays `succeeded`; an informational `StageProgress` audit event is recorded on backfill.
  - **Failed / non-enriched** -> unchanged destructive `reset()` + re-queue to `pending`.
- The authenticated resolver is injected via a `resolver_factory` (built lazily only when an enriched candidate exists, and closed in a `finally`); production wires `_default_linkedin_apply_resolver_factory` from `_run_detail_scraper`.
- QA regression-matrix entry updated in `docs/local-reliability-qa.md` to add the description-preservation invariant.

No user-facing CLI/API surface changes; the enrich command behaves the same except it no longer destroys enriched descriptions on a failed apply-URL retry.

## How validated

```
  workers/automation/tests/test_linkedin_authenticated_enrichment_retry.py \
  workers/automation/tests/test_enrichment_use_cases.py \
  workers/automation/tests/test_enrichment_detail_staleness.py
# 2 failed, 21 passed
```

The 2 failures (`test_scrape_detail_page_reports_expired_json_ld_as_inactive`, `test_scrape_detail_page_reports_closed_marker_as_inactive`) are **pre-existing and environmental** — `RuntimeError: No LLM provider configured` in `scrape_detail_page`'s LLM tier, a path this change never touches. Confirmed identical failures on clean `HEAD` (via `git stash`). All other tests in those files pass, including the 8 in the retry file.

```
uv --project workers/automation run --extra dev pytest -q workers/automation/tests/test_enrichment_aggregate.py
# 26 passed  (incl. 2 new backfill_application_url tests)

  test_enrichment_aggregate test_enrichment_queue_selectors test_enrichment_repository \
  test_enrichment_use_cases test_enrichment_events_pr3 test_enrichment_snapshot_pr3 \
  test_activity_enrich test_linkedin_authenticated_enrichment_retry
# 85 passed

  workers/automation/src/jobhunter/enrichment/detail.py \
  workers/automation/src/jobhunter/domain/enrichment/aggregate.py \
  workers/automation/tests/test_linkedin_authenticated_enrichment_retry.py \
  workers/automation/tests/test_enrichment_aggregate.py
# All checks passed!
```

New regression tests (`test_linkedin_authenticated_enrichment_retry.py`):
- enriched + missing apply URL, recovery **fails** (resolver returns no URL) -> description + enriched status retained, apply URL still `None`.
- enriched + missing apply URL, resolver **raises** -> description + enriched status retained.
- enriched + missing apply URL, recovery **succeeds** -> apply URL backfilled, description preserved, still enriched.
- genuinely failed row -> still `reset()` to `pending` (resolver never consulted).
- updated the existing test: without a resolver, the enriched-missing row is preserved (not reset).


workers/automation/src/jobhunter/enrichment/detail.py
```diff
@@ -1309,54 +1327,127 @@ def _reset_authenticated_linkedin_retry_candidates(
...
+                            if aggregate.full_description
+                            else ""
+                        ),
+                        "application_url": None,
+                    },
+                    resolver=resolver,
+                    page=None,
+                )
+                apply_url_value = resolved.get("application_url")
+                if not apply_url_value:
```
ebarti: PR Reviewer - pr-reviewer

Severity: Medium
Gate: Pass
Evidence: The attempt-count bound (`_attempt_count_from_json(attempts_json) >= _MAX_AUTHENTICATED_LINKEDIN_RETRY_ATTEMPTS`, =3) is read from the enrichment `attempts_json`. On the new backfill path an empty/failed resolution just `continue`s here — it never records an attempt, never touches `attempts_json`, and (unlike a successful backfill) never updates `updated_at`. So an already-enriched LinkedIn row that is missing an external apply URL keeps its original attempt_count (typically 1 for a cleanly-enriched row) forever and is re-selected by the candidate query on every enrichment run. The docstring for this function claims "bounded by attempt count so repeated enrich runs do not loop forever," but that bound now holds only for the reset path; the backfill path is effectively unbounded.
Reproduction: Reasoned from code + confirmed the query re-selects the row (enriched AND application_url IS NULL) and that `backfill_application_url`/the empty-resolution continue leave attempts_json and updated_at unchanged. The 4 new tests each run a single pass, so none exercise the cross-run repetition.

Impact: LinkedIn "Easy Apply" postings have no external apply URL by design, so this is the common case, not an edge. Every such enriched row drives the authenticated `LinkedInApplyUrlResolver` (real Chromium launch + LinkedIn navigation + the 6s linkedin site delay + nav/load timeouts) on every run, returning empty each time. That is wasted work every run and — because it repeatedly automates an authenticated LinkedIn session against the same never-resolving jobs — a real rate-limit / bot-detection risk to the user’s LinkedIn account. Pre-fix, these rows were bounded to 3 authenticated attempts (via reset+re-scrape) before stopping; this PR correctly stops destroying the description but drops that bound. This borders High on the account-safety angle; I am rating Medium because it is not a data-loss/preservation defect (the critical invariant is fully met) and per-run work is still capped by `limit`.

Required fix: bound the backfill retries too — e.g. on an empty/failed authenticated resolution record a lightweight authenticated-apply attempt (so `_attempt_count_from_json` climbs toward `_MAX` and the row is excluded after N tries), or stamp a "last authenticated apply attempt" timestamp and add a cooldown to the candidate query so a never-resolving Easy-Apply row is not re-driven every run. Add a regression test asserting a row that fails resolution is not re-attempted once the bound/cooldown is reached.

ebarti: Fixed in a91bb3eb (kept the description-preservation; did NOT reintroduce the destructive reset).

The backfill path now records a lightweight authenticated attempt on **every** pass, so the candidate query's existing `_MAX` (3) attempt-count gate bounds recovery exactly like the extraction cascade:

- Replaced `backfill_application_url` with `JobEnrichment.record_apply_url_recovery`: appends a terminal `EnrichmentAttempt` (succeeded when an external apply target is found + attached, failed with `APPLY_URL_UNRESOLVED` otherwise) and bumps `updated_at`. It **never** transitions the aggregate to `failed` and never touches `full_description` / enriched status / `enriched_at` / top-level extraction tier.
- `_reset_authenticated_linkedin_retry_candidates` now saves the recovery outcome on both hit and miss (previously it silently `continue`d on a miss, leaving the count static). A miss records the bounding attempt; the description stays intact.
- Trace for a never-resolving Easy-Apply row (starts at 1 enrichment attempt): run1 -> attempt 2, run2 -> attempt 3, run3+ -> skipped by `_attempt_count_from_json >= _MAX`. So the authenticated Chromium browser is driven at most `_MAX - 1` = 2 recovery times, then never again. Equivalent to the pre-fix bound, minus the data loss.
- Per-attempt status is internal only: the projection reads solely top-level aggregate fields and the enrich stage badge is driven by `job_stage_states` (left `succeeded`), so no contradictory "enriched + failed attempt" UI surface.

New cross-run regression `test_enriched_missing_apply_url_recovery_is_bounded_across_runs`: drives 5 runs against a never-resolving resolver and asserts `resolver.calls == _MAX - 1`, `attempt_count == _MAX`, and `full_description` intact the whole time. Plus aggregate unit tests for both recovery outcomes.

Verification:
```
pytest -q test_linkedin_authenticated_enrichment_retry.py test_enrichment_aggregate.py  ->  36 passed
broad enrichment sweep (8 files incl. _run_detail_scraper path)                        ->  87 passed
ruff check (changed src + test files)                                                  ->  All checks passed!
```

ebarti:
## Appended: offline determinism for enrichment staleness tests

Fixes the test-hygiene finding that `test_scrape_detail_page_reports_expired_json_ld_as_inactive` and `test_scrape_detail_page_reports_closed_marker_as_inactive` failed on a keyless machine with `RuntimeError: No LLM provider configured`, so the suite was only green with live API keys.

**Root cause (not the LLM call itself):** `scrape_detail_page` builds its three-tier cascade tuple *eagerly*, so `_make_llm_extractor()` runs during tuple construction — before the extraction loop. That construction resolves a real provider (`get_llm_adapter()` -> `LLMClient` -> `_detect_provider()`), which raises when no keys are set. These two cases actually win at **Tier 1 (JSON-LD)** / **Tier 2 (CSS)** and never *invoke* the LLM tier, but the tier is still *constructed*, and that construction was the failing step offline.

**Fix (test-only):** an `offline_llm_tier` fixture swaps the port the detail cascade resolves through (`detail.get_llm_adapter`) — the narrowest hexagonal seam — with an `_OfflineLlmPort` stub, mirroring the existing `_StubLlm(LlmPort)` convention in `test_enrichment_extractors.py`. The real `_make_llm_extractor` and `LlmExtractor` stay in the path. The stub raises if the LLM tier is ever actually invoked, preserving the regression signal that staleness detection must be proven by Tier 1/Tier 2 and must not silently reroute through the model. No production behavior or fallback-order change.

**Verification**
- Targeted file, keys stripped: `env -u ANTHROPIC_API_KEY -u OPENAI_API_KEY -u GOOGLE_API_KEY -u GEMINI_API_KEY -u GOOGLE_GENAI_API_KEY -u LLM_URL -u LLM_API_KEY -u LLM_MODEL uv --project workers/automation run --extra dev pytest -q workers/automation/tests/test_enrichment_detail_staleness.py` -> **8 passed** (was 2 failed, 6 passed).
- Full suite in a keyless ambient env: `uv --project workers/automation run --extra dev pytest -q` -> **1494 passed**.

Commit: `05e2fa35` — `test(enrichment): stub LLM tier so staleness tests run offline`

PR #217, opened 2026-07-02
feat(api): add loopback Host-header allowlist and harden artifact-open path validation

ebarti:
Stacked on #210; review/merge after #210.

## What changed

Two structural hardening items for the loopback-bound local Fastify API (single-user, no auth), both scoped to `apps/api`:

1. **DNS-rebinding defense (Host-header allowlist).** A new `onRequest` hook rejects any request whose `Host` header is not a loopback host — `127.0.0.1`, `localhost`, or `[::1]`, each with an optional `:port` — with `403 forbidden_host`. A rebinding attack page keeps its own hostname in the `Host` header, so it is blocked; the local web app and CLI always send a loopback `Host`, so legitimate access is unaffected. Loopback binding behavior is unchanged. The allowlist lives in `apps/api/src/local-origin.ts` (`isLoopbackHostHeader` + `LOOPBACK_HOST_PATTERNS`), mirroring the existing origin-guard style; the existing fail-open `isTrustedMutationSource` is untouched.

2. **Artifact path containment.** `GET /v1/artifacts/:artifactId/preview.pdf`, `.../preview.html`, `.../preview/page/:pageNumber.png`, and `POST .../open` now confirm the resolved artifact path (`realpath`) stays inside the JobHunter app directory before streaming bytes or handing the path to the OS opener, returning `403 artifact_path_forbidden` otherwise. Both the app dir and the target are `realpath`'d (handles the macOS `/var`→`/private/var` symlink), and containment is checked with `path.relative` (avoids the naive prefix-match escape). Registered-artifact lookup via the projection is unchanged, so a tampered projection row or a symlink escape can no longer read or launch files outside the app directory.

## Why

The API serves the full profile, resume PDF bytes, and job history and triggers actions, but had no Host-header allowlist and no `realpath` containment on the artifact routes that touch the filesystem. Both are concrete, in-scope hardening items for a local-first single-user app.

## Out of scope (deliberate follow-up)

Full bearer-token authentication and keychain-to-worker credential passing are intentionally **not** in this PR: they require coordinated frontend + dev-launcher changes and local-dev verification. `apps/api/src/config.ts` loopback-bind behavior (fail-open origin guard included) is unchanged.

## Tests

Added to `apps/api/test/server.test.ts`:
- Foreign `Host` header (`evil.example.com`, `attacker.test:8766`) on `GET /v1/health` and `GET /v1/profile` -> `403 forbidden_host`.
- Foreign-`Host` mutation is blocked before the handler runs (DB state unchanged) even with a loopback `Origin`.
- Loopback `Host` variants (`localhost`, `127.0.0.1`, `[::1]`, each with/without port) -> `200`.
- `POST .../open` on an unregistered artifact id -> `404 artifact_not_found`, opener not called.
- `POST .../open` and `GET .../preview.pdf` for an artifact whose resolved path is outside the configured app dir -> `403 artifact_path_forbidden`, opener not called.

Meaningfulness verified by temporarily neutering both guard helpers: exactly the 4 guard-dependent tests flipped `403 -> 200` while the 2 control tests stayed green, then reverted.

## Validation

- `pnpm api:check` -> pass (tsc `--noEmit`, includes `test/**/*.ts`).
- `pnpm api:test` -> `Test Files 18 passed (18)`, `Tests 297 passed (297)`.

## Docs

- `docs/local-ts-api.md`: security note on the Host allowlist + artifact-path containment, and the bearer-token follow-up.
- `docs/local-reliability-qa.md`: regression-matrix row mapping the invariant to the new tests.


apps/api/test/server.test.ts
```diff
@@ -496,6 +496,66 @@ describe("local TypeScript API", () => {
     await app.close();
   });
 
+  it("rejects requests whose Host header is not a loopback host", async () => {
```
ebarti, line 499: PR Reviewer - pr-reviewer

Severity: Low
Gate: Pass
Evidence: The Host-allowlist tests use only obviously-foreign hosts ("evil.example.com", "attacker.test:8766"). They do not pin the actual DNS-rebinding threat model — a hostname that is a loopback token with a suffix/prefix, e.g. "localhost.evil.com", "127.0.0.1.evil.com", "evil.localhost", a trailing-dot FQDN ("localhost."), or comma/space multi-value injection ("127.0.0.1, evil.com"). I verified via the compiled patterns that all of these are correctly rejected (the patterns are anchored ^...$ and isLoopbackHostHeader requires exactly one value), and that the six legitimate forms are accepted — so this is not a defect, only a missing regression pin on the security-critical control. Separately, the artifact-containment escape test covers only the open and preview.pdf routes; the identical guard on preview.html and preview/page/:n.png (server.ts:1636, 1678) has no dedicated escape test.
Reproduction: Confirmed by replicating LOOPBACK_HOST_PATTERNS + isLoopbackHostHeader in node against ~30 inputs (all legit accepted, all foreign/rebinding/injection/missing rejected) and by api:test (297 passed).

Required fix (optional, non-gating): add a rejection case for at least "localhost.evil.com" / "127.0.0.1.evil.com" (the rebinding shapes this hook exists to stop), and add a containment-escape assertion for the preview.html and page-PNG routes so a future refactor that drops a guard on one route is caught. Neither changes behavior; both harden the regression net for a security control.

Note (not a finding): expanded IPv6 loopback forms ("[0:0:0:0:0:0:0:1]", bare "::1", "[::ffff:127.0.0.1]") are rejected; real browsers/CLIs send "[::1]" so this is not a legitimate-access break, and stricter is the safe direction.

ebarti: Done in d61a329f (test-only, no production change).

DNS-rebinding subdomain shapes are now pinned: a dedicated test asserts 403 `forbidden_host` for `localhost.evil.com` and `127.0.0.1.evil.com`, plus the extra shapes you verified — trailing-dot `localhost.`, multi-value `127.0.0.1, evil.com`, and IPv4-mapped `[::ffff:127.0.0.1]`.

Artifact containment escape cases now cover all four guarded routes: `open` + `preview.pdf` (existing) and `preview.html` + `preview/page/:pageNumber.png` (new). Each asserts 403 `artifact_path_forbidden` and that the file is never streamed / the PDF page renderer is never invoked.

Verified: `pnpm api:check` clean; `pnpm api:test` -> 300 passed (18 files, +3 new). Also confirmed the new tests are load-bearing by temporarily neutering both guard helpers (all 3 new tests flip 403 -> 200) then reverting.

PR #219, opened 2026-07-02
test(projections): assert full cross-runtime column parity for list/detail/dashboard

ebarti:
## Stacked on #207

**Base is #207's branch (`fix/python-projection-score-audit-columns`), not `main` — review/merge after #207.** (The task-lead's suggested base ref `worktree-agent-a6cf1242da62fc50c` does not exist; #207's actual head branch is `fix/python-projection-score-audit-columns`.)

## The gap (verified)

The cross-runtime projection parity test asserted only the **4 audit JSON columns** (`employer_analysis_json`, `bullet_provenance_json`, `coverage_audit_json`, `voice_pass_json`). But the Python `ProjectionBuilder` and the TS `refreshProjections` dual-write **~30 more columns** — stage state, scores (incl. the `score_criteria_json` / `score_trace_json` / `score_correction_json` added in #207), compensation, apply status, dashboard counts, `current_substage` — with **no cross-runtime assertion and no column-set guard**. That is exactly why the Python-omits-score-columns drift (fixed in #207) shipped undetected.

## What changed

- **Shared fixture** (`packages/domain-types/test/fixtures/audit_projection_parity.json`): seeds the full canonical rows (scores, stage states, employer analysis, bullet provenance, materials) and adds an `expectedProjections` block with the full dual-written column set for `job_list_projections` (35 cols), `job_detail_projections` (16 cols), and `dashboard_projections` (10 cols), plus `projectionParity` metadata (`jsonColumns`, `nonDeterministicColumns`).
- **Both parity tests** (Python + TS): each seeds the shared rows, runs its **own** builder, and asserts the full projection column set matches key-for-key.
  - `*_json` columns are compared **parsed** (the two runtimes serialise with different whitespace/key order).
  - Wall-clock columns (`last_updated_at` / `generated_at`) are asserted **present** but excluded from value comparison.
  - **Column-set guard:** each runtime's emitted columns must equal the fixture's expected keys plus the wall-clock columns, so a one-sided column addition in **either** builder fails against the shared fixture.
- The pre-existing 4-audit-column assertions and the TS read-model DTO assertions are retained.
- Docs: `docs/architecture.md` (cross-runtime drift guard) and `docs/local-reliability-qa.md` (regression matrix row) updated to the expanded scope.

The `expectedProjections` values were derived **empirically**: both builders were run on the seed and confirmed to agree column-for-column (35 / 16 / 10) before being committed as the expectation.

## Verification

- `uv --project workers/automation run --extra dev pytest -q workers/automation/tests/test_audit_projection_parity.py workers/automation/tests/test_projection_builder.py` → **15 passed**
- `pnpm api:test` → **286 passed** (17 files, includes `audit-projection-parity.test.ts`)
- `pnpm api:check` → clean
- `uv --project workers/automation run --extra dev ruff check workers/automation/tests/test_audit_projection_parity.py` → clean
- `pnpm --filter @jobhunter/domain-types test` → **170 passed**; `check` clean
- `scripts/check-domain-type-parity.py` → **fails, but pre-existing and unrelated** (domain-event union drift; identical on the clean base branch, no projection columns involved).

### Negative drift checks (the point of the guard)

- Dropped `score_correction_json` from the Python writer (reproducing the #207 class) → **Python value parity fails** (`{'score_correction_json': None} != {...}`). Restored.
- Added a phantom column to the TS `job_list_projections` schema → **TS column-set guard fails** (`phantom_drift_col` in the diff). Restored.

## Latent divergences found (out of scope; not fixed here so the test stays green on current code)

These are real value-derivation differences the fixture deliberately avoids; each deserves its own fix PR, after which the fixture can be tightened:

1. **`job_list_projections.location`** — Python writes it raw; TS writes `normalizeJobLocation(location)`. The fixture uses a normalization fixed point (`"Berlin, Germany"`) so both agree.
2. **`dashboard_projections.ready`** — TS additionally requires `has_resume === 1`; Python does not. The fixture uses a fully-applied job (`current_state = succeeded`) so `ready = 0` on both, sidestepping it.


packages/domain-types/test/fixtures/audit_projection_parity.json
```diff
@@ -1,13 +1,128 @@
...
   "job": {
     "url": "https://example.com/jobs/parity-developer-platform",
     "title": "Senior Platform Engineer",
-    "site": "ParityCo",
+    "company": "ParityCo",
+    "site": "linkedin",
+    "strategy": "jobspy",
+    "location": "Berlin, Germany",
```
ebarti: PR Reviewer - pr-reviewer

Severity: Medium
Gate: Pass
Evidence: The fixture deliberately seeds two columns at cross-runtime fixed points so the new full-column guard stays green: job.location = "Berlin, Germany" (a normalizeJobLocation fixed point -- TS normalizes at projections.ts:1797, Python writes it raw at projection_builder.py:601) and a fully-applied job so dashboard.ready = 0 (TS requires has_resume===1 at projections.ts:2409, Python omits that check at projection_builder.py:1306). Both are genuine cross-runtime value-derivation divergences. They are documented in the PR description but NOT in-code: grep of the added lines finds no divergence/known-gap/follow-up note in the fixture `description`, the test files, or the docs updates.
Reproduction: Reasoned from code (verified both divergences at the cited lines). Independently confirmed the guard itself has teeth: corrupting the shared fixture fails BOTH runtimes (value assertion at test_audit_projection_parity.py:352; column-set guard at :348; TS likewise).

My call: landing the guard now with the fixture routed around these two pre-existing production bugs is correct -- asserting the divergent values would either block this test-only PR on unrelated production fixes (scope creep) or ship red. This does NOT fail the gate. But the coverage gap should be discoverable by someone reading the test, not only in the (ephemeral post-merge) PR description: a later fixture edit to a non-normalized location or a pending-state job would trip the divergence with a non-obvious cause, and as written the guard reads as if location/ready are fully parity-covered.

Required fix: Add an in-code marker of the known gap -- e.g. a `projectionParity.knownDivergences` block in the fixture (or comments on the location value and the dashboard ready value) naming the two columns, the exact runtime difference, and the follow-up that will tighten the fixture once the divergences are fixed. Test-only, no behavior change; preserves the auditability invariant of labeling what the guard does and does not currently cover.

ebarti: Done in f45c770a (test/fixture only, no behaviour change). Added a `projectionParity.knownDivergences` block to the shared fixture naming each divergent column, the Python-vs-TS difference, the fixture workaround that keeps it parity-safe, and the follow-up:

- `job_list_projections.location` — Python writes it raw, TS writes `normalizeJobLocation(...)`; fixture uses a normalization fixed point ("Berlin, Germany").
- `dashboard_projections.ready` — TS additionally requires `has_resume==1`; fixture uses a fully-applied job so `ready==0` on both.

Both parity tests now carry a comment pointing at that block at the assertion site, so editing `job.location` to a non-normalized value or the apply stage to pending points a maintainer straight to the divergence + follow-up rather than a mystery failure. Production fix tracked as B2b.

ebarti:
Rebased onto the new #207 tip and appended B2b (three real cross-runtime divergences converged).

**Rebase**: `fix/projection-parity-full` rebased cleanly onto the advanced `fix/python-projection-score-audit-columns` (now carrying the B3 watermark commit); pushed with `--force-with-lease`.

**Appended** — commit `b7fe345f` `fix(projections): converge location, ready, hidden-jobs across runtimes`. The parity test previously routed the fixture AROUND these via `knownDivergences`; they are now genuinely asserted. Canonical = existing TS behavior, so **TS production is unchanged** — only Python builder + shared fixture + tests changed.

- `job_list_projections.location`: Python wrote `jobs.location` verbatim while TS wrote `normalizeJobLocation(...)`. Ported the normalizer to `infrastructure/projections/location_normalization.py` (faithful mirror of `apps/api/src/location-normalization.ts`, kept in lockstep by this parity test) and applied it in the list projection.
- `dashboard_projections.ready`: Python now requires `has_resume == 1`, matching TS. A job parked at apply/pending with no resume is not "ready".
- Dashboard totals: Python now excludes jobs hidden via `jobhunter_hidden_jobs` (`unhidden_at IS NULL`) through a new `_hidden_projection_jobs()` helper, matching TS `rebuildDashboardProjection`. Table is owned by the TS write-model; a Python-only DB without it excludes nothing.

**Shared fixture**: `knownDivergences` removed; primary `job.location` is now non-normalized (`"Remote - Berlin, Germany"` → expected `"Berlin, Germany (Remote)"`); new `dashboardAggregateJobs` seed dashboard-only jobs — a non-hidden apply/pending-no-resume job (asserts the `ready`/has_resume predicate) and a hidden applied job (asserts hidden exclusion across total_jobs/applied/by_source/score_distribution/funnel).

**Test evidence (reproduce-then-fix)**: with workarounds removed + aggregate jobs seeded, the pre-fix Python builder produced `total_jobs=3` (vs 2), `applied=2` (vs 1), `ready=1` (vs 0), `by_source=[[linkedin,2],[indeed,1]]` (vs `[[linkedin,2]]`), `score_distribution=[[8,1],[7,1]]` (vs `[[8,1]]`), and raw location — parity test FAILED; after the builder fix it PASSES.
- `uv --project workers/automation run --extra dev pytest -q` → **1494 passed**; `ruff check .` → clean.
- New `test_location_normalization.py` (11 cases) mirrors `apps/api/test/location-normalization.test.ts`.
- `pnpm api:check` → clean; `pnpm api:test` → **286 passed (17 files)**, including both halves of the parity test.

Note for the #223 rebase: the outcome-conversion rollup has the same hidden-jobs asymmetry today (Python `_build_outcome_conversion` reads `active_rows`, which pre-B2b did not exclude hidden; TS `buildOutcomeConversion` takes the hidden-excluded `active`). It resolves automatically once #223 rebases onto this tip because the Python conversion reuses `active_rows`; just keep the B2b `active_rows` hidden-exclusion during conflict resolution and re-run both parity suites (the shared fixture now seeds a hidden applied job).

workers/automation/src/jobhunter/infrastructure/projections/location_normalization.py
```diff
@@ -0,0 +1,95 @@
+"""normalize_job_location — Python mirror of apps/api/src/location-normalization.ts.
+
+The TS and Python projection builders both write
+``job_list_projections.location`` and MUST produce byte-identical output. Keep
+this module in lockstep with the TS ``normalizeJobLocation``; the cross-runtime
```
ebarti: PR Reviewer - pr-reviewer-b2

Severity: Medium
Gate: Pass
Evidence: The lockstep pointer is one-directional. This module's docstring references `apps/api/src/location-normalization.ts`, but the canonical TS source has NO reciprocal comment naming this Python mirror (verified: grep for python/mirror/location_normalization/lockstep/parity in apps/api/src/location-normalization.ts returns no match). The cross-runtime parity fixture asserts `job_list_projections.location` on a single input ("Remote - Berlin, Germany" -> "Berlin, Germany (Remote)"), so a future TS-side behaviour change on any un-fixtured input (a new SPAIN_REGION_LABELS entry, a changed remote-token list, etc.) would silently diverge from this mirror with no failing parity test and no comment warning the TS editor a mirror exists. That is the exact silent cross-runtime divergence class this PR exists to eliminate.
Reproduction: Reasoned from code; separately confirmed both normalizers are byte-identical on 28 realistic inputs via a live tsx-vs-python diff (accented Spanish locations, remote-marker variants, empty/None, City-ST forms). Divergences appear only on pathological Unicode (leading/trailing U+FEFF, or a non-ASCII letter glued directly to a remote keyword) that does not occur in real job-location data.

The forked normalizer is currently kept in sync by human vigilance on the TS side, not by tooling.

Required fix: Add a one-line comment atop apps/api/src/location-normalization.ts pointing to workers/automation/src/jobhunter/infrastructure/projections/location_normalization.py and the parity test, so an editor of either file knows to keep both in lockstep. Optionally add parity/fixture cases covering the region-label and remote-marker branches so the parity guard (not just reviewers) catches TS-side drift.

ebarti: Fixed in 1727bdd4.

- Added a reciprocal lockstep doc-comment atop `apps/api/src/location-normalization.ts` (comment-only; TS behaviour untouched) naming this Python mirror and the shared parity table, and refreshed this module's docstring to point at the same table.
- Added a shared `locationCases` array to `packages/domain-types/test/fixtures/audit_projection_parity.json` covering the region-label expansion branch (AN/CT/MD, gated on a Spain country token) and the remote-marker branch (every remote token plus the parenthetical markers), with expected values derived from the algorithm.
- Converted BOTH `apps/api/test/location-normalization.test.ts` and `workers/automation/tests/test_location_normalization.py` to load the shared `locationCases`, deleting the previously hand-maintained lists that had already drifted (TS 7 cases vs Python 11). A TS-side change to a region label, country token, or remote token/marker now fails one of the two suites rather than diverging silently.

Validated: `pnpm api:check`, `ruff check`, `pnpm api:test` (297 passed), `uv pytest -q` (1512 passed).

PR #216, opened 2026-07-02
fix(materials): count keyword coverage only when evidence-grounded, not by requirement self-match

ebarti:
## What

The generation-time keyword-coverage audit (`domain/materials/coverage_audit.py`) — the anti-stuffing guard whose covered / declared / missing lists the user trusts — now does two things:

1. **Evidence grounding.** A planned keyword is credited **covered** only when it appears (word-boundary) in a bullet backed by **real profile evidence**:
   - the bullet carries a canonical profile evidence FK (`evidence_ids`), **or**
   - the keyword itself traces (word-boundary) to the profile evidence corpus (`build_evidence_corpus`).

   A `requirement_id` no longer credits coverage on its own.

2. **Three honest buckets (A6b).** The remaining planned keywords are split into **declared** (rendered in the resume's skills line but not demonstrated) versus **missing** (rendered nowhere the employer reads) — instead of lumping a genuinely-shipped declared skill into `missing`.

## Why (root cause)

The old rule counted a keyword as covered when it appeared in a "grounded" bullet, where grounded meant *any* `evidence_id` **OR** `requirement_id`. But the provenance builder binds a `requirement_id` to a bullet **purely by keyword match** (`provenance_builder._served_requirements` → `_contains_term`). So any planned keyword present in a line **auto-created** the requirement binding that marked the line grounded, and the keyword then counted as covered **with zero evidence check** — the guard was a near no-op for exactly the keyword set it measures, and it actively rewarded stuffed/fabricated keywords.

The clearest exploit is the skills line: it always has `evidence_ids=()`, but a `requirement_id` gets self-bound whenever a JD keyword appears in it, so keyword-stuffing the skills dump inflated audited coverage.

The circular path is real, confirmed end-to-end from canonical data through the real builder:

```
skills.generated_text  : Languages: Python, Go, Kubernetes
skills.requirement_ids : ('req_platform',)   # self-bound purely by the "kubernetes" match
skills.evidence_ids    : ()                    # no profile evidence
'kubernetes' in corpus : False
# before: 'kubernetes' counted as covered.  after (evidence grounding): not covered.
# after A6b: 'kubernetes' is 'declared' (shipped in the skills line), not 'missing'.
```

Per the auditability discipline: the requirement binding is **preserved** for provenance display; only the coverage crediting logic changed. Nothing was hidden or suppressed — a stuffed keyword no longer counts as covered, a real evidence-backed keyword still counts, and a declared-but-undemonstrated skill is reported honestly rather than as absent.

[...]

PR #220, opened 2026-07-02
fix(materials): PDF renderers honor mandatory bullet overflow so PDF matches the reviewed .txt

ebarti:
> Stacked on #210; review/merge after #210. Base is #210's head (`fix/resume-pdf-real-html-renderer`), not `main`.

## What
Render the resume PDFs (HTML and LaTeX) from the **same overflow-aware bullet set** as the reviewed plain-text resume, and keep the provenance audit trail byte-identical to it.

## Why (the bug — medium; submitted PDF weaker than the reviewed `.txt`)
The plain-text `ResumeAssembler` injects `_allow_mandatory_bullet_overflow` into its experience-updates map, so a **validated requirement-led candidate** that pins covered-achievement bullets past `max_experience_bullets` keeps every pinned bullet in the `.txt` the user reviews.

Both PDF renderers and the provenance builder built their **own** experience-updates maps *without* that flag:
- `infrastructure/materials/html_resume_pdf.py` `build_resume_document`
- `infrastructure/materials/latex_pdf.py` `build_latex`
- `domain/materials/provenance_builder.py` `build_bullet_provenance`

So `tailored_experience_bullets` trimmed those paths to the cap (default 4). The submitted PDF silently **dropped exactly the pinned requirement-coverage bullets the user reviewed** — the employer received a weaker resume than the one approved, the two render paths diverged from the assembler, and provenance (which explicitly claims byte-identity with the shipped resume) understated the shipped bullets.

## Fix
Centralize the map building in one helper — `experience_updates_by_id(tailored_payload)` in `resume_profile.py`, next to the flag it consumes and the `tailored_experience_bullets` selector. The `.txt` assembler, the HTML PDF renderer, the LaTeX PDF renderer, and the provenance builder all call it, so the mandatory-overflow allowance is injected in **exactly one place** and the paths can never select a different bullet set again. The normal `max_experience_bullets` cap still applies to non-mandatory bullets. The `.txt` assembler behavior is preserved verbatim (the helper is a direct extraction of its existing map logic).

## Scope note (no silent broadening)
`domain/materials/quality.py` `build_tailoring_change_annotations` also builds an experience-updates map, but its `tailored_text` is display-capped (`_annotation_lines(..., limit=5)`) and its claim audit matches by `location`, so it exhibits **no observable divergence** from this bug — I left it untouched rather than change code I could not back with a meaningful regression test. It is a candidate for the same consistency refactor in a follow-up.

## Tests
- `test_pdf_renderer_ports.py`
  - `test_pdf_renderers_render_all_mandatory_overflow_bullets_like_txt` — 6 pinned bullets past the cap of 4 render **identically** in the `.txt`, the HTML PDF document, and the LaTeX source (asserts the PDF experience-bullet set **equals** the `.txt` set).
  - `test_pdf_renderers_respect_max_experience_bullets_without_overflow` — no `generated_claim_mappings` → all three paths trim to `max_experience_bullets` (4) identically.
- `test_bullet_provenance.py::test_provenance_keeps_mandatory_overflow_bullets_matching_shipped_resume` — provenance emits one row per shipped bullet (6), byte-matching the assembled experience bullets.

## Validation
```
  test_resume_assembler.py test_pdf_renderer_ports.py test_resume_html_migration.py \
  test_tailor_voice_audit_integration.py test_bullet_provenance.py test_materials_quality.py
-> 91 passed

# full worker suite
uv --project workers/automation run --extra dev pytest -q workers/automation/tests
-> 1490 passed

# lint (changed files)
uv --project workers/automation run --extra dev ruff check <changed .py files>
-> All checks passed!
```

## Docs
`docs/local-reliability-qa.md` — extended the existing "Resume PDF rendering diverges from the accepted tailored payload" regression-matrix row with this failure mode and added `test_bullet_provenance.py` to its coverage list.


workers/automation/src/jobhunter/domain/materials/provenance_builder.py
```diff
@@ -300,11 +301,7 @@ def build_bullet_provenance(
         )
 
     # ---- Experience bullets (one row per rendered bullet) ----------------
-    experience_updates = {
-        str(update.get("id")): update
-        for update in tailored_payload.get("experience_updates") or []
-        if isinstance(update, dict) and update.get("id")
-    }
+    experience_updates = experience_updates_by_id(tailored_payload)
     for entry in get_experience_entries(profile):
```
ebarti, line 305: PR Reviewer - pr-reviewer

Severity: Low
Gate: Pass
Evidence: This centralizes the BULLET-level map (mandatory-overflow flag) across the .txt assembler, both PDFs, and provenance — verified: all four call `tailored_experience_bullets(entry, update, profile)` with this shared map, and the parity tests have teeth (reverting the HTML renderer to its own flagless map fails `test_pdf_renderers_render_all_mandatory_overflow_bullets_like_txt`). However, entry-level selection is NOT centralized: the .txt assembler (services.py:543), HTML PDF (html_resume_pdf.py:556), and LaTeX PDF (latex_pdf.py:378) filter experience entries to `required_experience_entry_ids`, but this provenance loop iterates ALL `get_experience_entries(profile)` with no such filter.
Reproduction: Reasoned from code. `get_required_experience_entry_ids` defaults to all entries when unset, so there is no divergence in the common case; but when `required_experience_entry_ids` is an explicit STRICT SUBSET, provenance (and coverage #216, which is computed over these rows) emits rows for bullets in entries the resume does not ship — the A4 entry-subset inflation.

This PR (A3) is correctly scoped to bullet overflow and this is not a regression it introduces — provenance never filtered entries. But it means the "byte-identical to the shipped resume" claim holds at the bullet level and NOT at the entry level for the strict-subset case, so A4 is still needed.

Required fix: none in this PR. Follow-up (A4): apply the same `required_experience_entry_ids` entry filter here (and confirm coverage) so provenance reflects only the shipped entries.

ebarti: Delivered by **#224** (`fix/provenance-shipped-entries`, stacked on this PR): the provenance loop now applies the same `required_experience_entry_ids` entry filter as the .txt assembler and both PDF renderers, with regression tests proving an omitted entry's bullets emit no provenance rows and its keywords stay uncredited (`test_strict_subset_provenance_and_coverage_reflect_only_shipped_entries`). Resolving.

PR #223, opened 2026-07-02
feat(operations): outcome-conversion projection (reply/interview/offer rates by source and score band)

ebarti:
Stacked on #219; review/merge after #219 (and #207).

## The gap (audit #1 product gap — the dead learning loop, surface half)

Gmail/manual outcome feedback (reply / interview / offer / rejection) was captured in `application_outcomes`, but nothing rolled it into rate metrics. The dashboard exposed only stage counts, source breakdown, and score distribution — no funnel **conversion** — so the user could not see where the funnel converts or which sources / score bands actually produce interviews.

## What changed (SURFACE ONLY — no behavior change)

- **New materialised projection column `dashboard_projections.outcome_conversion_json`**, computed in BOTH builders (Python `ProjectionBuilder._build_outcome_conversion` and TS `buildOutcomeConversion` in `refreshProjections`). It holds the funnel — `applied → reply → interview → offer` plus `rejection` — as **integer counts**, broken down **by source** and **by score band**.
  - **Score bands** mirror the user-facing scoring criteria (`SCORE_PROMPT`): 9-10 perfect, 7-8 strong, 5-6 moderate, 3-4 weak, 1-2 poor, plus `unscored`.
  - **Denominator** = applied jobs (same predicate as the existing `applied` counter). Outcome kinds are cumulative: reply ⊇ {recruiter_reply, interview, assessment, offer, rejection}, interview ⊇ {interview, assessment, offer}, offer ⊇ {offer}, rejection = {rejection}. So `reply ≥ interview ≥ offer` within each group.
  - **Counts only** are materialised, so the two runtimes agree byte-for-byte (no cross-runtime float drift).
- **Cross-runtime parity (#219):** `outcome_conversion_json` is added to the shared full-column parity fixture (`packages/domain-types/test/fixtures/audit_projection_parity.json`) — `jsonColumns.dashboard` + `expectedProjections.dashboard`. Both `test_audit_projection_parity.py` and `audit-projection-parity.test.ts` (column-set guard + value parity) now assert both builders produce the same column.
- **Read-model API:** `DashboardSummary` gains a `conversion` section (`totals` + `bySource` + `byBand`). Rates (reply/interview/offer/rejection over applied) are **derived from the counts in the read model** — `null` on zero-applied (no divide-by-zero). `costPerInterview` is `null` because per-run apply cost is not projected into the read model today (`apply_run_projections` has no cost column and no apply `job_events` payload carries `cost_usd`); the field/seam is in place for when it is.
- **Existing-DB migration:** the new column is added to the Python `SCORE_EVIDENCE_COLUMNS` migration list and the TS `ensureProjectionColumn` set, so local DBs that predate it self-heal on the next refresh.

This PR does **not** touch scoring, ranking, thresholds, or tailoring/apply eligibility. Acting on these metrics is a separate, user-gated decision.

## Follow-up (D3b)

Frontend rendering of the conversion section (a Dashboard funnel-by-source/band widget) is a follow-up. Projecting per-run apply cost so `costPerInterview` becomes non-null is a separate change (touches the apply lifecycle projection).

## Tests

- `workers/automation/tests/test_dashboard_projection.py`: seeds `application_outcomes` across multiple sources + score bands; asserts applied/reply/interview/offer/rejection counts per source and per band; empty/zero-applied case.
- `apps/api/test/projections.test.ts`: `/v1/dashboard/summary` returns the conversion section with correct counts + derived rates + `costPerInterview: null`; zero-applied case returns null rates and empty groups.
- Parity: both cross-runtime parity tests now cover the new column.

## Verification

- `uv --project workers/automation run --extra dev pytest -q workers/automation/tests/test_dashboard_projection.py workers/automation/tests/test_projection_builder.py workers/automation/tests/test_audit_projection_parity.py` → **28 passed**
- `pnpm api:test` → **288 passed (17 files)**
- `pnpm api:check` → clean
- `pnpm web:check` → clean (shared contract + web fixture)
- full `pytest -q` → **1493 passed**
- `ruff check` on changed Python → clean

## Docs

- `docs/architecture.md`: Read-Model Projections table + conversion-projection note.
- `docs/local-reliability-qa.md`: regression matrix entry.


packages/domain-types/test/fixtures/audit_projection_parity.json
```diff
@@ -716,6 +717,36 @@
           "total": 1
         }
       ],
+      "outcome_conversion_json": {
```
ebarti, line 807: PR Reviewer - pr-reviewer

Severity: Low
Gate: Pass
Evidence: The shared parity fixture's dashboard seed produces an all-zero funnel for this column (totals applied:1, reply/interview/offer/rejection:0; one source, one band, no application_outcomes rows). So the cross-runtime parity guard (column-set + value deep-equal, both runtimes) proves the two builders agree on the STRUCTURE and the trivial case, but never exercises the rich funnel arithmetic cross-runtime: non-zero reply/interview/offer/rejection counts, the cumulative outcome-kind sets, the >1 source case, or the bySource sort tie-break (`-applied, source`). Those are only checked independently per runtime (test_dashboard_projection.py seeds 5 jobs; projections.test.ts seeds 3) with different seeds, so a future divergence in the rich path would not be caught by the parity fixture.
Reproduction: Reasoned from the fixture (all funnel stages 0) + the two parity tests' comparison (normalizeRow/_normalize_row parse-then-deep-equal against this same fixture).

This is Low, not blocking: I diffed both builders line-by-line and they are structurally identical (same applied predicate, same band thresholds, same cumulative kind sets, same sort), and the funnel consumes `active`/`active_rows` fields (`source`, canonical-first `fit_score`) that the existing by_source/score_distribution columns already hold at parity. So real divergence risk is low today.

Required fix (optional, strengthens the non-negotiable parity guard): add to the canonical seed at least one `application_outcomes` row (e.g., an interview) on the seeded applied job and, ideally, a second applied job on a different source/band, so the parity fixture's outcome_conversion_json carries non-zero reply/interview counts and a multi-entry bySource/byBand — making the rich funnel + sort genuinely cross-runtime-verified rather than only the all-zero case.

ebarti: Done in 5a13e04c. Enriched the SHARED fixture so outcome_conversion_json now carries a non-zero cross-runtime funnel: 3 applied+scored jobs across 2 sources (greenhouse x2, linkedin x1 — unequal counts so there is no cross-runtime tie-order ambiguity in by_source_json) and 2 score bands (strong, moderate), with interview/offer/rejection outcomes. This exercises a multi-entry bySource (applied-desc sort) + byBand and the cumulative kind sets (offer implies interview + reply; rejection implies reply):

- totals: applied 3, reply 3, interview 2, offer 1, rejection 1
- bySource: greenhouse {applied 2, reply 2, interview 1, offer 1, rejection 1}, linkedin {applied 1, reply 1, interview 1}
- byBand: strong {applied 2, reply 2, interview 1, rejection 1}, moderate {applied 1, reply 1, interview 1, offer 1}

Both parity tests pass against the single shared fixture: test_audit_projection_parity.py (2 passed) and audit-projection-parity.test.ts (2 passed); full suites green (pytest 28, api:test 288/17 files).

Verified the guard now catches a builder-specific divergence in the non-trivial funnel: dropping `offer` from the TS interview-implies set fails audit-projection-parity.test.ts against the shared fixture (greenhouse/totals interview 1 -> 0), then reverted. The old all-zero seed could not have caught this.

PR #227, opened 2026-07-02
fix(projections): deterministic dashboard by_source tie-break

ebarti:
## What

Dashboard `by_source` ordering is now count-descending **then source-name ascending** in BOTH projection builders:

- `apps/api/src/projections.ts` — `.sort((a, b) => b[1] - a[1] || (a[0] < b[0] ? -1 : a[0] > b[0] ? 1 : 0))`
- `workers/automation/src/jobhunter/infrastructure/projections/projection_builder.py` — `sorted(source_counts.items(), key=lambda kv: (-kv[1], kv[0]))`

This mirrors the tie-break the outcome-conversion `bySource` rollup (#223) already uses in both runtimes.

## Why

On a count tie, both builders fell back to insertion order — the TS `Map` seeded by rowid scan order, the Python dict by job-id scan order — so `dashboard_projections.by_source_json` could diverge between runtimes for identical data (last-writer-wins flips the displayed order nondeterministically). Found while re-deriving the #223 parity fixture: an incidental tie broke parity differently per runtime. The fixture was de-tied there; this PR fixes the underlying nondeterminism.

## Validation

- New per-runtime tie regressions, each seeding a tied source pair reverse-alphabetically and asserting A→Z ordering: `apps/api/test/projections.test.ts` ("orders dashboard by_source by count desc then source name ascending") and `workers/automation/tests/test_dashboard_projection.py::test_by_source_orders_ties_by_source_name`. Unit tests rather than the shared parity fixture: a tie-ordering-only change would otherwise force re-deriving every unrelated column in `audit_projection_parity.json`.
- `pnpm api:check` clean; `pnpm api:test` 300 passed; `uv --project workers/automation run --extra dev pytest -q` 1515 passed; ruff clean.
- Verified inside the full 23-branch integration union: pytest 1653 / api:test 323, all green.

No doc update: determinism bug fix with no user-facing or public-behavior change (`by_source_json` consumers see identical content, now in a stable order).

Stacked on #223 (`feat/outcome-conversion-projection`); merge after it.

### DIFF
--- a/workers/automation/tests/test_dashboard_projection.py
+++ b/workers/automation/tests/test_dashboard_projection.py
@@ -337,6 +337,30 @@ def test_by_source_counts(conn: sqlite3.Connection) -> None:
     assert counts["TwoCo"] == 1
 
 
+def test_by_source_orders_ties_by_source_name(conn: sqlite3.Connection) -> None:
+    # Netflix leads on count; Acme and Wayfair tie at 2, seeded reverse-
+    # alphabetically so a count-only sort would leak insertion order. The
+    # tiebreak re-orders the tie A->Z, byte-identical to the TS builder.
+    seeded = [
+        ("https://example.com/w1", "Wayfair"),
+        ("https://example.com/w2", "Wayfair"),
+        ("https://example.com/n1", "Netflix"),
+        ("https://example.com/n2", "Netflix"),
+        ("https://example.com/n3", "Netflix"),
+        ("https://example.com/a1", "Acme"),
+        ("https://example.com/a2", "Acme"),
+    ]
+    for url, site in seeded:
+        _seed_job(conn, url, site=site)
+        record_job_event(conn, url, "discover", "JobDiscovered")
+    conn.commit()
+    ProjectionBuilder(conn_factory=lambda: conn).refresh()
+
+    row = _dashboard(conn)
+    by_source = json.loads(_row_value(row, "by_source_json", "[]"))
+    assert by_source == [["Netflix", 3], ["Acme", 2], ["Wayfair", 2]]
+
+
 def _apply_job(conn: sqlite3.Connection, url: str, *, site: str, fit_score: int) -> None:
     _seed_job(conn, url, site=site)
     conn.execute(
--- a/workers/automation/tests/test_dashboard_projection.py
+++ b/workers/automation/tests/test_dashboard_projection.py
@@ -338,17 +338,21 @@ def test_by_source_counts(conn: sqlite3.Connection) -> None:
 
 
 def test_by_source_orders_ties_by_source_name(conn: sqlite3.Connection) -> None:
-    # Netflix leads on count; Acme and Wayfair tie at 2, seeded reverse-
-    # alphabetically so a count-only sort would leak insertion order. The
-    # tiebreak re-orders the tie A->Z, byte-identical to the TS builder.
+    # Netflix leads on count; Acme and Wayfair tie at 2. The builder seeds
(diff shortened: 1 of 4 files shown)


workers/automation/tests/test_dashboard_projection.py
```diff
@@ -337,6 +337,30 @@ def test_by_source_counts(conn: sqlite3.Connection) -> None:
...
+    # alphabetically so a count-only sort would leak insertion order. The
+    # tiebreak re-orders the tie A->Z, byte-identical to the TS builder.
+    seeded = [
+        ("https://example.com/w1", "Wayfair"),
+        ("https://example.com/w2", "Wayfair"),
+        ("https://example.com/n1", "Netflix"),
+        ("https://example.com/n2", "Netflix"),
+        ("https://example.com/n3", "Netflix"),
+        ("https://example.com/a1", "Acme"),
+        ("https://example.com/a2", "Acme"),
```
ebarti: PR Reviewer - pr-reviewer

Severity: Medium
Gate: Pass
Evidence: This regression does NOT fail if the production tiebreak is removed, so it does not guard the branch it claims to. `SqliteProjectionStore.fetch_job_list` returns rows `ORDER BY job_id` (sqlite_projection_store.py:802), so `active_rows` — and thus the `source_counts` dict insertion order — is keyed by URL, not by this seed-list order. The seed URLs put Acme (`a1`,`a2`) before Wayfair (`w1`,`w2`), i.e. already A→Z for the tied pair. Under the pre-PR count-only sort `sorted(items, key=lambda kv: kv[1], reverse=True)` (stable, reverse preserves insertion order for ties) the result is exactly `[["Netflix",3],["Acme",2],["Wayfair",2]]` — identical to the assertion on L361. The comment on L341-343 ("seeded reverse-alphabetically so a count-only sort would leak insertion order") is false for this runtime because `ORDER BY job_id` re-sorts by URL before the counter runs.
Reproduction: Reproduced with a Python simulation of the exact insertion order + both sort keys — OLD (count-only) == NEW (tiebreak) == asserted. Contrast the TS twin: it reads `job_list_projections` with no ORDER BY (projections.ts:2503-2510), so its Map is seeded in rowid/insert order (Wayfair before Acme) and its OLD sort yields `[Netflix,Wayfair,Acme,ExampleCo]` ≠ assertion — the TS test genuinely discriminates; only this Python one does not.

A future reversion of the Python tiebreak (projection_builder.py:1551 back to count-only) would ship undetected. The #223 cross-runtime parity fixture was deliberately de-tied, so nothing else covers a Python tie — and Python is precisely the runtime whose URL-ordered scan this determinism fix is meant to tame.

Required fix: decouple the seed URLs from source-name order so `ORDER BY job_id` seeds the tied pair reverse-alphabetically. E.g. Wayfair `https://example.com/a-wayfair-1`/`-2`, Netflix `.../m-netflix-1..3`, Acme `.../z-acme-1`/`-2`. Verified: OLD then yields `[Netflix,Wayfair,Acme]` (fails the assertion) while NEW yields `[Netflix,Acme,Wayfair]` (passes). Update the L341-343 comment to state the ordering is by job_id/URL, not seed order.

ebarti: Fixed in 295c8d1a: seed URLs now put Wayfair first in job_id order (a-wayfair < m-netflix < z-acme) per your verified recipe, and the comment now describes the ORDER BY job_id mechanism. Teeth proven exactly as you prescribed: reverting the production key to count-only fails the test; restored, 14/14 pass. Full suite 1515 + ruff green.

PR #228, opened 2026-07-02
fix(materials): ground revision-gate coverage in shipped resume text

ebarti:
## What changed

The post-generation revision gate counted a requirement as covered whenever the model's self-declared `GeneratedClaimMapping`s asserted it — the shipped text was never consulted. The requirement-led audit chips separately bound requirements to bullets by verbatim keyword matching. The same accepted artifact could therefore persist **"Must-have coverage: 100% · passed"** beside a provenance audit showing **"4/9 requirements covered"** with must-haves flagged missing from the tailored resume (observed live on the Apply Review surface for the Digital Hub Director job, generation 32).

This PR makes **claim grounding** the single coverage criterion, owned by a new pure domain module `claim_grounding.py`:

- A coverage-bearing claim counts only when bound to a **shipped rendered line**: location resolution (`experience.{id}.bullets[i]` → `experience:{id}#{i}`, all payload-surface aliases) plus token-sequence text binding; a text-scan fallback covers location drift.
- A **voice-reworded** bullet honors its pre-voice text (voice keeps bullet identity 1:1 and re-runs the fabrication gate), so meaning-preserving rewording never unships a claim.
- The assembler's **policy-gated summary swap** (`allow_summary_rewrite: no`) makes a proposed-but-unshipped rewrite honestly ungrounded.
- `score_generated_resume_against_target` now requires the grounding; requirements asserted only by unshipped claims are reported as `claimed_only_requirement_ids`, count as **uncovered**, and their prioritized fixes say the claim text never shipped — feeding the existing revision loop so it repairs real gaps (self-healing) instead of echoing self-assessment.
- The per-candidate gate renders the candidate's shipped lines (pure, assembler-mirroring `build_bullet_provenance`) and records an inspectable `grounding` audit beside `fit_score`/`revision_decision`.
- Accepted artifacts persist provenance rows **enriched with grounded claim requirement/evidence links**, so the audit chips and the "N/M requirements covered" badge derive from the same criterion as the gate — the three surfaces can no longer disagree for new generations.
- The shipped (post-voice) artifact persists a lifecycle-labeled `post_generation_fit_final` record (`lifecycle: post_voice_shipped`, gate thresholds, explicit residual warnings when voice degraded grounded coverage after acceptance).

## Why

[...]

PR #229, opened 2026-07-02
fix(apply-review): label coverage basis and revision semantics from grounded audit

ebarti:
## What

Makes the Apply Review "Requirement-led tailoring audit" read/render surface honest about **which basis** each coverage number came from, and fixes the revision-attempt semantics. Stacks on #228.

- **API (`apps/api/src/application-feedback.ts`)**
  - `parseAuditRevision` now exposes `coverageBasis` (`grounded_shipped_text_v1` when the persisted `fit_score.coverage_basis` matches, else `judge_claimed_legacy`), `claimedOnlyRequirementIds` (default `[]`), and `revisionsUsed` (`max(0, attempt - 1)`; `null` when the attempt ordinal is absent).
  - New `parseAuditShippedFit` parses the lifecycle-labeled `post_generation_fit_final` into an optional `shippedFit` (`{lifecycle, score, mustHaveCoverage, claimedOnlyRequirementIds, passed, warnings, coverageBasis}`), null-safe for legacy artifacts.
- **Contracts (`packages/contracts/src/schemas.ts`)** — extend `ApplyReviewRequirementLedAuditRevision`, add `ApplyReviewRequirementLedShippedFit` + `ApplyReviewCoverageBasis`, and an optional `shippedFit` on the audit. All optional/nullable so legacy items validate.
- **Web (`ApplyReviewView` → `RevisionAudit`)** — when `shippedFit` is present, show its grounded must-have coverage as the primary figure tagged **"grounded (shipped text)"**, render its `warnings` as **post-acceptance audit findings**, and keep the gate line as history (relabeled "Gate-recorded coverage"). Always tag the gate basis **"grounded"** vs **"judge-claimed (legacy)"** as a visible tag. Replace "Attempt A/B" with **"Revisions used: N of M"**. Chip-state logic untouched.

## Why

For legacy artifacts, the audit showed a self-contradiction: a judge-claimed "Must-have coverage 100% · passed · Attempt 2/1" gate record next to the provenance-grounded "4/9 requirements covered" badge. #228 grounded the worker's gate so **new** artifacts agree. This PR makes the **read model** label the basis so legacy judge-claimed numbers are shown-but-labeled (never hidden), and the shipped grounded fit (`post_generation_fit_final`) drives the primary coverage figure. Per repo audit discipline: don't remove the embarrassing value — label its source.

## How validated

All commands run from the worktree, verbatim results:

- `pnpm api:check` — pass (exit 0)
- `pnpm api:test` — 19 files, **326 passed**
- `pnpm web:check` — pass (exit 0)
- `pnpm --filter @jobhunter/web test` — 146 files, **864 passed**
- `pnpm web:build` — built in 5.09s (pre-existing chunk-size warning only)

New tests:
- `apps/api/test/application-feedback.test.ts`: grounded metadata → `coverageBasis`/`claimedOnlyRequirementIds`/`revisionsUsed`/`shippedFit`; legacy metadata → `judge_claimed_legacy` + `shippedFit` null; revisions-used math incl. `attempt=1 → 0` and absent attempt → `null`.
- `apps/web/src/views/apply-review/ApplyReviewView.test.tsx`: (a) grounded item renders the shipped grounded coverage figure + "grounded (shipped text)" tag and "Revisions used: 1 of 1"; (b) legacy item renders the "judge-claimed (legacy)" tag with the old 50% still visible; (c) **Digital-Hub regression** — a grounded fixture (`coveredRequirements` length 4, `requirementCount` 9, `shippedFit.mustHaveCoverage` 0.5) renders no `100%` and no bare "passed" in the audit region, badge shows "4/9", and chips show both covered + missing (badge/chips/gate agree).

## Parity decision

`audit-projection-parity.test.ts` / `test_audit_projection_parity.py` guard **dual-written projection columns** (employer analysis, bullet provenance, coverage/voice JSON). The requirement-led audit + revision-gate records are read by the API directly from the raw `job_materials_artifacts` metadata (not a projection column), so they are outside the parity suite's scope — no extension needed. The worker side that produces the metadata is covered by `test_claim_grounding.py` / `test_requirement_led_tailoring.py` (#228); the TS read side is covered by `application-feedback.test.ts`.

## Notes / deviations

- The contracts objects are TypeScript `interface`s (not Zod schemas); the task said "Zod schemas" but the apply-review response types are plain interfaces, so those were extended instead.
- The strict "no `100%`" regression assertion is scoped to the "Requirement-led tailoring audit" region (the exact contradiction surface) plus a document-wide check for the specific false claim "Must-have coverage: 100%".


apps/web/src/views/apply-review/ApplyReviewView.test.tsx
```diff
@@ -932,6 +932,166 @@ describe("<ApplyReviewView>", () => {
...
+    expect(auditRegion.getByText("judge-claimed (legacy)")).toBeInTheDocument();
+    expect(auditRegion.getByText(/Fit gate:\s*7\.6\/10/)).toBeInTheDocument();
+    expect(auditRegion.getByText(/Must-have coverage:\s*50%/)).toBeInTheDocument();
+    expect(auditRegion.getByText("Revisions used: 0 of 1")).toBeInTheDocument();
+    expect(auditRegion.queryByText("grounded (shipped text)")).not.toBeInTheDocument();
+    expect(auditRegion.queryByText("grounded")).not.toBeInTheDocument();
+    expect(auditRegion.queryByText(/Attempt/)).not.toBeInTheDocument();
+  });
+
+  it("does not contradict the coverage badge with a judge-claimed 100% pass (Digital Hub regression)", async () => {
```
ebarti, line 1066: PR Reviewer - pr-reviewer

Severity: Medium
Gate: Pass
Evidence: apps/web/src/views/apply-review/ApplyReviewView.test.tsx:1015-1090. This "Digital Hub regression" fixture sets revision.mustHaveCoverage=0.5 and shippedFit.mustHaveCoverage=0.5, so the assertions queryByText(/100%/) (L1087), queryByText("passed") (L1088), and the document-wide "Must-have coverage: 100%" check (L1090) never had a 100%/passed value to find — they pass trivially, including on the pre-fix code (which would render "50%" from the same data). The legacy-labeling test (~L999) also uses 50% data. The reported defect (and PR title) is a judge-claimed "Must-have coverage 100% · passed · Attempt 2/1" gate record contradicting a "4/9 requirements covered" badge; no test ever constructs that record.
Reproduction: Reasoned from code (fixture values vs assertions). Full suite verified green: pnpm --filter @jobhunter/web test (146 files, 864 passed).

The test still discriminates old vs new via the labeling assertions (no "Attempt", presence of "grounded (shipped text)"), so it is valid coverage of the new labeling. But per CLAUDE.md audit discipline ("add or update a regression fixture that proves the exact invariant the human complained about ... reproduces the bad state from canonical data"), the exact 100%/passed legacy contradiction is not regression-locked, and the no-100%/no-passed assertions give false confidence.

Required fix: Add/extend a fixture whose legacy revision carries mustHaveCoverage: 1.0 (coverageBasis judge_claimed_legacy, a passed-style reason) next to a 4/9 badge, and assert the 100% renders WITH the "judge-claimed (legacy)" label and is not presented as grounded/authoritative — locking the exact screenshot invariant.

apps/web/src/views/apply-review/ApplyReviewView.tsx
```diff
@@ -696,31 +700,59 @@ function BulletOverflowAudit({
...
     <section className="apply-review-audit-section">
       <h4>Revision gate</h4>
+      {shippedFit ? (
+        <div className="apply-review-audit-shipped-fit" aria-label="Shipped grounded fit">
+          {shippedFit.mustHaveCoverage !== null ? (
+            <span className="tag muted">
+              Must-have coverage: {Math.round(shippedFit.mustHaveCoverage * 100)}%
+            </span>
+          ) : null}
+          <span className="tag ok">grounded (shipped text)</span>
```
ebarti: PR Reviewer - pr-reviewer

Severity: Low
Gate: Pass
Evidence: apps/web/src/views/apply-review/ApplyReviewView.tsx:724. The "grounded (shipped text)" tag is rendered unconditionally whenever shippedFit is present, ignoring the parsed shippedFit.coverageBasis. The sibling revision-line basis tag (L744-745) IS gated on revision.coverageBasis === "grounded_shipped_text_v1". The parser (apps/api/src/application-feedback.ts parseAuditShippedFit, L1854-1875) computes shippedFit.coverageBasis, but no consumer reads it, so the field is currently dead and this tag is presence-gated rather than basis-gated.
Reproduction: Reasoned from code. Not a current-data bug: the only producer, _final_fit_record (workers/automation/src/jobhunter/domain/materials/use_cases.py:2844), always sets coverage_basis = grounding.basis = "grounded_shipped_text_v1", so shippedFit.coverageBasis is always grounded when the record exists.

Because the tag is not tied to the value it asserts, a future/alternate producer that writes post_generation_fit_final with a different basis would surface a false "grounded (shipped text)" label — exactly the value-laundering the revision line already guards against, and which this PR set out to prevent.

Required fix: Gate this tag on shippedFit.coverageBasis === "grounded_shipped_text_v1" (label otherwise), mirroring the revision line, so the parsed field is load-bearing and the tag can never misrepresent the basis.

apps/web/src/views/apply-review/ApplyReviewView.test.tsx
ebarti: Fixed in 448593a1.

Added a dedicated legacy regression test (`labels a legacy judge-claimed 100% pass instead of hiding it, without contradicting the 4/9 badge (Digital Hub regression)`) that reproduces the original screenshot state: revision `score: 10`, `mustHaveCoverage: 1.0`, `attempt: 2`, `maxRevisionAttempts: 1`, `revisionsUsed: 1`, `reason: "passed"`, `coverageBasis: "judge_claimed_legacy"`, `shippedFit: null`, alongside `coveredRequirements` length 4 and `requirementCount: 9` (the 4/9 badge).

It asserts the 100% figure IS rendered but only inside the `.apply-review-audit-revision` block next to the `judge-claimed (legacy)` label, is not tagged grounded, renders `Revisions used: 1 of 1`, and coexists with the `4/9 requirements covered` badge. Verified it FAILS when the legacy label is gated off (`getByText("judge-claimed (legacy)")` throws), then reverted the experiment (not committed). The prior 0.5-coverage test is kept as the grounded-path coherence check.

apps/web/src/views/apply-review/ApplyReviewView.tsx
ebarti: Fixed in 448593a1.

The shipped-fit tag now gates on `shippedFit.coverageBasis === "grounded_shipped_text_v1"` (mirrors the revision-line gating at the lines just below it); a non-grounded basis renders the source label `judge-claimed (legacy)` instead of `grounded (shipped text)`.

Added a test (`does not tag a non-grounded shipped fit as grounded (shipped text)`) that renders a `shippedFit` with `coverageBasis: "judge_claimed_legacy"` and asserts the grounded tag is absent from the `.apply-review-audit-shipped-fit` block while the source label is present.

PR #230, opened 2026-07-03
docs(plans): temporal-native rearchitecture plan (P0-P5)

ebarti:
## What

Adds `docs/plans/2026-07-03-temporal-native-rearchitecture.md`: the phased plan (P0-P5) to make Temporal the single execution authority — per-job durable execution, raised/classified errors, workflow outcomes projected into the read model, engine-owned recovery replacing all three trigger-coupled reapers, an at-most-once apply path with a user-configurable approval gate (default ON), browser-layer dry-run enforcement, Temporal Schedules (off by default), and a single CLI-to-workflow execution path.

## Why

The 2026-07-02 six-agent resilience audit (all findings file:line-verified) found Temporal wired correctly but architecturally sidelined:

- Workflow results are never consumed (no `handle.result()`/`describe()` anywhere) — Temporal-layer failures are invisible in-app.
- Stage runners swallow all exceptions into `status="error"` normal returns, so activity retry policies almost never fire.
- Recovery is trigger-coupled reapers that mark work failed for a manual retry click and never re-drive it.
- Apply violates its documented at-most-once contract (unconditional running→pending rewind + post-submit-only durability → double-submission risk); dry-run is prompt-text-only under `bypassPermissions`.
- No overlap control, progress-blind heartbeats, zombie threads after timeout, JSON-RPC with no timeout at any layer and serial dispatch.

Fixing these individually would preserve the structural defect; this plan inverts the architecture at the root (user-authorized rip-and-replace).

Design decisions D1-D4 recorded in the doc were confirmed by the user on 2026-07-02: Python-native Temporal ownership; configurable apply approval gate defaulting to required; Schedules capability shipped disabled; single execution path.

## Validation

Docs-only change — no code paths affected. Content is grounded in the file:line-verified audit plus three read-only code explorations (per-job runner readiness, apply internals, contract/ADR surface); every work item cites its exact seam. Implementation PRs will stack per the phase order (P0 → P1 → {P2 ∥ P3} → P4 → P5), each with its own review + QA gates.

PR #231, opened 2026-07-03
fix(worker): bound LLM retries and cap score attempts (P1a)

ebarti:
## What

Two file-disjoint hardening slices called out in the Temporal-native rearchitecture plan (#230, P1 "Parallelization note" → P1a). Both stop a single stuck/hostile job from imposing unbounded LLM cost or latency. Python-only; runs in parallel with P0 and does not touch any file P0/P1b owns.

### 1. LLM retry hardening — `workers/automation/src/jobhunter/llm.py`
- **Broader transient retry.** Retry `429`, **all `5xx`**, and connection-level errors (`httpx.TransportError`: connect/read timeouts, network resets, protocol errors) with bounded exponential backoff. Previously only `429`/`503` and bare `TimeoutException` were retried, so a `500`/`502`/`504` or a connection reset failed the activity immediately.
- **Capped, jittered backoff.** A server `Retry-After` is now honored **up to a 60s ceiling** and every backoff sleep gets jitter. Previously `Retry-After` was slept verbatim (`float(retry_after)`), so a hostile or buggy value (e.g. `86400`) could park an activity for hours.
- **Fail-fast preserved.** Non-retryable `4xx` (other than `429`) still raise immediately. The public surface and return types of `chat()` / `chat_json()` are unchanged — retry *policy* only. No migration to the agent-SDK path.

### 2. Score attempt cap — `workers/automation/src/jobhunter/database.py` + `scoring/scorer.py`
- **Selector cap.** `pending_score` now excludes jobs whose score stage has failed 5+ times, mirroring the tailor/cover `< 5` selector cap (`_EFFECTIVE_SCORE_ATTEMPTS < 5`, read from `job_stage_states` via the existing `_SCORE_DOWNSTREAM_STATE_JOIN`, extended to expose `attempt_count`). Previously `pending_score` had **no cap**, so a permanently-failing job was re-selected and re-billed on every batch forever.
- **Increment on failure.** Score failures now advance `job_stage_states.attempt_count` (batch `run_scoring` and single `score_job_by_url`). The count is threaded through the `running`-state write because `set_stage_state` resets `attempt_count` to `0` when it is omitted — a bare `running` write would otherwise wipe the counter between batch re-selections (confirmed empirically). A succeeded job is excluded by its non-null fit score, so the success path is untouched.

## Why
Both were flagged in the plan as unbounded re-billing / unbounded sleep risks (`llm.py:392-406` uncapped `Retry-After`; `database.py:3505` uncapped `pending_score`).

## Validation
- `uv --project workers/automation run --extra dev pytest -q` → **1676 passed**
- `uv --project workers/automation run --extra dev ruff check .` → **All checks passed!**

New tests: `test_llm_client.py` (5xx retried→succeeds; connection error retried→succeeds; hostile `Retry-After: 86400` sleeps ≤ cap; `400` fails fast without retry; persistent transient bounded to `_MAX_RETRIES`). `test_score_queue_selectors.py` (`pending_score` excludes at cap, includes under cap). `test_scorer.py` (failure increments attempts across batches until the cap engages and the LLM is no longer called; single-job path increments too).

## Docs
Internal hardening with no change to public/CLI/API behavior → no doc update warranted per CLAUDE.md.

## Known follow-up (out of scope — `runner.py` is P0/P1b territory)
The streaming runner's `_count_pending("score")` uses a separate query (`_PENDING_SQL["score"]` in `pipeline/runner.py`) that does **not** include the new attempt cap, so it can over-count a capped-out job. Impact is **bounded and non-billing**: `run_scoring` selects via the (now capped) `get_jobs_by_stage`, returns empty, and the loop's no-progress guard bails after 3 no-op passes (marking the stage "stuck" rather than "done") — no extra LLM calls. The clean parity fix is to add `_SCORE_DOWNSTREAM_STATE_JOIN` + `_EFFECTIVE_SCORE_ATTEMPTS < 5` to `_PENDING_SQL["score"]`, which lives in the P1b-owned `runner.py`; leaving it for that slice to avoid a conflicting edit.

### DIFF
--- a/workers/automation/src/jobhunter/llm.py
+++ b/workers/automation/src/jobhunter/llm.py
@@ -11,6 +11,7 @@ LLM_MODEL env var overrides the model name for any provider.
 
 import json
 import logging
+import math
 import os
 import random
 import re
@@ -136,18 +137,26 @@ def _retry_wait(attempt: int, retry_after: str | None = None) -> float:
     """Seconds to sleep before the next transient-failure retry.
 
     Uses exponential backoff from ``_RATE_LIMIT_BASE_WAIT``, honoring a
-    server ``Retry-After`` (seconds) when present. The result is jittered to
-    break retry lockstep and hard-capped at ``_MAX_RETRY_WAIT`` so a hostile
-    or buggy ``Retry-After`` cannot sleep for hours inside an activity.
+    server ``Retry-After`` (seconds) when present and finite. The result is
+    jittered to break retry lockstep and clamped to ``[0, _MAX_RETRY_WAIT]``
+    before it is returned, so a hostile or buggy ``Retry-After`` (huge,
+    negative, NaN, or infinite) can neither park an activity for hours nor
+    reach ``time.sleep()`` with a value it rejects.
     """
     wait = _RATE_LIMIT_BASE_WAIT * (2 ** attempt)
     if retry_after is not None:
         try:
-            wait = float(retry_after)
+            parsed = float(retry_after)
         except (ValueError, TypeError):
-            pass
+            parsed = None
+        # Only honor a finite Retry-After; NaN/inf are meaningless, so fall
+        # back to the exponential backoff above.
+        if parsed is not None and math.isfinite(parsed):
+            wait = parsed
     wait += random.uniform(0, _RETRY_JITTER)
-    return min(wait, _MAX_RETRY_WAIT)
+    # Floor at 0 (a negative Retry-After would make time.sleep raise
+    # ValueError) then cap at the ceiling.
+    return max(0.0, min(wait, _MAX_RETRY_WAIT))
 
 
 _GEMINI_COMPAT_BASE = "https://generativelanguage.googleapis.com/v1beta/openai"
--- a/workers/automation/src/jobhunter/llm.py
+++ b/workers/automation/src/jobhunter/llm.py
@@ -12,6 +12,7 @@ LLM_MODEL env var overrides the model name for any provider.
 import json
 import logging
 import os
+import random
 import re
 import time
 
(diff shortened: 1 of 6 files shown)


workers/automation/src/jobhunter/llm.py
```diff
@@ -119,9 +120,34 @@ def create_client(provider: str | None = None, model: str | None = None) -> "LLM
...
+    or buggy ``Retry-After`` cannot sleep for hours inside an activity.
+    """
+    wait = _RATE_LIMIT_BASE_WAIT * (2 ** attempt)
+    if retry_after is not None:
+        try:
+            wait = float(retry_after)
+        except (ValueError, TypeError):
+            pass
+    wait += random.uniform(0, _RETRY_JITTER)
+    return min(wait, _MAX_RETRY_WAIT)
```
ebarti: PR Reviewer - pr-reviewer

Severity: High
Gate: Fail
Evidence: `_retry_wait` (llm.py:135-150) parses `Retry-After` with `float(retry_after)` and returns `min(wait, _MAX_RETRY_WAIT)` with no lower bound. A negative or non-finite header value flows straight through: `float("-100")` -> `-100`, `+ jitter` -> ~`-96`, `min(-96, 60)` -> `-96`; `float("nan")` -> `nan`, `min(nan, 60)` -> `nan`. The retry loop then calls `time.sleep(wait)` (llm.py ~430), which raises `ValueError: sleep length must be non-negative` (negative) / `ValueError: Invalid value NaN (not a number)` (nan).
Reproduction: Reproduced against real llm.py. `chat()` with a MockTransport returning HTTP 429 + header `Retry-After: -100` raised `ValueError: sleep length must be non-negative` after 1 call (the warning log even printed "Waiting -99.3s before retry 1/5"). `Retry-After: nan` -> `_retry_wait` returned `nan`. `_retry_wait(0, "inf")` correctly caps to 60; HTTP-date and other non-numeric strings correctly fall back to backoff.

A hostile/buggy server sending a negative or `nan`/`-inf` `Retry-After` (or `X-RateLimit-Reset-Requests`) on a 429/5xx crashes the retry with an unhandled `ValueError` instead of performing the bounded, jittered retry. This violates the phase-spec acceptance bar ("Retry-After parsing: hostile/malformed values (non-numeric, HTTP-date form, negative) must not crash or bypass the cap" and "jitter can't produce negative/zero sleeps") and the PR's own stated goal of hardening against a "hostile or buggy Retry-After". Blast radius is bounded in the batch scorer (the `ThreadPoolExecutor` result handler at scorer.py:440 catches it and records a stage failure, so the counter still advances and the cap engages), but it defeats the in-client retry on exactly the malformed input this PR targets, and `test_chat_caps_hostile_retry_after_header` only exercises the large-positive case (86400), leaving the crashing negative/NaN path uncovered.

Required fix: Clamp the returned wait to a non-negative, finite value bounded by the ceiling, e.g. reject non-finite parses and floor at 0 (`if not math.isfinite(wait): wait = _RATE_LIMIT_BASE_WAIT * (2 ** attempt)` then `return min(max(wait, 0.0), _MAX_RETRY_WAIT)`), or clamp the parsed `retry_after` to `>= 0` and drop NaN/inf. Extend `test_chat_caps_hostile_retry_after_header` (or add a case) covering `Retry-After: -100` and `Retry-After: nan` asserting a bounded non-negative sleep and a successful retry rather than a raised `ValueError`.

ebarti: Fixed in e040a336.

`_retry_wait` now (1) drops non-finite parsed `Retry-After` values — NaN/inf are treated as an absent/invalid header and fall back to the exponential-backoff path via `math.isfinite()` — and (2) clamps the single return point with `max(0.0, min(wait, _MAX_RETRY_WAIT))`, so BOTH the Retry-After branch and the backoff+jitter branch are covered. The function contract is now: always returns a finite value in [0, _MAX_RETRY_WAIT], so `time.sleep()` can never receive a negative or NaN value.

Regression tests added mirroring the hostile-header test: `test_chat_retries_on_negative_retry_after` (`Retry-After: -100`) and `test_chat_retries_on_nan_retry_after` (`Retry-After: nan`) — each asserts `time.sleep` is called once with a finite value in [0, 60] and the request retries and succeeds on the next attempt.

Verification: `uv --project workers/automation run --extra dev pytest -q` -> 1678 passed; `uv --project workers/automation run --extra dev ruff check .` -> All checks passed!

ebarti: PR Reviewer - pr-reviewer

Severity: High -> Resolved
Gate: Pass
Verified against head e040a336. `_retry_wait` now drops non-finite parsed values via `math.isfinite()` (NaN/inf/-inf -> exponential-backoff fallback) before assignment to `wait`, and the single return point `max(0.0, min(wait, _MAX_RETRY_WAIT))` clamps both the Retry-After branch and the backoff+jitter branch to a finite [0, 60]. Confirmed:
- Unit: `_retry_wait` returns finite values in [0, 60] for `-100`, `nan`, `inf`, `-inf`, `0`, `86400`, HTTP-date, and None across attempts 0-4.
- Integration (previously crashed): 429 + `Retry-After: -100` now retries and succeeds (slept 0.0); `nan` and `-inf` retry and succeed via backoff (slept ~13s). No `ValueError`.
- `pytest workers/automation/tests/test_llm_client.py` -> 13 passed (incl. the two new negative/NaN cases asserting a finite sleep in [0, 60] and a successful retry). `ruff check` clean.
- No NaN leak path: the only `wait = parsed` assignment is guarded by `math.isfinite(parsed)` (short-circuited after `parsed is not None`), and jitter is always finite, so the clamp never receives NaN.

Resolving. The finding is fully addressed.

PR #233, opened 2026-07-03
feat(temporal): close the workflow loop — events, projections, deterministic IDs, reconciler (P0)

ebarti:
## What & why

Implements **P0 — Visibility contract & loop closure** of the Temporal-native rearchitecture (authority: plan PR #230, `docs/plans/2026-07-03-temporal-native-rearchitecture.md`). The 2026-07-02 resilience audit found Temporal wired but sidelined — workflow results unread, failures invisible, recovery via trigger-coupled reapers. This PR makes execution visible, consumes workflow outcomes, gives deterministic IDs real no-overlap, and closes the JSON-RPC/fetch hang.

## What changed

1. **`Workflow*` event family (6 types)** — `WorkflowStarted`, `WorkflowCompleted`, `WorkflowFailed`, `WorkflowCanceled`, `WorkflowTimedOut`, `WorkflowTerminated` — added in lockstep across the Python registry (`domain/events/workflow.py` + `__init__.py`), the TS registry (`packages/domain-types/src/events/workflow.ts` + union/const array), and the web invalidation router + fixtures (61 → 67 types). `every-event-has-handler` and `AssertEnumExhaustive` stay green.
2. **Finalize activities + loop closure** — new `infrastructure/temporal/finalize.py` (`record_workflow_started` / `record_workflow_outcome` + workflow-side emit helpers) wired into `JobPipelineWorkflow.run` and `ApplyWorkflow.run`: start marker at the top, terminal event on exit (Completed on success, Failed on stage/exception failure). Registered in `registry.py`. Workflow bodies stay deterministic (IO is in activities).
3. **Deterministic workflow IDs + overlap policies** — `WorkflowStartSpec` gains `id_conflict_policy` / `id_reuse_policy`; the default starter passes `USE_EXISTING` + `ALLOW_DUPLICATE` (no-op for `run-{uuid}`). `apply_action` derives a stable `apply-{jobKey}` id for single-job applies so a double-click attaches to the running run instead of double-submitting.
4. **`workflow_run_projections`** — Python-sole-writer table + `_rebuild_workflow_runs` fold under the shared `operations_projections` watermark; mirrored read-only in `apps/api/src/projections.ts`.
5. **Read model + API** — `listWorkflowRuns` repointed to `workflow_run_projections` (all workflow types; apply rows enriched via LEFT JOIN to `apply_run_projections`). New `GET /v1/workflow-runs/:runId` + `WorkflowRunDetail` contract, api-client method, web `useWorkflowRunDetailQuery` hook, and a workflow-aware `/runs` detail drawer. Non-apply runs render with their `workflowType` + correct 12-state badge.
6. **Describe-based reconciler** — `_reconcile_workflow_runs` in the worker heartbeat loop (15s): describes open rows and terminalizes CLOSED executions (matching terminal event) or NOT_FOUND (dev-server history loss → `WorkflowTerminated`), leaving RUNNING alone. This is the reaper replacement and the backstop for killed/timed-out/cancelled workers.
7. **JSON-RPC hang closure** — Python `JsonRpcServer.serve` dispatches on a bounded thread pool with a stdout lock (a slow handler no longer HOL-blocks `cancel`; responses correlate by `id`); the TS subprocess adapter adds a per-request timeout; the api-client wraps every fetch in an `AbortController` timeout.

### Deletions
- Dead RPC handlers `reset_stage` / `mark_applied` / `mark_skipped` / `cancel_stage` (no live TS dispatch) + registrations + contract param/result schemas + `RpcMethods` entries + tests.
- The 5 stale legacy result schemas for the workflow-mode rescore/tailor/retailor handlers (they now return the workflow-start shape) + their contract tests.

## Deviations from the spec
- **`analyze_job` kept.** The plan listed it as a dead handler, but `apps/api/src/local-actions.ts` still dispatches the `analyze_job` RPC. Per the plan's "keep a live caller and flag" rule, it stays.
- **TS overloaded `StageStarted`/`StageFailed` synthesis (`server.ts`) deferred.** `recordPipelineWorkflowStarted` has 6 live dispatch callers feeding *per-job stage* progress; run-level `Workflow*` events don't replace that per-job visibility, so removing it now would regress the jobs view. Tracked for a follow-up (fits P1's per-stage event work).
- **Cancellation status.** Cancelling the pipeline/apply workflows currently surfaces as a visible *failed* terminal — both workflows catch the cancellation `ActivityError` as a stage failure (pre-existing). Temporal also cancels newly-scheduled activities during workflow cancellation, so finalize can't record from the cancel path. True `WorkflowCanceled` status for these workflows is P1/CC6; the reconciler already maps genuinely-CANCELED executions to `WorkflowCanceled`.
- **Apply batch/continuous IDs** stay on `run-{uuid}` (TODO referencing P2) — only single-job apply gets `apply-{jobKey}`, per the plan's "don't touch P2 apply semantics" guidance.
- **Historical apply runs** that predate the `Workflow*` events don't appear in the unified runs list until they re-run (accepted cutover loss per `feedback_no_strangler.md`); the dashboard's recent-apply panel is unchanged.

## How validated
All commands run in the worktree; results below.

- `uv --project workers/automation run --extra dev pytest -q` → **1676 passed**
- `pnpm api:check` → clean · `pnpm api:test` → **323 passed**
- `pnpm web:check` → clean · `pnpm --filter @jobhunter/web test` → **898 passed** · `pnpm --filter @jobhunter/web test-d` → **10 passed, no type errors**
- `pnpm test` (api:test + web:build + python:test) → **green**

New tests: workflow-runs projection fold, finalize normal/failure paths, deterministic-id double-start (no duplicate execution), reconciler (CLOSED / NOT_FOUND / RUNNING / CANCELED), concurrent JSON-RPC out-of-order dispatch, and the subprocess adapter per-request timeout.

Refs plan PR #230.

## QA follow-up (post-review — commit `1a18c4cc`)
QA Gate: FAIL flagged one High + two Lows; addressed here.

- **High — /runs e2e regression.** The QA e2e seed only populated `apply_run_projections`, so after the read-model repoint to `workflow_run_projections` the /runs view rendered empty (three specs failed: `runs.spec.ts:3`, `route-visual-qa.spec.ts:330`, `:444`). `apps/api/test/qa-seed.ts` now creates `workflow_run_projections` and seeds a coherent terminal row for `qa-run-1` (ApplyWorkflow → succeeded) plus its `WorkflowStarted`/`WorkflowCompleted` `job_events`; the LEFT JOIN to `apply_run_projections` still hydrates the job title/employer/mode. Direct inserts simulate the Python sole-writer inside the test DB (that rule governs runtime code, not test seeds). `route-visual-qa.spec.ts` was updated to the `WorkflowRunDrawer` labels ("Workflow run details") that the P0 route repoint introduced — previously masked by the empty table.
- **Low — stale doc-comment** in `packages/contracts/src/schemas.ts` corrected to name `workflow_run_projections` as the unified source (`apply_run_projections` = apply-specific detail).

### Clarifications requested by QA
- **`run_stage` / discover overlap control is intentionally out of P0.** P0 gives every workflow a durable start marker + terminal event, and deterministic IDs for single-job apply; it does **not** add single-instance/overlap control for discover. Deterministic single-instance discover IDs (`discover-{tenant}`) arrive with **P4** per the plan's workflow taxonomy, at which point a second discover start coalesces onto the running instance.
- **Kill-worker detection latency.** When the worker dies mid-run the terminal row does **not** appear in 15–30s. Temporal must first notice the lost worker via `heartbeat_timeout` (~2 min for discover); the next reconciler tick (≤15s) then describes the now-CLOSED/NOT_FOUND execution and writes the terminal `Workflow*` event. The run *does* terminalize durably — latency is bounded by `heartbeat_timeout + one reconciler interval`, not by the 15s cadence alone.

### QA-fix validation (commit `1a18c4cc`, isolated ports 8795/5195)
- Target specs → **3 passed**: `runs.spec.ts:3`, `route-visual-qa.spec.ts:330`, `route-visual-qa.spec.ts:444`.
- Full e2e → **15 passed / 5 failed**; the 5 are the documented pre-existing failures on `main` (`dashboard.spec.ts:3`, `jobs-drawer.spec.ts:297`, `jobs-drawer.spec.ts:371`, `route-visual-qa.spec.ts:507`, `token-foundation.spec.ts:349`) — unrelated to /runs.
- `pnpm api:test` → **323 passed** · `pnpm --filter @jobhunter/web test` → **898 passed** · `pnpm api:check` / `pnpm web:check` → clean.
- Running the full e2e regenerates `docs/assets/screenshots/*.png` via `docs-screenshots.spec.ts`; those artifacts were restored (not committed).

## Reviewer follow-up (post-review — commit `3996ade1`)

Code review returned `Gate: PASS` (0 Blocker/High) with 1 Medium + 4 Lows; all folded in on the same branch.

### MEDIUM — reconciler could flip `failed` → `succeeded`
Both `JobPipelineWorkflow` / `ApplyWorkflow` encode stage/apply failure in their return value, so a failing run closes **COMPLETED** on the Temporal side even though finalize already recorded `WorkflowFailed`. The describe-reconciler mapped COMPLETED→succeeded and the last-terminal-wins fold let a late `WorkflowCompleted` overwrite the earlier `WorkflowFailed`. Fixed in **two layers**:
- **Layer 1** (`_record_reconciled_outcome`): `BEGIN IMMEDIATE`, re-read the row, skip if already terminal — the reconciler is a backstop for open rows, never an overwriter of terminal truth.
- **Layer 2** (`_project_workflow_run`): the fold is now **first-terminal-wins** — a later terminal `Workflow*` event stays in the timeline but does not replace the first terminal status.
- Regression tests (both fail pre-fix): reconciler leaves an already-terminal row untouched (no `WorkflowCompleted` emitted); fold keeps `failed` when `WorkflowFailed` precedes `WorkflowCompleted`.

### L2 — create-if-missing from the outcome event
Already provided by the fold: `record_workflow_outcome` → `_emit` → `refresh` → `_rebuild_workflow_runs` → `upsert_workflow_run` (`INSERT … ON CONFLICT DO UPDATE`), and `_project_workflow_run` builds a row from a terminal-only sequence. Added a regression test proving an outcome with **no** preceding start marker still terminalizes a row (passes on current code) — **no code change needed**. Note: the narrower "start marker fails *before the try* → no outcome emitted → no row" case is a structural interruptibility concern (the outcome never runs, so an outcome-UPSERT cannot help); flagged for **P1b**, not restructured here (the start-marker-before-try placement is deliberate for cancel-safety).

### L3 — `finalize.py` docstring
Corrected: cancellation is **not** recorded via `asyncio.shield` (no shield exists); the describe-reconciler owns cancel mapping.

### L4 — stale `cancel_stage` references
Removed in `handlers.py`, `cli.py` (the docstring listed `cancel_stage` + other now-TS handlers), `write-model.ts` (2 comments), and the live `docs/job-pipeline-architecture.md` line (post-hoc cancel is now the TS `cancelJobAction`). Left intentionally: historical `docs/delivered.md` + implemented-plan entries and the generated `.planning/codebase/ARCHITECTURE.md` snapshot (`cancelStage` in `CancelStageButton.tsx` / `server.ts` are live locals, unrelated to the deleted handler).

### L5 — orphaned `ApplyRunDrawer`
Deleted the component + story + a11y test after the `/runs` route swap to `WorkflowRunDrawer`. That orphaned its sole-dependency hook `useApplyRunQuery`, so it was removed too (hook + unit test + type-level test + operations barrel export) and the `applyRunsKeys` comment refreshed — **scope note:** slightly beyond the named 3 files, but required to satisfy "no dead components". Zero remaining `ApplyRunDrawer` / `useApplyRunQuery` imports.

### Reviewer-batch validation (commit `3996ade1`, isolated ports 8795/5195)
- Python → `ruff` clean · **1679 pytest passed** (incl. the 3 new regression tests).
- API → `pnpm api:check` clean · `pnpm api:test` **323 passed**.
- Web → `pnpm web:check` clean · `test-d` **9 passed** · unit **895 passed** (was 898; −3 from the removed `useApplyRunQuery.test.ts` ×2 + `ApplyRunDrawer.a11y.test.tsx` ×1).
- Full e2e → **15 passed / 5 failed**; the 5 are the documented pre-existing failures on `main` (`dashboard.spec.ts:3`, `jobs-drawer.spec.ts:297`, `jobs-drawer.spec.ts:371`, `route-visual-qa.spec.ts:507`, `token-foundation.spec.ts:349`). Target /runs specs re-run explicitly → **3 passed** (`runs.spec.ts:3`, `route-visual-qa.spec.ts:330`, `route-visual-qa.spec.ts:444`). Regenerated `docs/assets/screenshots/*.png` restored (not committed).

### LOW — reconciler-closed runs carry a reason (from the live QA demo)
Runs terminalized by the **describe-based reconciler** (`temporal workflow terminate`d executions, NOT_FOUND after dev-server loss, `temporal workflow cancel`, timeouts) landed with `errorCode` **and** `errorMessage` both null, so `/runs` showed a terminal run with no explanation (the heartbeat-timeout `failed` path already carried detail). Root cause: only `WorkflowFailed` carried `errorCode`; `WorkflowTerminated`/`WorkflowTimedOut` carried `errorMessage` only, `WorkflowCanceled` carried neither — and the reconciler passed no reason on the CLOSED path.

Fix (source-of-truth at the owning layer, not UI masking):
- Added `error_code` to `WorkflowTerminatedPayload` / `WorkflowTimedOutPayload` / `WorkflowCanceledPayload` (+ `error_message` to `WorkflowCanceledPayload`); `create_*` emit the camelCase keys. The fold already reads `errorCode`/`errorMessage` generically, so it picks them up with **no fold change**.
- `finalize.build_workflow_outcome_event` now propagates `error_code` (and `error_message` for canceled) on those branches.
- `_reconcile_one_workflow_run` / `_record_reconciled_outcome` stamp the reconciler's own provenance: `reconciled_not_found` (NOT_FOUND), `reconciled_terminated` (TERMINATED), `reconciled_closed_<status>` (canceled / timed_out / failed), plus a human-readable message quoting the Temporal execution status. A reconciled **success** keeps null reason fields (a success has no error).
- Mirrored the added fields in `@jobhunter/domain-types` `events/workflow.ts` + the web event fixture to keep the Python/TS `Workflow*` payloads 1:1.
- The read model already selects `error_code`/`error_message` for workflow runs (`read-model.ts`), so no read-model or UI change was required.
- Doc: one-line note in the `docs/architecture.md` reconciler bullet.
- Regression: extended the reconciler tests to assert the reason fields are non-null for reconciler-closed rows (failed, terminated/NOT_FOUND, canceled) — both extended assertions fail pre-fix.

### Follow-up 2 validation (commit `585cc538`, isolated ports 8795/5195)
- Python → `ruff` clean · **1679 pytest passed** (reconciler + fold + projection green).
- API → `pnpm api:check` clean · `pnpm api:test` **323 passed**.
- Web → `pnpm web:check` clean · `test-d` **9 passed** · unit **895 passed** · `pnpm web:build` OK.
- Full e2e → **15 passed / 5 failed**; the 5 are the same documented pre-existing failures (`dashboard.spec.ts:3`, `jobs-drawer.spec.ts:297`, `jobs-drawer.spec.ts:371`, `route-visual-qa.spec.ts:507`, `token-foundation.spec.ts:349`). Target /runs specs re-run explicitly → **3 passed** (`runs.spec.ts:3`, `route-visual-qa.spec.ts:330`, `:444`). Regenerated `docs/assets/screenshots/*.png` restored (not committed).


workers/automation/src/jobhunter/cli.py
```diff
@@ -1480,6 +1496,120 @@ def _worker_heartbeat_iteration(
...
+    _record_reconciled_outcome(
+        conn,
+        run,
+        status=status,
+        temporal_run_id=description.run_id or run.get("temporal_run_id"),
+    )
+    return True
+
+
+def _record_reconciled_outcome(
```
ebarti, line 1602: PR Reviewer - pr-reviewer

Severity: Medium
Gate: Pass
Evidence: `_reconcile_status_map()` (cli.py:1502-1512) maps `WorkflowExecutionStatus.COMPLETED -> "succeeded"`. `_reconcile_one_workflow_run` reads `open_workflow_runs()` as a snapshot, then `await ...describe()` per row; `_record_reconciled_outcome` (cli.py:1580) records unconditionally without re-checking the row is still open. `projection_builder.py::_project_workflow_run` uses last-terminal-wins (`status = _WORKFLOW_TERMINAL_STATUS[event_type]` on each terminal event). `JobPipelineWorkflow.run` / `ApplyWorkflow.run` catch `ActivityError`, return a failure result, emit `WorkflowFailed`, then RETURN normally -> Temporal marks the execution COMPLETED while the projection row is already "failed" (proven by `test_workflow_finalize.py::test_finalize_records_failed_on_stage_failure`: `execute_workflow` returns normally yet the row asserts `status == "failed"`).
Reproduction: Reasoned from code; confirmed the completed-with-failure shape via the finalize test. TOCTOU window = [reconciler reads `open_workflow_runs`, reconciler `describe`s that row].

Because stage/apply failure is encoded in the workflow's return value, Temporal reports COMPLETED even though finalize recorded WorkflowFailed. If a reconciler tick reads a row as `in_progress` (before finalize's projection refresh) and then describes that execution after it reached COMPLETED, it records a second terminal event (WorkflowCompleted); the fold's last-terminal-wins then flips the row from "failed" to "succeeded". A genuinely failed run is shown as succeeded — the precise failure-visibility guarantee P0 exists to provide. The headline kill-worker demo is unaffected (finalize always precedes COMPLETED), so this is a backstop-path race, hence Medium. No test exercises the COMPLETED mapping or the finalize-vs-reconciler interaction.

Required fix: Make the reconciler write conditional on the row still being open — re-SELECT status immediately before recording in `_record_reconciled_outcome` and skip if already terminal; and/or make `_project_workflow_run` prefer the FIRST terminal event so a finalize-recorded outcome is never overwritten by a later reconciler event. Add a regression test: seed a completed-with-failure ("failed") row, run the reconciler against a COMPLETED describe, assert the status stays "failed".

workers/automation/src/jobhunter/pipeline/workflow.py
```diff
@@ -114,6 +118,61 @@ class JobPipelineWorkflow:
 
     @workflow.run
     async def run(self, payload: JobPipelineWorkflowInput) -> JobPipelineWorkflowResult:
+        started_at = workflow.now()
+        await emit_workflow_started(
```
ebarti, line 122: PR Reviewer - pr-reviewer

Severity: Low
Gate: Pass
Evidence: `await emit_workflow_started(...)` runs BEFORE the try/except that records outcomes (pipeline/workflow.py:122-129; same in apply/workflow.py:73-80). `emit_workflow_started` executes the `record_workflow_started` activity, which retries `maximum_attempts=5` (finalize.py `_FINALIZE_RETRY`).

If the start-marker activity exhausts all 5 retries due to a persistent local SQLite write failure (worker death is different — Temporal re-delivers and retries on the next worker), the exception propagates out of `run` before any row exists. Temporal marks the execution FAILED, but `workflow_run_projections` has no row, so `open_workflow_runs()` never returns it and the describe-reconciler cannot backstop a row that was never created — the failed run is invisible in-app. Narrow (requires total DB-write failure), hence Low, but it is the one path where neither finalize nor the reconciler yields visibility.

Required fix: Acceptable to defer with a documented note; or guarantee a terminal record even when the start marker cannot be written (e.g., create the projection row at the workflow-start dispatch surface, or emit WorkflowFailed if the start marker fails).

workers/automation/src/jobhunter/infrastructure/temporal/finalize.py
```diff
@@ -0,0 +1,274 @@
+"""Finalize activities + workflow-side helpers for Temporal loop closure (P0).
+
+Every workflow emits a ``WorkflowStarted`` marker at the top and exactly one
+terminal event (``WorkflowCompleted`` / ``WorkflowFailed`` /
+``WorkflowCanceled``) through a finalize activity that always runs — including
```
ebarti: PR Reviewer - pr-reviewer

Severity: Low
Gate: Pass
Evidence: finalize.py:4-9 module docstring states the terminal event is recorded by "a finalize activity that always runs — including the cancellation path, via ``asyncio.shield`` so the record survives the cancel." There is no `asyncio.shield` anywhere in this module, and pipeline/workflow.py:130-136 + apply/workflow.py comments (and PR deviation 3) state the opposite: the cancel path deliberately does NOT record via finalize (Temporal cancels newly-scheduled activities during cancellation); the describe-reconciler owns WorkflowCanceled.

The docstring documents a cancellation guarantee (shield) that does not exist and contradicts the actual design, on a safety-critical module — a maintainer could wrongly assume finalize covers cancel.

Required fix: Correct the docstring: finalize records terminal events on the success and exception paths only; cancellation/timeout/termination/crash are terminalized by the describe-based reconciler.

apps/web/src/routes/runs.$runId.tsx
```diff
@@ -1,12 +1,12 @@
...
 export const Route = createFileRoute("/runs/$runId")({
-  component: ApplyRunRoute,
+  component: WorkflowRunRoute,
 });
 
-function ApplyRunRoute() {
+function WorkflowRunRoute() {
   const { runId } = Route.useParams();
-  return <ApplyRunDrawer runId={runId} />;
+  return <WorkflowRunDrawer runId={runId} />;
```
ebarti, line 11: PR Reviewer - pr-reviewer

Severity: Low
Gate: Pass
Evidence: This route now renders `WorkflowRunDrawer`. `git grep -n ApplyRunDrawer` against the PR head shows the only remaining references are its own `views/dashboard/ApplyRunDrawer.stories.tsx` and `ApplyRunDrawer.a11y.test.tsx` — no production import.
Reproduction: Reproduced with `git grep -n ApplyRunDrawer origin/feat/temporal-p0-visibility-loop-closure`.

`apps/web/src/views/dashboard/ApplyRunDrawer.tsx` is now orphaned production code after this route swap. The repo's standing directive is rip-and-replace / no dead legacy path, so superseded code should be removed rather than left behind.

Required fix: Confirm nothing else renders `<ApplyRunDrawer>` (the dashboard recent-apply panel deep-links to `/runs/:runId`, which now resolves to WorkflowRunDrawer), then delete `ApplyRunDrawer.tsx` and its stories/a11y test; or note explicitly why it is retained.

apps/api/src/read-model.ts
```diff
@@ -4806,45 +4832,59 @@ function columnNames(db: SqliteDatabase, tableName: string): Set<string> {
...
+  if (hiddenJoin) {
     where.push("h.job_url IS NULL");
   }
   const whereSql = where.length ? ` WHERE ${where.join(" AND ")}` : "";
-  const rows = allRows<ApplyRunProjectionRow>(
+  const rows = allRows<WorkflowRunProjectionRow>(
     db,
-    `SELECT arp.* FROM apply_run_projections arp${deletedJoin}${hiddenJoin}${whereSql}
-     ORDER BY arp.started_at DESC, arp.run_id DESC`,
+    `SELECT wrp.*${applySelect} FROM workflow_run_projections wrp${applyJoin}${deletedJoin}${hiddenJoin}${whereSql}
```
ebarti, line 4886: PR Reviewer - pr-reviewer

Severity: High
Gate: Fail
Evidence: This function repoints the unified runs read model to `FROM workflow_run_projections wrp` (line 4886), with the `tableExists(db, "workflow_run_projections")` guard at line 4851 returning an empty list when the table is absent. The e2e seed `apps/api/test/qa-seed.ts` (not in this PR's diff) hand-rolls its own schema and seeds only `apply_run_projections` (`CREATE TABLE` at :246, `INSERT ... qa-run-1` at :501); it neither creates nor seeds `workflow_run_projections`. Against the e2e seed DB, `listWorkflowRuns` therefore hits the missing-table guard and returns empty, so `/runs` renders zero rows.
Reproduction: Reasoned from code (read-source repoint + un-updated seed) and independently reproduced by QA running the e2e suite. Failing specs at PR head: `apps/web/e2e/tests/runs.spec.ts:3` (expects the "Open workflow qa-run-1 in Temporal Web UI" deep-link), `apps/web/e2e/tests/route-visual-qa.spec.ts:330`, and `:444`.

This PR moves the unified runs read model from `apply_run_projections` to `workflow_run_projections` but does not update the e2e seed, which still only populates `apply_run_projections` (and whose in-file comment still claims the view reads the apply table and normalizes the legacy `finished` status). The `/runs` product-path e2e coverage is red at PR head. Production runtime is unaffected — `apps/api/src/projections.ts` `ensureSchema` plus the Python sole-writer create and populate `workflow_run_projections` — so this is a test-fixture regression rather than a runtime defect, but it breaks required e2e verification for a user-facing surface and blocks merge under the repo's QA gate.

Required fix: Extend `apps/api/test/qa-seed.ts` to `CREATE TABLE workflow_run_projections` (mirroring the production DDL in `sqlite_projection_store.py` / `projections.ts`) and `INSERT` a coherent terminal `qa-run-1` row (`workflow_id`, `workflow_type`, `status`, `started_at`/`finished_at`, `input_summary_json`), plus matching `Workflow*` rows in seeded `job_events` so the run drawer timeline is consistent. Keep the existing `apply_run_projections` seed for the apply-enrichment LEFT JOIN, and update the stale in-file comment. Add a note that the direct insert simulates the Python writer in the test DB. Acceptance: `runs.spec.ts:3`, `route-visual-qa.spec.ts:330`, and `:444` green at PR head. (Tracked as team task #43.)

workers/automation/src/jobhunter/cli.py
ebarti: PR Reviewer - pr-reviewer — verified fixed at 585cc538 (commit 3996ade1).

Two-layer fix, both pinned by regression tests that are red pre-fix:
- **Layer 1** — `_record_reconciled_outcome` now takes `BEGIN IMMEDIATE`, re-SELECTs the row, and returns without emitting when the row is already terminal (skip-if-terminal). Pinned by `test_reconciler_does_not_overwrite_already_terminal_row`, which asserts the row stays `failed` AND that no `WorkflowCompleted` event is even written.
- **Layer 2** — the fold `_project_workflow_run` is now first-terminal-wins (`if status in _WORKFLOW_TERMINAL_STATUSES: continue`), so a later reconciler-mapped `WorkflowCompleted` cannot overwrite an earlier finalize `WorkflowFailed`. Pinned by `test_fold_is_first_terminal_wins`.

The real reconciler path passes a `get_connection()` conn after a read-only SELECT, so `own_txn` is True and `BEGIN IMMEDIATE` fires. Python suite 1679 passed. Resolving.

workers/automation/src/jobhunter/pipeline/workflow.py
ebarti: PR Reviewer - pr-reviewer — accepted as deferred to P1b.

The common case (start marker missing but an outcome event present) is now covered by the create-if-missing fold, proven by `test_outcome_without_start_marker_still_terminalizes`. The residual case — the start-marker activity exhausts its 5 retries AND the workflow aborts before any outcome emits, so no projection row exists and the reconciler (which only scans existing rows) can't find it — is a structural interruptibility concern, documented in the PR body as deferred to P1b, with the start-marker-before-try placement kept deliberately for cancel-safety.

I accept the deferral. Flagging for P1b: the fix should let the reconciler (or a Temporal-execution lister) discover runs that never got a projection row, otherwise this narrow visibility hole persists. Not a P0 blocker (Low; requires the finalize DB write to fail 5x at start).

workers/automation/src/jobhunter/infrastructure/temporal/finalize.py
ebarti: PR Reviewer - pr-reviewer — verified fixed at 585cc538 (L3). The docstring no longer claims cancellation is recorded via `asyncio.shield`. It now correctly states cancellation is NOT recorded by finalize (Temporal cancels newly-scheduled activities during cancellation; no shield exists) and that the describe-based reconciler owns cancel mapping. Resolving.

apps/web/src/routes/runs.$runId.tsx
ebarti: PR Reviewer - pr-reviewer — verified fixed at 585cc538 (L5). `ApplyRunDrawer.tsx` (+ its story and a11y test) and its now-orphaned sole consumer `useApplyRunQuery` (+ hook test, type test, barrel export) are deleted. `rg "ApplyRunDrawer|useApplyRunQuery" apps/web/src` returns zero matches; web:check clean, web test 895 passed, test-d 9 passed. Resolving.

apps/api/src/read-model.ts
ebarti: PR Reviewer - pr-reviewer — verified fixed at 585cc538 (commit 1a18c4cc).

`qa-seed.ts` now creates the full `workflow_run_projections` table and seeds `qa-run-1` (a succeeded `ApplyWorkflow` with a WorkflowStarted/WorkflowCompleted timeline), plus the matching `Workflow*` rows in `job_events`, and keeps the `apply_run_projections` row so the LEFT-JOIN job title/employer still hydrate.

I re-ran the /runs e2e at head on isolated ports (8796/5196, fresh seeded DB): **`runs.spec.ts` and the run-drawer "Workflow run details" overlay test pass** (4 passed across the two spec files). The only failure was `route-visual-qa.spec.ts:507` (job-drawer requirement-fit *visual snapshot*, a devicePixelRatio dimension mismatch 1174x375 vs 559x447) — a pre-existing, environment-driven failure on an unrelated surface this PR does not touch, and one of the 5 pre-existing failures the PR body documents. Resolving.

PR #232, opened 2026-07-03
docs(plans): Codex implementation spec for phases P1b-P5

ebarti:
## Purpose — this PR is the standing instruction set for Codex

This PR intentionally **stays open** while phases P1b–P5 are implemented. Its single file, `docs/plans/2026-07-03-temporal-rearch-implementation-spec.md`, is everything an external implementing agent needs: the diff of this PR **is** the assignment.

## What the spec contains

- **§0 Protocol:** one phase per PR in order P1b → (P2 ∥ P3) → P4 → P5; symbol-first anchors with a verify-then-STOP rule; non-negotiable ground rules (worktree off fresh main, Conventional Commits, never commit data/secrets, never run real apply, never weaken parity tests); the verification command matrix; the PR report template.
- **§3 Substrate contract:** exactly what P0 (#TBD) and P1a (#231) provide on `main`; the implementer must verify each symbol before starting and STOP if any is missing.
- **§4–§8 One section per phase (P1b, P2, P3, P4, P5):** objective, merged-preconditions, exact branch + PR title, *files you may touch* / *files you must NOT touch* lists, numbered work items with code-level prescriptions (error taxonomy verbatim, retry tables, settings plumbing per layer, exact UI copy, CDP dry-run recipe, DDL), mandated deletions each with a grep pre-check, the full test list, and a **binary Definition of Done checklist**.
- **§9 Cross-phase acceptance:** the kill-worker demo.
- **§10 STOP conditions.**

## How to verify each phase

Every phase's Definition of Done is objectively checkable: the §0.3 command matrix (pytest/ruff/api:check/api:test/web suites/`pnpm test`) with required results, deletion greps that must return empty, and named fault-injection tests (e.g. kill-after-intent → `needs_verification`, kill-mid-tailor → resume at failed step, kill-mid-discover → discovery completes after restart).

## Sequencing gate

**P1b must not start until the P0 PR is merged to `main`.** When P0 merges, §3 of this spec will be reconciled against what P0 actually shipped and a "ready for Codex" comment will be posted here.

## Validation

Docs-only. The spec was authored from the 2026-07-02 file:line-verified resilience audit and the merged architectural plan (#230); anchors are symbol-first so line drift cannot mislead the implementer.

PR #234, opened 2026-07-03
docs(plans): OSS release remediation implementation spec for Codex

ebarti:
## What

Adds `docs/plans/2026-07-03-oss-release-remediation-spec.md` — a self-contained, prescriptive implementation spec (same format as the temporal spec in #232) for making the repository safe to publish as open source while preserving every capability: live submit, CapSolver, email applications (rebuilt as a controlled owned-send path), LinkedIn/Indeed defaults, AGPL-3.0-only, DCO, and the existing git history.

Three workstreams, each item one PR with binary Definition of Done + verification commands:

- **W0 — Privacy & release hygiene** (from `main`, first): untrack the private planning corpus, scrub owner-derived fixture data, `release_check.py` v2 (case-insensitive needle classes, file-type checks, structural checks, source tripwires, self-test), an unfiltered privacy CI gate, publish-workflow guard, and a disposition gate for the internal concerns document.
- **W1 — Apply & runtime safety** (stacked after temporal P5; explicitly builds on P2's substrate — approval gate in the atomic claim, `ApplySubmitIntended`, `needs_verification`, CDP dry-run guard — and never re-implements it): approval freshness binding (materials/profile/URL + dry-run-evidence precondition with explicit partial override), dry-run guard hardening (coverage classification, `dry_run_violation` replacing the applied→dry-run coercion, live-mode intent breadcrumbs), agent sandbox (explicit `--allowedTools`, minimal env, `--max-budget-usd`, scoped Gmail), typed attestations replacing fabricated screening defaults, out-of-band credential typing, CapSolver moved into an owned MCP tool, the email-application flow built properly (agent detects; owner-approved recipient; owned send with intent), and fail-closed surface defaults (`jobhunter apply` dry-run by default + legacy live-path deletion).
- **W2 — Public surface & governance** (parallel where files allow): PyPI distribution rename (owner picks; repo/product name unchanged), responsible-use docs + doctor warnings, `Sec-Fetch-Site` CSRF gate, token-based AI spend ledger + ceilings, DCO enforcement, and truthful Tier-2 auth-chain reporting.

Sequencing against the in-flight temporal program is explicit: W0 first (one rebase of the temporal stack expected), W1 stacked after P5, W2.3 after #233, W2.6/doctor items after P5. Owner decision checkpoints are enumerated (dist name, spend defaults, risk acceptances, the visibility flip itself).

## Why

The going-public audit and plan reviews established that release safety requires both current-tree privacy hygiene and apply-path safety gates, and that the apply-safety slice must build on the temporal P2 work rather than collide with it. This spec encodes the owner's locked posture (preserve capabilities; comply via hard gates + disclosure) in a form Codex can implement literally, one gated PR at a time.

## Validation

- `python3 scripts/release_check.py` — the new spec file produces **zero** findings (verified by filtering the scanner output for the file path). Pre-existing findings in fixture files remain, as expected; they are exactly what W0.2 removes.
- Docs-only change: no code paths touched; no test/build impact.
- The spec itself contains no owner identity strings, per its own §0.2 ground rule.

## Deviations

None.

## Deferred

Implementation of every workstream item — this PR is the spec only.

### DIFF
--- a/docs/plans/2026-07-03-oss-release-remediation-spec.md
+++ b/docs/plans/2026-07-03-oss-release-remediation-spec.md
@@ -63,7 +63,7 @@ of one temporal-stack rebase — an owner call, not the default.
 Sequencing summary:
 
 ```
-now         ──► W0.1→W0.2→W0.3→W0.4→W0.5→W0.6   (first: privacy CI gates every later PR)
+now         ──► W0.1→W0.2→W0.5→W0.3→W0.4→W0.6   (first: privacy CI gates every later PR)
             ──► W2.1, W2.2(docs), W2.5            (fully parallel)
 #233 merged ──► W2.3
 P2 merged   ──► W1.1→W1.2→W1.3→W1.4→W1.5→W1.6→W1.7  (stacked; do NOT wait for P4/P5)
@@ -268,6 +268,11 @@ actually occurred, not just four exact-case tokens, and is itself tested.
 **Branch:** `feat/w0-3-release-check-v2` ·
 **PR title:** `feat(release): broaden release_check needles, file-type checks, and add a self-test`
 
+**Ordering:** stacks on W0.5, not W0.2 — item 7 fails on any tree where
+the tag trigger is live under the blocked distribution name, which
+includes the post-W0.2 tree. Section numbers are stable IDs, not
+sequence; see §0.1 and the W0.5 ordering note.
+
 Work items:
 1. **Case-insensitive identity matching.** All identity needles match
    case-insensitively.
@@ -302,7 +307,9 @@ Work items:
 7. **Structural checks:** fail if any tracked path starts with
    `.planning/`; fail if `.github/workflows/publish.yml` has an enabled
    tag trigger while `workers/automation/pyproject.toml` still declares the
-   PyPI-blocked distribution name (this check is retired by W2.1).
+   PyPI-blocked distribution name. W0.5 lands before this item, so the
+   check is green on arrival; it exists to prevent a silent re-enable and
+   is retired by W2.1. Do not soften it with a bypass flag.
 8. **Source tripwires** (pinned to specific files): fail if
    `workers/automation/src/jobhunter/apply/prompt.py` interpolates the
    CapSolver key env var into prompt text, contains the hardcoded
@@ -321,7 +328,7 @@ Work items:
 
 **Definition of Done**
 - [ ] `python3 scripts/release_check.py` exits 0 with zero findings on the
-      post-W0.2 tree, and non-zero when any self-test violation is
+      post-W0.5 tree, and non-zero when any self-test violation is
       introduced.
 - [ ] Byline collision test passes.
 - [ ] Self-test runs in the Python test suite.
@@ -353,13 +360,24 @@ Python/TS CI workflows keep their path scoping.
 **Branch:** `ci/w0-5-guard-publish` ·
 **PR title:** `ci(release): disable tag publishing until distribution rename`
 
+**Ordering:** lands immediately after W0.2 and BEFORE W0.3, out of
+numeric order. W0.3 item 7 fails whenever the tag trigger is live under
+the blocked distribution name — true of the tree today — so the trigger
+must be disabled first for W0.3's exit-0 DoD to be satisfiable; landing
+this first also closes the accidental-publish window sooner. Do not
+resolve the conflict the other way (a bypass flag on item 7): the item-8
+`--strict-prompt` pattern is reserved for tripwires whose fix is gated on
+W1, and this one's fix is the two-line edit right here.
+
 Work items: change `.github/workflows/publish.yml` trigger to
 `workflow_dispatch` only, with a header comment pointing at W2.1. The W0.3
-structural check enforces this cannot silently revert.
+structural check (landing next in the stack) enforces this cannot
+silently revert.
 
 **Definition of Done**
 - [ ] No `push`/tag trigger remains in `publish.yml`.
-- [ ] `python3 scripts/release_check.py` passes.
+- [ ] `python3 scripts/release_check.py` passes (v1 at this point in the
+      stack).
 
 ### W0.6 — CONCERNS disposition gate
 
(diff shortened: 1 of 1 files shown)


ebarti:
Review response — all five findings accepted and addressed in the latest commit:

- **W0.6 (Blocker):** disposition table is now an explicitly PRIVATE, off-repo owner artifact (untracked local `.planning/` or owner-designated location). PR comments/descriptions/issues are named as forbidden delivery channels since they flip public with the repo; a new DoD line enforces it.
- **W1.2 (Blocker):** the item now replaces P2's interception policy wholesale (the plan/spec scoping discrepancy is called out): dry-run allows only GET/HEAD, blocks every other method to every origin from every frame/target; WebSocket construction is blocked at the DOM layer with `Network.webSocketCreated` as a worker backstop; coverage classification switched from same-site to CDP resourceType (Document/XHR/Fetch demote from any origin; Ping/Beacon don't); live breadcrumbs record every mutating continuation across all origins, and rewind-instead-of-park now requires a zero-continuation log — never origin/type filtering. Fixtures added: cross-origin iframe ATS-vendor POST, redirect-then-POST, WebSocket page.
- **W1.1 (High):** `partial_override` is now `partial_override_run_id` — it must reference an existing partial-coverage dry-run for the current materials generation and URL. Override weakens full→partial, never waives evidence; new `override_evidence_invalid` skip reason plus refusal fixtures (no-run and stale-run arms) and a DoD line.
- **W1.3 (High):** `browser_file_upload` is excluded; a new owned `upload_artifact(kind)` tool resolves the run's reviewed artifact path itself (model never supplies a path) and attaches via `DOM.setFileInputFiles`. Raw `search_emails`/`read_email` are no longer exposed to the apply agent — replaced by `get_verification_code()` which returns only extracted OTP codes/links within the spawn-fixed sender/time scope. Exfiltration fixtures are now automated at the tool layer (decoy-body-secret unreachable; out-of-scope empty; out-of-run upload refused), with the full-agent live probe retained as manual defense-in-depth.
- **W1.5 (Medium):** `type_credential` now requires, via the server's own CDP inspection: correct input type, visibility (non-zero box, in-viewport, no hidden/zero-opacity ancestors), top frame or visible same-registrable-domain frame, and document focus — distinct error per failed check, refusal fixtures for hidden/off-screen/cross-origin-iframe fields. W2.2 gains the unique-dedicated-job-site-password guidance.

Spec remains release_check-clean (zero findings for this file).

ebarti:
Sequencing rework after a full collision analysis against the PR #232 spec (all five phase file-ownership lists):

- **§0.1 is now a per-item gate table** instead of a blanket "W1 after P5". Gates are defined as "merged to main" — never branch from the unmerged temporal stack.
- **Parallel immediately (Gate: none):** W0.1–W0.6, W2.1, W2.2 (docs part), W2.5. W2.3 unblocks at PR #233.
- **W1.1–W1.7 unblock at P2, not P5.** P4/P5 touch almost none of their files (residual contention: P5 adds `dailyBudgetUsd` to contracts + settings form — trivial rebases). W1.4 is marked soft-gated: semantically independent of P2, owner may pull it forward if P2 stalls.
- **Gate P5:** W1.8 (`cli.py`/`actions.py` are P5-owned, and P5 itself deletes the in-process branches W1.8 targets — item now says verify-don't-re-delete), W2.6, W2.2's doctor warnings (with a note that P4 rewrites `discovery/**`, so board-default seams must be re-located by symbol).
- **W2.4 rewritten as a delta on P5's spend system.** P5 already ships `llm_spend`, `dailyBudgetUsd` (default 25), the `check_spend_budget` preflight, and dual-client usage recording — the previous W2.4 would have built a duplicate. It now adds only what P5 lacks: per-lane attribution at the same recording seam, apply-lane capture from the `claude_code_cli.py` result-stats parse (P5 hooks `llm.py` + agent-SDK client, which the apply subprocess bypasses), per-lane token ceilings enforced inside P5's preflight, and doctor/health per-lane visibility. Explicit STOP if a second spend table or preflight appears.
- **Fully-delivered #232 view** is stated in §0.1: all gates collapse; W0 first (privacy CI still gates every later PR), W1.1→W1.8 stacked with W2 parallel, optional two-lane W1 split ({W1.1→W1.2→W1.7} ∥ {W1.3→W1.4→W1.5→W1.6}) only under active coordination.
- Critical-path note: P5 sits on the release path only via `cli.py` items + W2.4; pulling W1.8 forward if P5 stalls is an enumerated owner call.

Spec remains release_check-clean.

ebarti:
All three findings accepted:

- **Localhost exemption (High):** removed entirely — W1.2 now states there is NO exemption, explicitly removing P2's carve-out. The harness only needs GET to serve pages; mutating page-origin requests to localhost are blocked and recorded like any other, which is what makes the harness's zero-receipt assertion meaningful. DoD updated to say "no localhost carve-out exists".
- **Needle-capture ordering (High):** W0.2 gains "Work item 0" — before any replacement, the exact forbidden values are captured into a private off-repo needle inventory (same location discipline as the W0.6 artifact, same never-commit/never-post rule), with a new W0.2 DoD line. W0.3 now sources needles from that inventory, with pre-W0.2-parent-commit harvesting via git history as the documented fallback (history is kept per §1, so the fallback is always available).
- **Server shape (Medium):** pinned exactly as recommended — W1.3 creates the dedicated owned apply-tools MCP server (`infrastructure/apply_tools/mcp_server.py`) hosting `upload_artifact`; W1.5 extends it with `type_credential` ("do NOT create a second owned server"); W1.6 adds `solve_captcha` to it with the solver implementation in `infrastructure/captcha/` as an imported module, key confined to the server's env block, and the allowlist parity test covering both key-present and key-absent configurations.

Spec remains release_check-clean.

ebarti:
**Implementation-round amendment (d5199650)** — from Codex's W0.3 STOP, upheld.

W0.3 item 7 (scanner fails while `publish.yml` has a live tag trigger and `workers/automation/pyproject.toml` still declares the blocked distribution name) was unsatisfiable alongside W0.3's own exit-0 DoD, because the trigger-disabling edit belongs to W0.5, which was ordered *after* W0.3. Verified on main at a33f1169: the `v*` trigger is live and the old name is declared, so the check would fire on arrival.

Resolution: **W0 order is now W0.1→W0.2→W0.5→W0.3→W0.4→W0.6.** Section numbers stay stable IDs. Item 7 is unchanged (explicitly: no bypass flag — the item-8 `--strict-prompt` pattern stays reserved for W1-gated tripwires). W0.3's DoD basis moves from the post-W0.2 tree to the post-W0.5 tree; ordering notes added to both sections. Side benefit: the accidental-publish window closes one PR sooner.

Status: W0.1 (#242) and W0.2 (#243) merged with full sweeps green.

PR #235, opened 2026-07-03
feat(worker): raise classified errors into Temporal retry and make activities interruptible (P1b)

ebarti:
## What
- Add the classified worker error taxonomy and map activity failures into Temporal `ApplicationError` types.
- Remove whole-stage exception-to-status swallows so Temporal retry policy owns real failures while `_run_stage_observed` still records StageFailed evidence.
- Add cooperative cancellation events through activity wrappers, discovery sources, enrichment, scoring, tailoring, and cover loops.
- Replace shielded activity execution with a bounded activity executor, cooperative cancel wait, abandoned-thread metric/logging, and worker concurrency controls.
- Add per-stage retry policies, workflow cancellation outcomes, dispatch-time workflow-start projection rows, and the P1b score pending-count parity cap.
- Update P1b docs and regression coverage for transient retry, non-retryable fail-fast, cancellation, dispatch visibility, and score-cap behavior.

## Why
Plan PR #230 and phase P1b of the Temporal-native implementation spec in PR #232 require worker failures to be raised and classified so Temporal can apply retry semantics directly, while activity cancellation becomes cooperative and bounded. This phase also closes the shipped P0 projection visibility/cancellation gaps before the later per-job workflow and apply-safety phases build on the same workflow-run read model.

## Validation
- `uv --project workers/automation run --extra dev pytest -q`
  - `1709 passed in 26.46s`
- `uv --project workers/automation run --extra dev ruff check .`
  - `All checks passed!`
- `pnpm api:check`
  - `tsc --noEmit --project tsconfig.json`
  - exit 0
- `pnpm api:test`
  - `Test Files  19 passed (19)`
  - `Tests  323 passed (323)`
- `pnpm web:check`
  - `tsc --noEmit --project tsconfig.json`
  - exit 0
- `PATH="/Users/eloibarti/.cache/codex-runtimes/codex-primary-runtime/dependencies/node/bin:$PATH" pnpm --filter @jobhunter/web test`
  - `Test Files  145 passed (145)`
  - `Tests  895 passed (895)`
  - `Test Files  9 passed (9)`
  - `Tests  9 passed (9)`
  - `Type Errors  no errors`
- `pnpm web:build`
  - `7195 modules transformed.`
  - `built in 5.44s`
- `pnpm test`
  - API: `Test Files  19 passed (19)`, `Tests  323 passed (323)`
  - Web build: `7195 modules transformed.`, `built in 4.78s`
  - Python: `1709 passed in 28.21s`
- `git diff --check`
  - no output
- `grep -n "asyncio.shield" workers/automation/src/jobhunter/infrastructure/temporal/run_in_activity.py`
  - no output
- `grep -n "\"error: " workers/automation/src/jobhunter/pipeline/runner.py`
  - no output

## Deviations
- The web Vitest command was run with the bundled Node v24.14.0 first on `PATH`. The system Node v26.3.1 in this shell exposes native `localStorage` behavior that makes jsdom storage unavailable unless `--localstorage-file` is provided; that flag then locks under parallel Vitest workers. No web source or test files were changed for this environment issue. API/root commands ran with the system Node because the native `better-sqlite3` module in this install was built for that runtime.

## Deferred
- The remaining repo-wide `asyncio.shield` occurrence under `apply/activities.py` is intentionally left for P2, which owns apply cancellation-drain removal and repo-wide shield completion.
- P3, P2, P4, and P5 remain in the stacked series and will target this branch or its successors as specified.


workers/automation/src/jobhunter/pipeline/runner.py
```diff
@@ -599,7 +600,7 @@ def _run_discovery_source(
                     ),
                 },
             )
-            return f"error: {exc}"
+            raise SourceUnavailableError(f"{label} failed: {exc}") from exc
```
ebarti, line 603: PR Reviewer - reviewer

Severity: Medium
Gate: Pass
Evidence: runner.py:603 now does `raise SourceUnavailableError(...)` where main did `return f"error: {exc}"`. The per-source calls in `_run_discover` are unguarded assignments (`stats["jobspy"] = _run_discovery_source(...)` at ~1566, likewise for ats_api/workday/smartextract), so the first source that raises exits `_run_discover` and every later source is skipped. `_DISCOVER_RETRY` is `maximum_attempts=1`, so there is no retry. On main a failing source returned an `"error:"` string and discovery continued with the healthy sources (partial yield). Dead code confirms the old contract: runner.py:1575, 1629, 1677, 1719 still test `stats[...].startswith("error")`, which can never be true now. `test_discovery_source_failure_records_failed_progress` was updated to expect the raise, but no test covers multi-source partial resilience.
Reproduction: Reasoned from code; dead branches verified via grep.

If any single discovery source (JobSpy, ATS, Workday, Smart Extract) hits a transient failure, the whole discover stage aborts and yields zero jobs from the other healthy sources -- a resilience regression versus main's partial-results behavior, on a core flow where scraping sources fail transiently. Scored Medium (not High) only because the spec's item-2 example ("HTTP 401 from a source -> AuthenticationError") authorizes source-level raising, the spec deletion note says "targeted per-item error records inside loops are fine" (a per-source failure is such a record), and P4 owns discovery decomposition. The team should decide whether this yield regression is acceptable for P1b.

Required fix: Choose explicitly and remove the dead code either way. (a) Restore per-source resilience: catch `SourceUnavailableError` around each `_run_discovery_source` call in `_run_discover`, record the source failure, and continue so healthy sources still yield (mark the stage partial). Or (b) accept the whole-stage abort and delete the four now-unreachable `startswith("error")` branches (1575/1629/1677/1719) and their stale per-source error prints.

workers/automation/src/jobhunter/enrichment/activities.py
```diff
@@ -105,19 +131,21 @@ def _run_selected_enrichment(payload: EnrichActivityInput) -> dict[str, Any]:
...
+        scraper_kwargs["cancel_event"] = cancel_event
+    stats = _run_detail_scraper(get_connection(), **scraper_kwargs)
     elapsed = time.time() - t0
     errors = (
         {"enrich": f"{stats.get('error', 0)} enrichment error(s)"}
         if int(stats.get("error") or 0) > 0
         else {}
     )
-    status = "failed" if errors else "ok"
+    status = "partial" if errors else "ok"
```
ebarti, line 148: PR Reviewer - reviewer

Severity: Low
Gate: Pass
Evidence: enrichment/activities.py:148 returns `status="partial"` together with a non-empty `errors` dict on per-item enrichment failures. enrichment's `_raise_on_failure` (line 175) only inspects `status` and treats "partial" as success (does not raise), whereas the scoring/materials/discovery copies of `_raise_on_failure` raise on `errors or status`. The workflow then calls `_stage_result_failure` (pipeline/workflow.py:392), which returns a failure for any non-empty `errors` -- so the "partial" enrich result still fails the stage. Net behavior matches main (selected-enrich errors fail the stage), so this is not a behavior bug.

The four `_raise_on_failure` copies have drifted (enrich ignores `errors`; the others check it), and returning "partial" alongside a populated `errors` dict is self-contradictory: a reader cannot tell whether per-item enrich errors are tolerated or fatal. This is a maintainability smell that can hide a future defect when someone trusts the "partial" status.

Required fix: Make the enrich path self-consistent -- either return "ok" with per-item errors recorded per item and no top-level `errors` dict (genuinely tolerated), or keep the `errors` dict and align enrich's `_raise_on_failure`/status with the other stages. Consider hoisting one shared `_raise_on_failure` to stop the four copies from drifting.

ebarti:
Fix commit `bf3c8989` addresses the review Medium on the unguarded per-source discovery calls (inline comment at `workers/automation/src/jobhunter/pipeline/runner.py:603`).

**What was wrong:** P1b changed `_run_discovery_source` to raise `SourceUnavailableError` instead of returning an `"error: …"` string, but the per-source calls in `_run_discover` were unguarded — so the first failing source aborted the entire discover stage (and with discover's `max_attempts=1` retry policy, nothing retried). On main a failing source was tolerated and discovery continued with the healthy sources.

**Fix (runner.py + tests only):**
- New `_run_source_isolated` helper wraps each per-source call. A raised `SourceUnavailableError` is caught (the per-source failure is *already* durably recorded inside `_run_discovery_source` before the raise — DiscoveryRunFailed event, per-source StageFailed, failed operational attempt), the stage is marked `partial`, and the loop **continues** with the remaining sources.
- If at least one source succeeds, the stage completes with partial semantics (healthy sources' jobs committed, failed sources visible in per-source detail) and no exception escapes.
- If **every** source fails, the stage raises an aggregated `SourceUnavailableError` naming each source's cause; this propagates through `_run_stage_observed` so the stage-level `StageFailed` observability invariant still fires.
- Deleted the 4 now-dead `startswith("error")` branches (~1575/1629/1677/1719).

**Regression tests** (`test_pipeline_observability.py`, reusing the existing discover-driving fixtures):
- `test_discover_tolerates_failed_source_and_yields_healthy_sources` — JobSpy raises, ATS/Workday/Smart-extract yield → stage completes `partial`, JobSpy's failure recorded per-source, healthy sources committed.
- `test_discover_raises_when_all_sources_fail` — all four raise → aggregated `SourceUnavailableError` propagates from the stage.

**Verification (from the fix worktree at `bf3c8989`):**
- `uv --project workers/automation run --extra dev pytest -q` → **1711 passed** (1709 baseline + 2 new), 0 failures.
- `uv --project workers/automation run --extra dev ruff check .` → **All checks passed!**
- `git grep -n 'startswith("error' -- workers/automation/src/jobhunter/pipeline/runner.py` → empty.

workers/automation/src/jobhunter/pipeline/runner.py
ebarti: @/tmp/p1b_reverify_reply.txt

PR #78, opened 2026-05-19
feat: learn scoring policy from corrections

ebarti:
## Summary
- Add `CorrectionSignal` and correction-derived `CalibrationAnchor` data to versioned scoring policies.
- Persist a new scoring policy version when Python score corrections are applied, while preserving current rubric weights and fit-band thresholds.
- Keep persisted policy anchors non-sensitive by serializing derived fields such as `job_ref_hash`, correction delta/direction, dimension summaries, evidence counts, and source policy metadata instead of raw job URLs or correction rationale.
- Mirror the policy update in the local TypeScript API `correctScore` path because that endpoint writes directly to SQLite.
- Document the correction-to-policy behavior in the scoring architecture, pipeline, and local API docs.

## Validation
- `uv --project workers/automation run --extra dev pytest -q workers/automation/tests/test_scoring_policy.py workers/automation/tests/test_score_use_cases.py workers/automation/tests/test_score_repository.py` -> 36 passed
- `uv --project workers/automation run --extra dev ruff check workers/automation/src/jobhunter/domain/scoring/policy.py workers/automation/src/jobhunter/domain/scoring/use_cases.py workers/automation/src/jobhunter/domain/ports/scoring.py workers/automation/src/jobhunter/infrastructure/scoring/sqlite_repository.py workers/automation/src/jobhunter/domain/scoring/__init__.py workers/automation/tests/test_score_use_cases.py workers/automation/tests/test_scoring_policy.py` -> All checks passed
- `corepack pnpm api:check` -> passed
- `corepack pnpm api:test` -> 6 files passed, 97 tests passed
- `git diff --check scoring/policy-model...HEAD` -> passed

PR #238, opened 2026-07-03
feat(apply): at-most-once submission, binding configurable approval gate, browser-layer dry-run (P2)

ebarti:
## What

- Added deterministic apply workflow IDs at the RPC dispatch seam and enforced single-attempt live apply activity retries while keeping dry-run retry behavior.
- Added durable `ApplySubmitIntended`, real SQLite-backed apply-run persistence, intent-aware recovery, and the `needs_verification` state across Python, contracts, web handlers, fixtures, badges, and tests.
- Replaced the old orphan/reaper behavior: deleted `_rescue_orphaned_running_apply`, `_NoopApplyRunRepository`, `mark_orphans_as_failed`, and the apply `asyncio.shield` cancellation drain; `release_lock` remains but no longer rewinds running work to pending.
- Added default-on `applyApprovalRequired` settings plumbing, API persistence for approve-submit decisions, atomic apply claim enforcement, and approval-aware Apply Review / job action routing.
- Added browser-layer dry-run protection through CDP request blocking/form-submit interception, apply evidence artifacts, and honest verification-confidence derivation.
- Updated README, architecture, QA, and decisions docs for the new apply safety model.

## Why

Implements the Apply safety phase from plan PR #230 and this spec's P2: live application submission must be at-most-once under retry/crash interleavings, live submit must be gated by a binding approval decision by default, dry-run must be physically unable to submit, and ambiguous post-intent crashes must surface for human verification rather than silently retrying.

## Validation

- `pnpm install --frozen-lockfile` — completed successfully; lockfile remained current.
- `uv --project workers/automation run --extra dev pytest -q workers/automation/tests/test_apply_use_cases.py::test_execute_keeps_upload_files_after_local_chrome_launch workers/automation/tests/test_domain_types.py::TestStageState::test_kind_count workers/automation/tests/test_domain_types.py::TestStageState::test_kind_names workers/automation/tests/test_materials_queue_selectors.py::test_acquire_job_picks_up_materials_only_tailored_jobs workers/automation/tests/test_workflow_job_pipeline.py::test_pipeline_workflow_runs_apply_as_child_workflow` — `5 passed in 0.93s`.
- `uv --project workers/automation run --extra dev pytest -q` — `1719 passed in 43.20s`; trailing Temporal heartbeat capability warning after test shutdown was the expected non-failure from §0.3.
- `uv --project workers/automation run --extra dev ruff check .` — `All checks passed!`.
- `pnpm api:check` — TypeScript `tsc --noEmit --project tsconfig.json` completed successfully.
- `pnpm api:test` — `Test Files  19 passed (19)`, `Tests  324 passed (324)`.
- `pnpm web:check` — TypeScript `tsc --noEmit --project tsconfig.json` completed successfully.
- `pnpm --filter @jobhunter/web test src/shared/layout/ThemeToggle.test.tsx` — `Test Files  1 passed (1)`, `Tests  1 passed (1)` after the test setup localStorage shim.
- `pnpm --filter @jobhunter/web test` — `Test Files  145 passed (145)`, `Tests  903 passed (903)`.
- `pnpm --filter @jobhunter/web test-d` — `Test Files  9 passed (9)`, `Tests  9 passed (9)`, `Type Errors  no errors`.
- `lsof -nP -iTCP:8767 -iTCP:5174` — no output; both e2e ports were free.
- `pnpm --filter @jobhunter/web e2e` — `15 passed (1.9m)` with exactly the five §0.3 known baseline failures and no new failures: `dashboard.spec.ts:3`, `jobs-drawer.spec.ts:297`, `jobs-drawer.spec.ts:371`, `route-visual-qa.spec.ts:507`, `token-foundation.spec.ts:349`.
- `pnpm test` — API `19 passed / 324 tests`; web build completed (`✓ built in 6.05s`); Python `1719 passed in 41.27s` with the expected Temporal warning.
- `grep -rn "_rescue_orphaned_running_apply" workers/automation/` — empty.
- `grep -rn "_NoopApplyRunRepository" workers/automation/` — empty.
- `grep -rn "mark_orphans_as_failed" workers/automation/` — empty.
- `grep -rn "asyncio.shield" workers/automation/src/jobhunter/apply/` — empty.
- `sed -n '615,640p' workers/automation/src/jobhunter/apply/launcher.py` — `release_lock` still exists and now only records `LockReleased` + commits; no pending rewind branch remains.
- `git diff --check` — clean.
- `git grep -n "<JobActions\|JobActions(" -- apps/web/src` — only `JobActions.tsx` and the single `JobDetailDrawer.tsx` consumer.
- `git grep -n "_run_stage_observed" -- workers/automation/src apps/api/src apps/web/src packages` — invariant remains in `pipeline/runner.py`; P2 did not alter it.

## Deviations

- Used §0.2 implicit may-touch closure for `apps/web/src/views/jobs/JobDetailDrawer.tsx`, the single `JobActions` consumer outside the context, to pass the approval setting into the component at the composition layer. This was also explicitly added to the amended P2 may-touch block.
- Used §0.2 QA-gate closure for `apps/web/src/test/setup.ts`: the required full web Vitest gate failed because this Node/jsdom runtime exposes `window.localStorage` as unavailable; added a minimal test-only memory storage shim when localStorage is missing. This keeps production code untouched and lets the required gate run.
- P2 e2e did not finish fully green because the spec identifies the exact five failures as pre-existing on `main`; this run had exactly those five and no additional failures.

## Deferred

- Manual QA gate items listed in P2 remain for the human/QA pass: fresh install default approval, direct API bypass proof, approve-to-claim flow, gate-off warning, hostile dry-run fixture, evidence artifact visibility, and `needs_verification` fault-injection resolution.
- P4 and P5 are intentionally left to the next stacked branches.


workers/automation/src/jobhunter/apply/launcher.py
```diff
@@ -639,6 +640,130 @@ def release_lock(url: str, run_ctx: dict | None = None) -> None:
...
+                if not _workflow_terminal_or_gone(conn, workflow_id or run_id):
+                    continue
+                if run_id and _has_apply_submit_intent(conn, url, run_id):
+                    if _has_apply_terminal_result(conn, url, run_id):
+                        continue
+                    set_stage_state(
+                        conn,
+                        url,
+                        "apply",
+                        "needs_verification",
```
ebarti, line 683: PR Reviewer - pr-reviewer

Severity: Blocker
Gate: Fail
Evidence: `acquire_job` excludes only `running`/`succeeded` from claiming — batch path via `NOT EXISTS (... state IN ('running','succeeded'))` (launcher.py:280-285) and targeted path via `_has_active_apply` (only `running`, :132) / `_has_succeeded_apply` (:146). `needs_verification` is excluded nowhere. With `max_apply_attempts=3` (config.py:1069) and `set_stage_state(...,"needs_verification")` leaving `attempt_count` low (COALESCE resets it to 0), a parked row stays eligible, and the approval gate still passes because `approve_submit` remains the latest decision. `test_apply_recovery_after_submit_intent_needs_verification` (test_apply_regressions.py:824) asserts only the state transition, never a subsequent `acquire_job`. `docs/local-reliability-qa.md:51` documents this exact invariant as guaranteed.
Reproduction: Reproduced with a script against a temp DB using the repo's own helpers: acquire (gate ON, approve_submit) → record `ApplySubmitIntended` → `recover_ambiguous_running_apply()` parks `needs_verification` → the very next `acquire_job` (both targeted and non-targeted batch paths) re-claims the row and flips it back to `running`. Output: "BATCH RE-CLAIM (targeted) job returned: True -> state=running"; "BATCH RE-CLAIM (non-targeted) job returned: True -> state=running".

The core at-most-once objective is violated: after a live crash past submit intent, the next apply batch re-claims the `needs_verification` job and submits again (double submission to the employer). Parking the state is not enough while the claim query still treats `needs_verification` as claimable and the prior `approve_submit` still satisfies the gate.

Required fix: Exclude `needs_verification` from BOTH claim paths — add it to the batch `NOT EXISTS` apply-state set and add a `needs_verification` guard to the targeted path (alongside `_has_active_apply`/`_has_succeeded_apply`). Add a regression test that calls `acquire_job` after `recover_ambiguous_running_apply` and asserts it returns `None` while the latest decision is still `approve_submit`.

workers/automation/src/jobhunter/apply/launcher.py
```diff
@@ -819,32 +944,131 @@ def _build_use_case():
...
-    PR 4 of the Temporal stack removed the bespoke ``apply_runs`` table.
-    The ``ApplyRun`` aggregate stays in-memory inside
-    ``SubmitApplicationUseCase`` / ``ApplySaga``; persistence happens
-    via ``record_job_event`` (the launcher emits ``ApplyRunStarted`` /
-    ``ApplicationSubmitted`` / ``ApplicationFailed`` events whose
-    payloads feed ``apply_run_projections``). The repository port is
-    no longer needed but the saga signature still accepts one — this
-    no-op satisfies the protocol without writing anywhere.
-    """
+    def save(self, run) -> None:
```
ebarti, line 965: PR Reviewer - pr-reviewer

Severity: High
Gate: Fail
Evidence: The new real `SqliteApplyRunRepository.save()` persists every aggregate event to `job_events` and is called 7 times during the saga (process_manager.py:163,197,213,232,257,279,319). `run_job` then still calls `_persist_saga_event_timeline(outcome.apply_run, ...)` (launcher.py:1247), which re-writes every non-`_LAUNCHER_OWNED_EVENT_TYPES` event with no `event_id` dedup (launcher.py:1094-1138). Under the old `_NoopApplyRunRepository` (save was a no-op) only the timeline persister wrote these, so there was no duplication.
Reproduction: Reproduced with a script against a temp DB: build an `ApplyRun` with `SagaStarted/BrowserLaunched/AgentStarted/AgentResult`, call `SqliteApplyRunRepository().save(run)` (what the saga does) then `_persist_saga_event_timeline(run, run_id)` (what run_job does). Result — job_events: `SagaStarted x2, BrowserLaunched x2, AgentStarted x2, AgentResult x2`; and `apply_run_projections.events_json` timeline carries each twice.

Every saga wrapper event (including the new `AgentResult` carrying `raw_output` evidence) is doubled in the apply timeline that the Apply Review / audit surface reads — an auditability regression on the exact evidence surface this phase strengthens. `test_apply_saga_writes_full_event_timeline_to_job_events` cannot catch it: it calls `_persist_saga_event_timeline` alone and collapses duplicates via `recorded.setdefault(...)` + membership assertions.

Required fix: Make the two writers mutually exclusive — now that the repository persists during the saga, either drop the post-saga `_persist_saga_event_timeline` call, or dedup it by `event_id` the way `_existing_apply_event_ids` does (and stamp `payload["event_id"]`), or have the repository skip the saga-wrapper events. Add a test asserting each event type appears exactly once in `job_events`/`events_json` after a full run.

workers/automation/src/jobhunter/apply/workflow.py
```diff
@@ -106,10 +112,13 @@ async def run(self, payload: ApplyWorkflowInput) -> ApplyWorkflowResult:
             expected_app_dir=payload.expected_app_dir,
             expected_db_path=payload.expected_db_path,
         )
+        if payload.continuous:
+            workflow.continue_as_new(payload)
```
ebarti, line 120: PR Reviewer - pr-reviewer

Severity: Medium
Gate: Pass
Evidence: The activity is now always dispatched with `continuous=False` and `limit=25` (workflow.py:121,137). With `continuous=False`, `worker_loop` breaks immediately when the queue is empty (launcher.py:1354-1358 — no poll wait). `ApplyWorkflow.run` then calls `workflow.continue_as_new(payload)` with no throttle (workflow.py:115-116). The old design ran the activity in `continuous=True` poll mode, which slept `POLL_INTERVAL` between empty polls. `test_apply_workflow_continuous_batch_is_bounded_and_continues_as_new` mocks `execute_activity`, so the empty-queue path is not exercised.
Reproduction: Reasoned from code: continuous mode + empty queue → activity returns `(0,0)` immediately → `continue_as_new` fires immediately → new run repeats with no delay (hot loop), churning Temporal history and DB claim attempts.

Continuous `ApplyWorkflow` will busy-loop `continue_as_new` whenever the queue is empty, replacing the previous throttled poll with an unbounded spin. Also, the spec's "1 hour of workflow time" bound (item 9) is not implemented — only the 25-iteration bound.

Required fix: Throttle before continue-as-new when no work was done (e.g., `await workflow.sleep(poll_interval)` when `applied+failed == 0`), or keep a bounded in-activity poll. Add a test that drives the empty-queue continuous path and asserts a delay/timer precedes continue-as-new. Consider implementing the 1-hour time bound alongside the 25-iteration bound.

apps/web/src/contexts/operations/jobsListQuery.ts
```diff
@@ -19,13 +19,14 @@ const STATE_RANK: Record<StageState, number> = {
...
   exhausted: 1,
   blocked: 2,
-  running: 3,
-  queued: 4,
-  pending: 5,
-  stale: 6,
-  canceled: 7,
-  skipped: 8,
-  succeeded: 9,
+  needs_verification: 3,
```
ebarti: PR Reviewer - pr-reviewer

Severity: Low
Gate: Pass
Evidence: `apps/api/src/read-model.ts` STATE_RANK ranks `needs_verification: 2` (more severe than `blocked: 3`), while `apps/web/src/contexts/operations/jobsListQuery.ts:22` ranks `needs_verification: 3` (less severe than `blocked: 2`). These two rank tables were byte-identical before this PR; this change diverges them.
Reproduction: Reasoned from code: for a job with one stage `blocked` and another `needs_verification`, the server-side worst-state selection and the client-side `jobsListQuery` worst-state selection will pick different representative states, producing inconsistent badges between API-derived and client-derived views.

Server and client no longer agree on the relative severity of `needs_verification` vs `blocked`.

Required fix: Use the same rank for `needs_verification` in both `read-model.ts` and `jobsListQuery.ts` (pick one position and apply it to both tables).

workers/automation/src/jobhunter/infrastructure/apply/claude_code_cli.py
```diff
@@ -397,6 +397,28 @@ def _parse_result(self, output: str, *, dry_run: bool) -> SubmissionResult:
         return Failed(error="no_result_line", retryable=True)
 
 
+def _applied_confidence(output: str) -> float:
```
ebarti, line 400: PR Reviewer - pr-reviewer

Severity: Low
Gate: Pass
Evidence: Spec item 8 defines `verification_confidence == 1.0` as "structured `RESULT:APPLIED` AND a confirmation artifact exist". The implementation instead returns 1.0 when `_has_confirmation_evidence(output)` matches free-text substrings in the agent's raw output ("application submitted", "thank you for applying", "confirmation:", ...), not the existence of the persisted `apply_confirmation` `job_artifacts` row (claude_code_cli.py:400-417). Separately, the `apply_confirmation` artifact itself is synthetic — `_persist_agent_artifacts` wraps the same `raw_output` in `<pre>` (launcher.py), so it is not independent of `apply_agent_output`. This deviation is not listed in the PR Deviations section.
Reproduction: Reasoned from code + matrix test (`test_applied_confidence_matrix`): "RESULT:APPLIED\nconfirmation: submitted" → 1.0; "RESULT:APPLIED" → 0.6; the 1.0 tier depends on agent-authored text rather than independent confirmation evidence.

The hardcoded 1.0 is correctly removed and a matrix test exists, but the 1.0 tier is driven by fuzzy text matching of the agent's own output rather than the independent confirmation artifact the spec named — a weaker auditability guarantee than intended, and undeclared.

Required fix: Declare the deviation in the PR, and prefer keying the 1.0 tier on an independent confirmation signal (e.g., a real page snapshot/screenshot for `apply_confirmation`) rather than substrings of the agent's own narration. At minimum, gate 1.0 on the structured `RESULT:APPLIED` marker being present in addition to the confirmation signal.

ebarti:
## Gate summary — P2 (review + live QA) — handover to fix pass

**Review: Gate FAIL** — Blocker 1 / High 1 / Medium 1 / Low 3 (see inline comments). **Live QA at stack top a370149c (isolated stack, fresh JOBHUNTER_DIR): Gate FAIL** — same Blocker reproduced live; every other apply-safety gate PASSED.

### Blocker (r3521492989 — QA-corroborated live)
`needs_verification` rows are re-claimed by the NEXT acquisition — both batch (`worker_loop`→`acquire_job`) AND a targeted re-click — flipping parked rows back to `running` and re-running the agent. Root cause (QA): `acquire_job` resets prior state ∉ {pending, running} back to pending (launcher.py:389-401), and neither the batch NOT EXISTS (launcher.py:280-285) nor the targeted active-check excludes `needs_verification`; the prior `approve_submit` is still the latest decision so the approval gate passes again. Defeats at-most-once (spec §5 item 5). Required regression: park via kill-after-intent → follow-up acquisition on BOTH paths leaves the row `needs_verification`, no second agent run.

### Fix scope for this PR
1. The Blocker above (exclusion inside the BEGIN IMMEDIATE claim txn; no auto-requeue path out of needs_verification).
2. High r3521494069 — saga events double-written: real `SqliteApplyRunRepository.save()` persists aggregate events during the saga AND `run_job` still calls `_persist_saga_event_timeline` after, no event_id dedup → SagaStarted/BrowserLaunched/AgentStarted/AgentResult each ×2 in `job_events` AND `apply_run_projections.events_json`. Single writer (repository = the durable mid-saga evidence mechanism); count-based exactly-once test.
3. Medium r3521494102 — continuous ApplyWorkflow busy-loops continue_as_new on an empty queue (no throttle) and the spec §5 item 9 1-hour bound is unimplemented (only the 25-iteration bound). Durable poll sleep + both bounds + workflow-level tests.
4. Low — STATE_RANK divergence: read-model.ts ranks needs_verification=2 vs jobsListQuery.ts=3 (identical before this PR). Unify with one source of truth.

### Deferred (tracked follow-ups — do NOT fix here)
verification_confidence keying on agent-output text + synthetic apply_confirmation artifact (same raw_output re-wrapped); init_db vs feedback.py DDL drift for application_review_decisions.

### QA results detail (a370149c)
Fresh-install default applyApprovalRequired=true (API + web) PASS · UI-bypass with gate ON blocked (stage stays pending, 0 ApplyRunStarted, "Apply approval required" logged) PASS · approve→claim vs blocked-without-approval PASS · toggle-OFF role=alert exact spec copy PASS · CDP dry-run guard live vs local hostile auto-POST page: 0 cross-origin POST/PUT/PATCH, live mode did POST PASS · evidence artifacts (apply_agent_output every AgentResult, apply_confirmation only when applied, served via /v1/artifacts) PASS · settings round-trip + snake_case fallback PASS · apply e2e 15 passed / 5 failed = exactly the documented baseline, zero new PASS. Note (pre-existing, not P2): apply-eligibility requires an APPROVED resume_pdf in job_materials_artifacts.

workers/automation/src/jobhunter/apply/launcher.py
ebarti: PR Reviewer - pr-reviewer

Severity: Blocker
Gate: Pass (fixed)
Evidence: All three re-claim vectors are closed at d62adecb inside the BEGIN IMMEDIATE claim txn: (a) batch path — `needs_verification` added to the NOT EXISTS active-state set (launcher.py:294); (b) targeted path — new `_has_needs_verification_apply` guard rolls back + returns None (launcher.py:157-163, 367-369); (c) state-reset vector — the prior-state reset set now includes `succeeded` and `needs_verification`, so the reset branch no longer rewinds a parked row to `pending` (launcher.py:407-412). No auto-requeue path out of `needs_verification` remains: `recover_ambiguous_running_apply` selects only `state='running'` rows (launcher.py:667-668); the generic `recover_orphaned_running_stages` excludes apply (`ORPHAN_RECOVERY_STAGES=("discover","enrich")`); `release_lock` records an event only, no rewind.
Reproduction: Reproduced live on a temp SQLite with the repo's own helpers (approve_submit recorded, gate ON): acquire → running; record `ApplySubmitIntended`; `recover_ambiguous_running_apply` → `needs_verification`; then BOTH batch AND targeted `acquire_job` returned None and the row stayed `needs_verification`/`APPLY_NEEDS_VERIFICATION`. `test_apply_recovery_after_submit_intent_needs_verification` now performs the follow-up acquisition on both paths and asserts the parked state; `docs/local-reliability-qa.md:51` cites `test_apply_regressions.py`. Full pytest 1719 passed.

Verdict: Fixed.

workers/automation/src/jobhunter/apply/launcher.py
ebarti: PR Reviewer - pr-reviewer

Severity: High
Gate: Pass (fixed)
Evidence: `_persist_saga_event_timeline` (function + the `run_job` call) is deleted, leaving one writer: `SqliteApplyRunRepository.save()` dedups by (run_id, event_id) via `_existing_apply_event_ids` + a `seen` set and stamps `payload["event_id"]` (launcher.py:965-1000, 1016-1040). Coverage equality holds: the saga records-then-saves on every path (success 319, browser-fail, timeout, crash), and `ApplyRun.complete()` only replaces status (records no events), so no wrapper event is dropped after the last save; launcher-owned terminals (ApplicationSubmitted/Failed/DryRunCompleted) are written once by `mark_result`, not by `save()`. `ApplySubmitIntended` has a single writer (recorded on the aggregate at process_manager.py:220, persisted by save; launcher.py:753 only reads it).
Reproduction: Reproduced the real incremental save pattern against a temp DB (8 record→save checkpoints + 2 redundant saves): each of SagaStarted/BrowserLaunched/AgentStarted/ApplySubmitIntended/ClaudeLaunched/AssistantText/ToolUse/AgentResult persisted exactly once. Test is now COUNT-based (`Counter`, `counts[evt_type]==1` in job_events and `event_types.count(required)==1` in events_json). WorkflowStarted-dup follow-up (task #78) untouched. Full pytest 1719 passed.

Verdict: Fixed. Non-blocking nit: the count-based test calls `save()` once with a full aggregate rather than the 7× incremental path; the incremental dedup is sound (verified independently), but a saga-driven DB-backed count test would be stronger.

workers/automation/src/jobhunter/apply/workflow.py
ebarti: PR Reviewer - pr-reviewer

Severity: Medium (addressed) → residual Low
Gate: Pass
Evidence: Empty-queue busy-loop fixed — `if result.applied + result.failed == 0: await workflow.sleep(_APPLY_CONTINUOUS_EMPTY_POLL_DELAY=30s)` before `continue_as_new` (workflow.py:118-120, 69). 25-iteration bound: activity dispatched `limit=25` (test asserts `payload.limit==25`). 1-hour bound: continuous activity `start_to_close_timeout=_APPLY_CONTINUOUS_BATCH_TIMEOUT=1h` (workflow.py:68, 143-147; test asserts `timeout==timedelta(hours=1)`). Empty-queue test asserts `sleep==timedelta(seconds=30)` precedes continue_as_new. `test_apply_workflow_continuous_batch_is_bounded_and_continues_as_new` passes.
Reproduction: Reasoned from code + workflow test run.

Residual (Low, non-blocking): spec §5 item 9 says the loop must `continue_as_new` "after ... 1 hour of workflow time." Here the 1-hour bound is a hard activity `start_to_close_timeout`; a continuous batch exceeding 1h makes the live activity (max_attempts=1) time out and the workflow FAIL rather than continue gracefully. In practice the 25-iteration bound triggers continue_as_new well before 1h, and this still shrinks the spec's cited ~4h silent death to ~1h. Consider a workflow-time-based graceful continue if 25 real live applies > 1h is realistic.

Verdict: Original Medium addressed (throttle + both bounds + tests); residual is Low.

apps/web/src/contexts/operations/jobsListQuery.ts
ebarti: PR Reviewer - pr-reviewer

Severity: Low
Gate: Pass
Evidence: `apps/web/src/contexts/operations/jobsListQuery.ts:18-27` now ranks `needs_verification: 2, blocked: 3`, matching `apps/api/src/read-model.ts:80-88` — the immediate server/client severity divergence is resolved. web:check + web test (903) + test-d (9) pass.
Reproduction: Read both tables at d62adecb.

Residual (Low, non-blocking): the tables are equal-by-value but there is still no single source of truth (shared constant) and no cross-file equality test, so they can silently drift again. The finding's "at minimum equal values + an equality test" is half-met (equal values; test missing). Consider extracting a shared STATE_RANK constant or adding an equality test.

Verdict: Immediate divergence fixed; residual Low (no drift guard).

PR #248, opened 2026-07-04
docs: full accuracy refresh of user, developer, and architecture docs

ebarti:
## What changed

Full documentation accuracy pass, verified claim-by-claim against the code at `ab73ea84` (after the Temporal-native rearchitecture P0–P5 deleted the in-process pipeline). Built from 9 parallel code-grounded audits plus a cross-cutting functionality-coverage sweep; 26 files, +2,399/−2,701.

**Rewritten**
- `docs/job-pipeline-architecture.md` — the old doc substantially described the deleted in-process pipeline. Now documents the six Temporal workflows and 19 activities, the universal workflow envelope (finalize + spend preflight + describe-based reconciler), deterministic IDs/`USE_EXISTING`, error taxonomy → retry mapping, stage-by-stage walkthroughs (incl. the root-workflow preparation fan-out and the three apply safety mechanisms), spend ceiling, off-by-default schedule (+ worker-restart gotcha), persistence map, and the 68-type event → dual projection builders → 250 ms SSE poller flow. All diagrams redrawn.

**Heavily corrected**
- `docs/architecture.md` — System Shape prose + mermaid rebuilt around Temporal (was: "three components", no Temporal in the diagram, dead RPC→CLI edge); full 11-method JSON-RPC truth; ApplyWorkflow retry truth (1s, no non-retryable filter); `workflow_run_projections`; worker-health gate + Host/CSRF hardening; approval-gate claim scoped honestly; span table completed (+6 live spans); ensemble model IDs.
- `docs/ddd-target.md` — reframed as realised architecture (not a future target); 10 domain contexts; 7 projections; `NeedsVerification` stage state; single prefix-dispatched `LlmAdapter` + the 3-SDK analysis ensemble as the second LLM path; deleted the dead in-process dispatch seam and the `PipelineRunManager`/`_StageTracker` narrative; dispatch modes `sync/workflow/streaming`; event-schema home corrected (plain TS union, not Zod/contracts).
- `docs/frontend-target.md` — removed the never-built SSE-Zod validation story; 8 views (was 3); ports/query-keys/mutation-map/test counts refreshed to reality; orphaned (`ScoreBreakdown`) and stub (`useImportJobMutation`) components labeled honestly; CI-gating claims corrected; custom data-grid truth (TanStack Table types only).
[...]

PR #249, opened 2026-07-04
feat(docs): publishable VitePress docs site with Cloudflare Pages deploy

ebarti:
## What

Adds a browseable static docs site generated from the existing `docs/` tree, in preparation for publishing on **jobctl.dev** after the jobctl rename:

- **VitePress 1.6.4** site configured in `docs/.vitepress/config.ts`, consuming the markdown in `docs/` as-is (no frontmatter, no content restructuring). Root scripts: `pnpm docs:dev` / `docs:build` / `docs:preview`.
- **Curated sidebar/nav**: User Guide (`docs/user/`), Developer Guide, Architecture, Reference. `docs/INDEX.md` serves as the homepage (`rewrites`), `docs/developer/README.md` becomes `/developer/`. Local full-text search, dark mode, `lastUpdated`, clean URLs.
- **Publish boundary**: `docs/plans/`, `docs/incidents/`, `docs/backlog.md`, and `docs/delivered.md` are excluded (`srcExclude`) — internal planning docs stay repository-only. A markdown-it rule rewrites any link that resolves outside the published set (repo-root files like `../README.md`, or the excluded internals) into absolute GitHub `blob`/`tree` URLs, so the deployed site ships **zero broken relative links** while VitePress's strict dead-link check stays fully enabled for everything else.
- **Mermaid** diagrams render via `vitepress-plugin-mermaid` + mermaid 11.
- **CI** (`.github/workflows/docs-site.yml`): the build (which fails on dead internal links) runs on every PR touching `docs/` — this is the docs link-integrity gate. On pushes to `main`, a deploy job publishes `docs/.vitepress/dist` to Cloudflare Pages via `wrangler-action@v3`, but only once the `DOCS_DEPLOY_ENABLED` repo variable is `true` and the Cloudflare secrets exist; until then it skips cleanly.

Two workspace-specific landmines are handled in config, with comments:

- The workspace-wide `esbuild@0.28.1` override (security pin) refuses to lower plain destructuring to Vite 5's legacy default browser targets. The site sets `build.target`/`optimizeDeps.esbuildOptions.target` to `es2022` instead of touching the shared pin.
- mermaid's CJS deps (`dayjs`, `@braintree/sanitize-url`) must be pre-bundled for `docs:dev`; under pnpm's strict layout they only resolve through mermaid, hence the nested `optimizeDeps.include` entries.

## Why

The docs refresh (#248) made `docs/` accurate; this makes it publishable and browseable as a real website, regenerated automatically on every merge to `main`, per the plan to host it at jobctl.dev (free tier, PR preview URLs, custom domain attachable later).

## Validation

- `pnpm docs:build` green (strict dead-link check enabled, no `ignoreDeadLinks`).
- Built output audited: 19 pages exactly matching the published set; zero `plans/`/`incidents/`/`backlog`/`delivered` leakage; zero raw `../` hrefs; escaping links correctly rewritten to GitHub URLs (`blob` for files, `tree` for directories).
- Playwright (headless Chromium) sweep of all 18 routes against `docs:preview`: all HTTP 200, **27/27 mermaid diagrams render, zero mermaid syntax errors, zero unrendered containers** — verified against both current `main` content and the #248 overlay (post-merge equivalent). Screenshots inspected (sidebar, search, outline, diagram rendering all correct).
- The sweep found one real content bug: the Apply sequence diagram added in #248 used `;` in message text (a mermaid statement separator), so it never rendered. Fixed on the #248 branch (`05d0a24e`), not here — this PR contains no changes to published doc content.
- `pnpm check` green (contracts/api-client/api/web typechecks + ruff).
- PII sweep of built HTML (usernames, personal emails, phone patterns, `/Users/` paths): clean.
- QA stage: no product UI/API surface changed (tooling + static site only); validation above exercises the actual shipped artifact.

Note: `release-check` fails on `main` for the pre-existing `docs/plans/2026-07-03-oss-release-remediation-spec.md` finding (owned by the OSS W0 work); unrelated to this PR.

## After merge — one-time Cloudflare setup (5 min)

1. Cloudflare dashboard → Workers & Pages → Create → Pages → **Direct Upload** project named `jobctl-docs` (or just let the first Actions deploy create it).
2. Create an API token with **Cloudflare Pages: Edit** permission.
3. GitHub repo → Settings → Secrets and variables → Actions: add secrets `CLOUDFLARE_API_TOKEN`, `CLOUDFLARE_ACCOUNT_ID`; add variable `DOCS_DEPLOY_ENABLED=true`.
4. When the domain exists: Pages project → Custom domains → add `jobctl.dev`.

Merge order note: independent of #248, but merge #248 first if convenient — the site picks up the refreshed content automatically on the next `main` build. `AGENTS.md`/`docs/local-development.md` edits here are additive and may need a trivial conflict resolution against #248.


docs/.vitepress/config.ts
```diff
@@ -0,0 +1,128 @@
...
+  defineConfig({
+    title: "JobHunter",
+    description:
+      "Local-first, AI-assisted job application pipeline: discovery, scoring, tailored materials, and supervised apply.",
+    srcExclude: ["plans/**", "incidents/**", "backlog.md", "delivered.md"],
+    cleanUrls: true,
+    lastUpdated: true,
+    rewrites: {
+      "INDEX.md": "index.md",
+      "developer/README.md": "developer/index.md",
```
ebarti: PR Reviewer - pr-reviewer

Severity: High
Gate: Fail
Evidence: The built artifact `docs/.vitepress/dist/index.html` contains `href="./developer/README"`, but the developer overview is emitted at `/developer/` because this rewrite maps `developer/README.md` -> `developer/index.md`. There is no `dist/developer/README.html` (only `index.html` and `screenshot-playbook.html`). The source link is `docs/INDEX.md:31` — `[developer/README.md](developer/README.md)`. VitePress strips `.md` to `/developer/README` but does not apply the rewrite to this inbound link, and the strict dead-link checker (`ignoreDeadLinks` intentionally unset) does NOT flag it because the source file `developer/README.md` still exists.
Reproduction: Reproduced from the current-config build output: `grep 'developer/README' docs/.vitepress/dist/index.html` -> `href="./developer/README"`; `ls docs/.vitepress/dist/developer/` -> `index.html screenshot-playbook.html` (no `README.html`).

The homepage "Developer Docs" bullet links to `/developer/README`, which 404s on the static host (Cloudflare Pages) / renders the VitePress NotFound page. This contradicts the PR's "zero broken relative links" guarantee and shows the strict dead-link gate has a blind spot for links whose targets were remapped by `rewrites`, so this class of break will not be caught going forward. Mitigation: the developer guide is still reachable via the top nav "Developer" and the sidebar, so it is not orphaned.

Required fix: Repoint the inbound link to the rewritten output. Simplest: change `docs/INDEX.md:31` to `[developer/README.md](developer/)`. Alternatively make link handling rewrite-aware (map any link whose resolved source path is a `rewrites` key to its output path). Re-verify by building and confirming `dist/index.html` links to `/developer/` and that no emitted href points at a path lacking a corresponding `.html`. Consider a post-build check that every internal href resolves to an emitted file, since the VitePress dead-link gate misses rewrite targets.

docs/.vitepress/config.ts
```diff
@@ -0,0 +1,128 @@
...
+  const path = hashIndex === -1 ? href : href.slice(0, hashIndex);
+  const hash = hashIndex === -1 ? "" : href.slice(hashIndex);
+  if (path === "") return null;
+  const repoPath = posix.normalize(posix.join("docs", posix.dirname(pageRelativePath), path));
+  const escapesDocs = !repoPath.startsWith("docs/");
+  const isUnpublished =
+    UNPUBLISHED_PREFIXES.some((prefix) => repoPath.startsWith(prefix)) ||
+    UNPUBLISHED_FILES.has(repoPath);
+  if (!escapesDocs && !isUnpublished) return null;
+  const view = path.endsWith("/") ? "tree" : "blob";
```
ebarti: PR Reviewer - pr-reviewer

Severity: Low
Gate: Pass
Evidence: `rewriteEscapingLink` selects `tree` only when the link path ends with `/` (this line). A link to an unpublished or repo directory written WITHOUT a trailing slash (e.g. `[plans](../plans)` or `](plans)`) resolves to a directory but is emitted as a GitHub `blob/main/...` URL, which 404s on GitHub for directories (only `tree/...` works). No such link exists today — the only directory link in the published set is `plans/` (with trailing slash) — so this is latent, not live.
Reproduction: Reasoned from code; current published content audited clean (only `plans/` with a trailing slash present).

Required fix: Optional hardening, safe to defer. If addressed, infer directory targets from the absence of a file extension rather than a trailing slash, or add a short comment here documenting that directory links must carry a trailing slash so future authors don't emit a dead `blob` URL.

docs/.vitepress/config.ts
ebarti: Fixed in 5b893831, structurally rather than by editing INDEX.md (which #248 also touches, and a content edit would leave the blind-spot class open):

1. `PAGE_REWRITES` is now the single source of truth — fed to VitePress `rewrites` and to the link rule, which rewrites inbound links to relocated pages to their emitted routes. The homepage now ships `href="/developer/"`.
2. New post-build gate `scripts/check-docs-site-links.mjs` (wired into `pnpm docs:build`, so PR CI runs it): every href/src in the built HTML must resolve to an emitted page or asset. Validated both directions — passes on the fixed build (1,478 references / 148 files), and re-doctoring `index.html` back to `./developer/README` makes it fail with exactly that reference.

On the Low: leaving the trailing-slash heuristic as-is per your defer — worth noting GitHub auto-redirects blob/tree mismatches, so a slash-less directory link would still land correctly after one redirect.

ebarti: PR Reviewer - pr-reviewer

Resolved in 5b893831 — verified independently. From a clean `pnpm docs:build`: the homepage now emits `href="/developer/"` (was `./developer/README`), no reference to the old target remains in `dist`, the publish boundary is still clean (19 HTML files; excluded paths appear only as GitHub rewrites), and the checker is green (1478 references / 148 emitted files). Negative test reproduced: doctoring index.html back to `./developer/README` and running `node scripts/check-docs-site-links.mjs` fails with exactly `/index.html -> ./developer/README` (exit 1) — the new post-build gate catches the whole rewrite-blind-spot class, not just this instance. The shared `PAGE_REWRITES` + `routeForRewrittenPage` fix resolves both rewrite entries correctly (`INDEX.md` -> `/`, `developer/README.md` -> `/developer/`). Marking resolved.

PR #250, opened 2026-07-04
fix(pipeline): recover workflow runs and surface worker concurrency

ebarti:
## Summary

- migrate legacy `workflow_run_projections` tables on both Python worker and TypeScript API initialization paths
- rebuild workflow-run projections when workflow events were already watermarked but the Python-owned workflow projection table is missing rows
- make dashboard pipeline progress consult terminal workflow-run state so stale `StageStarted` progress no longer presents dead workflows as running
- expose the running Temporal worker activity-slot configuration through worker heartbeats, `GET /v1/health`, API client types, and the Settings page
- document `JOBHUNTER_MAX_CONCURRENT_ACTIVITIES` as a worker-startup environment setting and keep legacy heartbeat tables readable

## Root Cause

A workflow startup activity persisted `WorkflowStarted`, then synchronously refreshed read-model projections. Existing local databases could still have the older `workflow_run_projections` shape without `input_summary_json`, which caused projection upsert to fail before discovery reached source planning or scraping. Separately, the shared operations watermark could already be past those workflow events, leaving the workflow-run projection empty.

The runtime-concurrency issue was observability: Temporal activity concurrency was controlled by `JOBHUNTER_MAX_CONCURRENT_ACTIVITIES` at worker startup, but the app did not expose the actual running value. That made the Pipelines `Workers` field look like it was controlling Temporal worker capacity even though Temporal worker slots are process configuration.

This PR fixes the immediate schema/backfill failure, makes the UI/API truthful about terminal workflow state, and surfaces the actual Temporal worker activity capacity. A broader follow-up should make projection refresh non-critical to workflow execution and should separately address JobSpy search-combination parallelism.

## Validation

- `uv --project workers/automation run --extra dev pytest -q workers/automation/tests/test_workflow_runs_projection_from_events.py`
- `corepack pnpm --filter @jobhunter/api exec vitest run test/projections.test.ts -t "projection schema upgrades"`
- `corepack pnpm --filter @jobhunter/api exec vitest run test/server.test.ts -t "pipeline progress"`
- `uv --project workers/automation run --extra dev pytest -q workers/automation/tests/test_temporal_concurrency.py workers/automation/tests/test_runtime_identity.py`
- `corepack pnpm --filter @jobhunter/api test -- server.test.ts`
- `corepack pnpm --filter @jobhunter/web test -- SettingsPanel.test.tsx useHealthQuery.test.ts`
- `corepack pnpm --filter @jobhunter/api-client check`
- `corepack pnpm api:check`
- `corepack pnpm web:check`
- `corepack pnpm web:test-d`
- `uv --project workers/automation run --extra dev ruff check workers/automation/src/jobhunter/infrastructure/runtime_identity.py workers/automation/src/jobhunter/infrastructure/temporal/worker.py workers/automation/src/jobhunter/infrastructure/temporal/concurrency.py workers/automation/tests/test_runtime_identity.py workers/automation/tests/test_temporal_concurrency.py`
- `git diff --check` / `git diff --cached --check`
- Restarted the local dev stack from `/Users/eloibarti/.codex/worktrees/fdc8/JobHunter`; `GET /v1/health` reports worker `healthy`, task queue `jobhunter-default`, `maxConcurrentActivities: 4`, and `activityExecutorMaxWorkers: 6`
- Browser QA: `/settings` shows Temporal runtime Activity slots `4`, Executor threads `6`, Startup env `JOBHUNTER_MAX_CONCURRENT_ACTIVITIES`, and Worker health `healthy`; browser returned to `/pipelines`
- Browser reload at `/pipelines` confirmed Discover no longer shows the stale running status and reports the terminal workflow reason instead.


apps/api/src/read-model.ts
```diff
@@ -4479,12 +4480,87 @@ function listPipelineProgress(db: SqliteDatabase): DashboardSummary["progress"]
...
+): DashboardSummary["progress"][number] {
+  const workflowId = progress.workflowId || progress.runId;
+  if (!workflowId) {
+    return progress;
+  }
+  const workflow = workflowRunStatus.get(workflowId);
+  if (!workflow) {
+    return progress;
+  }
+  const status = pipelineProgressStatusForWorkflow(workflow.status);
```
ebarti, line 4535: PR Reviewer - code-reviewer

Severity: High
Gate: Fail
Evidence: progressWithWorkflowStatus (apps/api/src/read-model.ts:4523-4549) looks up workflow_run_projections by `progress.workflowId || progress.runId` and applies the terminal status with no recency guard. `discover_workflow_id()` (discovery/workflow.py:57) returns a STABLE `discover-{tenant}`; the fold `_project_workflow_run` is first-terminal-wins and never regresses to in_progress on a later WorkflowStarted; job_events is append-only (no pruning, verified) and the projection is upsert-only. So once any discover run terminates, the discover-local row is permanently stuck at that terminal status.
Reproduction: Reasoned from code. The PR's server.test.ts only covers the single-terminal / no-rerun case.
Finding: After the first discover run reaches ANY terminal state (the normal WorkflowCompleted -> succeeded, or a reconciler WorkflowTerminated, which is the exact condition this PR targets), every later discover run is masked on the dashboard: the live StageStarted running progress is overwritten with the prior run's terminal status, message and finishedAt. A second discover run shows succeeded / 100 percent the instant it starts (or failed with the stale reconciled_not_found message if the prior run was terminated) while it is actually running. This is the inverse of the bug the PR fixes and a truthfulness regression on the primary recurring flow. Apply re-runs of the same job (reused apply_workflow_id) share the exposure.
Required fix: Only apply the override for the run that produced this progress. Simplest guard: skip the override when workflow.finishedAt is earlier than progress.updatedAt (the StageStarted occurredAt) so a newer run is not masked. Add a regression test for start -> terminal -> new start on a reused workflow_id (discover-local) asserting the live run still reads as running.

apps/api/src/projections.ts
```diff
@@ -649,6 +656,9 @@ export function ensureProjectionTables(db: SqliteDatabase): boolean {
   schemaChanged =
     ensureProjectionColumn(db, "dashboard_projections", "outcome_conversion_json", "TEXT NOT NULL DEFAULT '{}'") ||
     schemaChanged;
+  for (const [columnName, definition] of WORKFLOW_RUN_PROJECTION_COLUMNS) {
+    schemaChanged = ensureProjectionColumn(db, "workflow_run_projections", columnName, definition) || schemaChanged;
```
ebarti, line 660: PR Reviewer - code-reviewer

Severity: Medium
Gate: Pass
Evidence: ensureProjectionColumn (apps/api/src/projections.ts:665-679) and _ensure_column (workers/automation/src/jobhunter/infrastructure/projections/sqlite_projection_store.py:354-367) both do check-then-ALTER (PRAGMA table_info, then ALTER TABLE ADD COLUMN) on workflow_run_projections with no catch for a duplicate-column error. Under `pnpm dev`, the TS API (first dashboard poll -> refreshProjections -> ensureProjectionTables) and the Python worker start concurrently against the same SQLite file.
Reproduction: Reasoned from code (cross-process TOCTOU on one file).
Finding: On the first upgrade of a legacy DB (the population this PR targets) both processes can read table_info before either ALTERs, then both ALTER the same missing column; the loser fails with a duplicate-column error. On the TS side this surfaces as a 500 from the dashboard/health read; on the worker side the finalize refresh activity fails and Temporal retries. It self-heals on the next poll/retry and follows the codebase's existing check-then-ALTER pattern, but the window is hit precisely for migrating users, and this PR is the first to make both processes ALTER this table.
Required fix: Make the add idempotent under races: catch the duplicate-column error (or re-check table_info after acquiring the write lock) in ensureProjectionColumn and _ensure_column so a concurrent add is a no-op rather than an unhandled exception.

ebarti (commented):
Stack review (PR #250, the base of the #250 -> #252 stack); only changes introduced by #250 vs main were reviewed. Two findings: one High correctness/truthfulness regression in the new dashboard terminal-state override, and one Medium concurrency risk in the new schema migration. Details inline.

apps/api/src/read-model.ts
ebarti: Fixed in 8fe011ae on `reopen/pr250-workflow-recovery` (PR #256, the reopened successor of this merged-then-reverted PR). `progressWithWorkflowStatus` now skips the terminal override when the projection row's `finished_at` predates the latest stage activity, so a live re-run of a reused workflow id (`discover-local`) keeps showing its real progress. Regression test added: start → terminal → re-start keeps `running` (apps/api/test/server.test.ts). The underlying fold-level run-identity gap gets its own fix on PR #252's branch.

apps/api/src/projections.ts
ebarti: Fixed in 9a0db450 on `reopen/pr250-workflow-recovery` (PR #256). Both `ensureProjectionColumn` (TS) and `_ensure_column` (Python) now re-check the schema when ALTER throws and treat duplicate-column as an upgrade the other process already applied, instead of failing initialization. Deterministic TOCTOU regressions added on both sides (projections.test.ts, test_workflow_runs_projection_from_events.py).

PR #252, opened 2026-07-04
fix(reliability): discover fault isolation, truthful failures, honest read models

ebarti:
# fix(reliability): discover fault isolation, truthful failures, honest read models

## Incident (2026-07-04, workflow `discover-local`, run `019f2d0c-7441-7641-ae61-b7f6f5b1127d`)

A discover run consumed 1h55m, froze at "67% — Smart extract complete", and failed with the literally useless error `discover:enrichment failed: failed`. The dashboard kept reporting `running`/67% after the workflow was closed, and `/v1/workflow-runs` reported a stale `terminated (reconciled_not_found)` from a *previous* run. This PR is the full root-cause fix package.

## Root cause chain (7 compounding defects, all evidence-verified)

**RC-1 — Trigger (environment).** Playwright's `chromium_headless_shell-1208` had been garbage-collected from the shared `~/Library/Caches/ms-playwright` by a `playwright install` run from another worktree with a newer Playwright pin. Every `p.chromium.launch(headless=True)` failed instantly with `BrowserType.launch: Executable doesn't exist`. Reproduced offline against a DB snapshot; also the cause of the PDF-generation failures in the post-restart worker log. The LinkedIn enrichment batch was immune — the authenticated resolver launches real Chrome via `config.get_chrome_path()` (`linkedin_apply_resolver.py:158`), which is why the run *looked* healthy for 31 minutes. Nothing validated browsers at worker startup or in `doctor`.

**RC-2 — Blast radius.** The Indeed site batch crashed on browser launch at 13:35:40 (event-free — before its first per-job event). No per-job or per-site fault containment existed: the exception propagated via `future.result()` → `_run_discovery_enrichment_until_idle`'s catch → entire enrichment stage `failed` → workflow failed. Meanwhile the LinkedIn thread successfully enriched ~40 jobs; that work was fine in the DB but the run reported total failure.

**RC-3 — 31-minute error latency.** `ThreadPoolExecutor.__exit__` (`shutdown(wait=True)`) held the Indeed exception from 13:35:40 until the LinkedIn thread finished at 14:06:43.

**RC-4 — Retry futility.** Temporal retried the deterministic environment error 3× (`maximumAttempts=3`; attempt 3 started 14:07:53.215, failed 14:08:43 `RETRY_STATE_MAXIMUM_ATTEMPTS_REACHED`). Each retry re-drove a ~37s authenticated LinkedIn browser pass against `linkedin.com/jobs/view/4436755805` (no external apply URL exists), burning its bounded 3-attempt recovery budget — the attempt timestamps in `job_enrichments.attempts_json` align to the activity attempt starts to within 150ms.

**RC-5 — Observability collapse (`failed: failed`).** `discovery/activities.py` fabricated `activity_result = {"status": status, "errors": {}}`, discarding the `error_class`/`error_message` the runner had captured; `_raise_on_failure` then rendered `errors or result.get("error") or result.get("status")` → the string `failed`. The only faithful traceback went to `.dev/logs/worker.log`, which `scripts/dev` truncates on restart (`: > "$log"`). The Temporal enrichment path also bypasses `_run_stage_observed`, so no durable StageFailed pipeline event was written either. Net: a 2-line environment fix presented as an inscrutable pipeline failure.

**RC-6 — Progress read model lied.** The Temporal enrichment path emits zero discover progress events (the legacy `finish_discovery` emitted "Detail enrichment"/"Preparation" steps; `run_discovery_enrichment_stage` emitted nothing), so progress froze at 4/6 = 67% forever, and neither activity nor workflow failure finalized it.

**RC-7 — Workflow projection lied.** `workflow_run_projections` is keyed by `workflow_id`. An earlier run had been closed by the reconciler as `terminated/reconciled_not_found` (11:16:55). The new run's `WorkflowStarted` updated `started_at`/`temporal_run_id` but a terminal-state guard refused to clear `status`/`error*`/`finished_at`, and the new run's `WorkflowFailed` (14:08:43, present in `events_json`) never applied — a chimera row served as `terminated` by `/v1/workflow-runs`.

**Also fixed while in there:**
- Latent: source-family statuses `skipped_limit` / `skipped_quality` / `skipped_disabled` were not in `_SUCCESS_STATUSES` — a limit-skipped family would have failed the whole workflow the same way.
- Review finding: the decomposed `DiscoverWorkflow` failed the entire workflow *before* enrichment/preparation when ANY source family failed; legacy semantics (tolerate partial source failure, always run enrichment+preparation, fail only when all sources fail) are restored.

## Changes

**Python worker**
- `enrichment/detail.py`: per-job exception containment in `scrape_site_batch` (job marked failed `ENRICH_INTERNAL_ERROR`, batch continues); per-site containment in `_run_detail_scraper` (both parallel and sequential) with `site_errors` reporting; systemic failure (all sites failed, zero processed) raises `ConfigurationError` (non-retryable) for broken-environment signatures, `TransientNetworkError` otherwise; per-row containment in the LinkedIn retry scan; authenticated LinkedIn rescan gated to the first pass of a stage run.
- `pipeline/runner.py`: `_run_enrich` returns real stats + `site_errors`; `_run_discovery_enrichment_until_idle` continues draining through partial passes, accumulates `site_errors`, and captures `error_class`/`error_message`/`error_code`/`retryable`/`error_traceback` on crashes; `run_discovery_enrichment_stage` emits durable discover progress events (StageStarted / StageCompleted / StageFailed with terminal `status: failed` payload).
- `discovery/activities.py`: failures raise `ApplicationError` with the real `class: message`, `type` = domain error code, `non_retryable` from classification, and structured details (passes, pending, siteErrors, traceback tail). The literal `failed: failed` is impossible. `skipped_*` statuses count as success.
- `discovery/workflow.py`: partial-source semantics restored — enrichment + preparation always run; workflow fails only when ALL source families failed (after enrichment/preparation), or on real enrichment failure (with the real error surfaced); enrichment failure no longer skips preparation fan-out.
- NEW `infrastructure/preflight.py` + worker startup gate + `doctor` check: missing Playwright Chromium now fails the worker at startup in seconds with the exact fix command (escape hatch `JOBHUNTER_SKIP_BROWSER_PREFLIGHT=1`), instead of 2 hours into a run.

**TypeScript API**
- `workflow_run_projections`: a new execution's `WorkflowStarted` supersedes a stale terminal state (clears error/finished_at, resets status); the current execution's terminal event applies; duplicate `WorkflowStarted` stays idempotent; refresh-from-events converges.
- Discover progress surface: terminal workflow state finalizes progress (never `running` for a closed workflow).

**Ops**
- `scripts/dev`: process logs rotate to `<name>.log.1` on restart instead of truncating — post-mortems survive.
- Environment repaired immediately during the incident response: `chromium_headless_shell-1208` reinstalled alongside 1217; headless launch verified from the affected venv.

## Docs
- README: per-venv `playwright install chromium` requirement, shared-cache GC gotcha across worktrees (`PLAYWRIGHT_SKIP_BROWSER_GC=1`), doctor/worker preflight.
- `docs/local-reliability-qa.md`: 7 new regression-matrix entries (partial-source continuation; site/job fault isolation; no `failed: failed`; progress advance + terminal finalization; projection precedence; preflight; log rotation).
- Pipeline architecture doc: failure-behavior section updated to the new semantics.

## Validation

- `uv --project workers/automation run --extra dev pytest -q` → **1796 passed** (includes the new incident regressions: `test_discover_reliability.py`, extended `test_workflow_discovery.py`, extended `test_workflow_runs_projection_from_events.py`, `test_preflight.py`, `test_doctor_playwright.py`; the literal invariant `"failed: failed" not in str(err)` is asserted).
- `uv --project workers/automation run --extra dev ruff check .` → all checks passed.
- `pnpm api:check` → clean; `pnpm api:test` → **334 passed** (incl. "supersedes a stale terminal workflow-run projection when a reused workflow id restarts and fails", duplicate-start idempotence, within-run reconciler-race backstop preserved, "finalizes discover progress from the re-folded workflow verdict").
- Offline incident reproduction: the crash was reproduced against a snapshot of the production DB (`ValueError`/browser-launch path) before fixing; the environment itself was repaired live (`chromium_headless_shell-1208` reinstalled alongside 1217, headless launch verified).
- Web app untouched (no `apps/web` changes) — web suites not run.
- pr-reviewer and qa gate results recorded in PR comments.

### Restack validation (2026-07-05)

- `git diff --check origin/reopen/pr250-workflow-recovery..HEAD`
- `corepack pnpm api:check`
- `corepack pnpm api:test` -> 337 passed
- `uv --project workers/automation run --extra dev ruff check .` -> all checks passed
- `uv --project workers/automation run --extra dev pytest -q workers/automation/tests/test_discover_reliability.py workers/automation/tests/test_workflow_discovery.py workers/automation/tests/test_workflow_runs_projection_from_events.py workers/automation/tests/test_preflight.py workers/automation/tests/test_doctor_playwright.py` -> 77 passed


apps/api/src/read-model.ts
```diff
@@ -4910,12 +4914,164 @@ function columnNames(db: SqliteDatabase, tableName: string): Set<string> {
...
+  let errorMessage: string | null = null;
+  let retryable = false;
+  let startedAt: string | null = null;
+  let finishedAt: string | null = null;
+  let durationMs: number | null = null;
+
+  for (const event of events) {
+    const payload = event.payload;
+    if (event.eventType === "WorkflowStarted") {
+      if (phase !== "in_progress") {
```
ebarti: PR Reviewer - pr-reviewer

Severity: Medium
Gate: Pass
Evidence: This re-fold reopens the row to `in_progress` on ANY `WorkflowStarted` after a terminal (`phase !== "in_progress"`), with no run-id check. The Python writer fold (`projection_builder.py::_starts_new_execution`) only reopens for a genuinely NEW execution (different `temporalRunId`, or absent run ids a later `occurredAt`) and otherwise preserves the terminal — asserted by `test_within_run_duplicate_started_does_not_reopen_terminal` (Python suite passes 40/40). I ran the exact 3-event sequence from that test (Started/Terminated/Started, all run id `temporal-same`) through a faithful port of this algorithm: TS returns `{status:"in_progress", errorCode:null, finishedAt:null}` while Python asserts `terminated / reconciled_not_found`. Because `applyWorkflowLifecycleFold` overrides the stored row on every read path (`listWorkflowRuns`, `getWorkflowRunDetail`, `loadPipelineWorkflowRunStatus`), the two folds disagree and, on this input, the API would serve a closed run as `in_progress`. No TS test covers same-run-after-terminal (all TS reopen tests use a differing run id, where reopening is correct), so the gap is masked.
Reproduction: Reproduced with a standalone node port of foldWorkflowRunEvents on the test_within_run_duplicate_started_does_not_reopen_terminal event sequence (TS=in_progress, Python-asserted=terminated).

Severity is Medium (not High) because the divergent input is not reachable through the current emit paths: `emit_workflow_started` is awaited before the workflow can reach its outcome, so a same-run start always has a lower event_id than that run's terminal. It is a defensive-invariant divergence in a truthfulness-critical backstop the Python writer explicitly guards — worth closing so a future change (redelivered/reordered starts, a replay/recovery path) can't silently un-terminalize a closed run.

Required fix: Make the reopen run-scoped, mirroring `_starts_new_execution` — only reset to `in_progress` when the start's `temporalRunId` differs from the folded run id, or (run id missing) its `occurredAt` is after the folded `finishedAt`; stay idempotent otherwise. Add a TS test mirroring `test_within_run_duplicate_started_does_not_reopen_terminal`.

workers/automation/src/jobhunter/discovery/activities.py
```diff
@@ -260,9 +271,48 @@ async def discovery_preparation_fanout_activity(
...
+    """
+    error_class = result.get("error_class")
+    error_message = result.get("error_message")
+    status = str(result.get("status") or "failed")
+    if error_class and error_message:
+        message = f"{stage} failed: {error_class}: {error_message}"
+    elif error_message:
+        message = f"{stage} failed: {error_message}"
+    else:
+        message = f"{stage} failed: {status}"
```
ebarti, line 381: PR Reviewer - pr-reviewer

Severity: Low
Gate: Pass
Evidence: When a result has `status="failed"` and a falsy `error_message`, this else branch produces `f"{stage} failed: {status}"` = "discover:enrichment failed: failed", ignoring the available `error_class`. `_run_discovery_enrichment_until_idle` sets `error_message=str(exc)`, which is "" for a bare exception (e.g. `raise ValueError()`), so the exact "failed: failed" collapse the PR claims impossible is still reachable (milder than the original — the details dict still carries errorClass). Not covered: `test_stage_failure_error_falls_back_to_status_without_class` uses status="stuck: ...", not "failed".
Reproduction: Reasoned from code; no test exercises status="failed" with an empty error_message.

Required fix: Incorporate `error_class` in the else branch (e.g. `f"{stage} failed: {error_class or status}"`), and/or guarantee `_run_discovery_enrichment_until_idle` records a non-empty error_message (fall back to error_class / repr(exc)). Add a regression for the empty-message exception to keep the "failed: failed impossible" invariant airtight.

docs/local-reliability-qa.md
```diff
@@ -132,6 +132,13 @@ VITE_JOBHUNTER_API_BASE_URL=http://127.0.0.1:8766 pnpm web:dev -- --port 5173
...
+| A single job's enrichment crash fails the entire site batch (or the activity) instead of marking only that job failed with `ENRICH_INTERNAL_ERROR` while the rest of the batch continues | `workers/automation/tests/test_discover_reliability.py` |
+| A Temporal activity failure is surfaced as the placeholder string `failed: failed` instead of the real error class and message; the propagated `ApplicationError` must carry the actual error type, message, and details (the literal `failed: failed` must never appear) | `workers/automation/tests/test_p1b_error_inversion.py`; `workers/automation/tests/test_discover_reliability.py` |
+| Discover progress stalls at a stale running percentage (e.g. a frozen `running 67%`) instead of advancing through the "Detail enrichment" and "Preparation" steps on the Temporal path, or fails to show a terminal `failed` status when the workflow fails | `workers/automation/tests/test_workflow_discovery.py`; `workers/automation/tests/test_discover_reliability.py`; `apps/api/test/server.test.ts` |
```
ebarti, line 163: PR Reviewer - pr-reviewer

Severity: Low
Gate: Pass
Evidence: This row (and the PR description) states discover progress advances through the "Detail enrichment" AND "Preparation" steps on the Temporal path. Only "Detail enrichment" emits progress (`run_discovery_enrichment_stage`). `DiscoveryPreparationFanoutInput` has no progress fields and `_start_preparation_workflows` (discovery/workflow.py:263) passes none, so the preparation fanout emits no progress event. With `progress_total = len(families) + 2`, the bar tops out at (len+1)/(len+2) (~83% for 4 families) until the workflow terminalizes. The 67% freeze IS fixed (enrichment now advances + terminal finalization), but the "Preparation" step claim is inaccurate.
Reproduction: Reasoned from code (discovery/activities.py:74-85, discovery/workflow.py:263-283); no progress test asserts a "Preparation" step.

Required fix: Either wire progress_completed/progress_total into the preparation fanout so it emits a "Preparation" step (matching legacy finish_discovery and this doc row), or correct the doc/PR to state the final step is closed by terminal finalization rather than a Preparation progress event.

ebarti:
Review findings addressed in 1764fafa:

1. **Medium — read-side re-fold reopened on any WorkflowStarted**: `foldWorkflowRunEvents` now mirrors the Python writer's `_starts_new_execution` guard (run-id difference authoritative, wall-clock fallback when ids absent). New regression: *ignores a late duplicate start for an already-terminalized execution*.
2. **Low — `failed: failed` corner in `_stage_failure_error`**: added class-only and bare-`failed`-status fallbacks; two new tests assert the literal string is unreachable from every branch.
3. **Low — docs claimed Preparation progress that didn't exist**: implemented instead of downgrading the docs — `discovery_preparation_fanout_activity` now emits the legacy-shaped Preparation StageStarted/StageCompleted/StageFailed progress events (without this, successful runs would freeze at the penultimate step instead of reaching 100%). Workflow threads `progress_completed/total` into the fanout input.

Verification after fixes: pytest **1802 passed**, ruff clean, `pnpm api:check` clean, `pnpm api:test` **335 passed**.

ebarti:
## QA gate report (closed with direct evidence; the qa agent stalled mid-run and was stopped)

**Scenario 1 — read-model truthfulness against the REAL incident DB** (API from this branch on an isolated port against a `sqlite3 .backup` snapshot of production):
- `/v1/workflow-runs` list + detail for `discover-local`: **status `failed`, finishedAt `2026-07-04T14:08:43`** — the stale `terminated / reconciled_not_found` chimera is gone with zero data repair.
- `/v1/dashboard/summary` discover card initially still showed the frozen `running 67%` (progress events carry only `discovery:<family>:<hex>` run ids, never the owning workflow id) → **found + fixed in 4986c9a1** (deterministic `discover-<tenant>` fallback in the workflow-status linkage). Re-probe against the same snapshot: `failed / "Workflow failed: Activity task failed" / 14:08:43`, last-known step (Smart extract, 4/6, 67%) honestly preserved.

**Scenario 2 — preflight fail-fast** (simulated GC'd cache via `PLAYWRIGHT_BROWSERS_PATH=<empty>`, `TEMPORAL_ADDRESS` pointed at an unused port):
- `jobhunter worker`: exits **code 1 in ~1s**, before any Temporal connection, printing the install command, the multi-worktree GC warning, and the `JOBHUNTER_SKIP_BROWSER_PREFLIGHT=1` escape hatch.
- `jobhunter doctor`: `playwright chromium (scraping + PDF)` reports **MISSING** with the empty cache and **OK** with the real cache.

**Scenario 3 — dev log rotation**: `rotate_log` preserves the prior log's content as `<name>.log.1`, leaves empty logs unchurned, and both `start_one`/`start_run_one` call it.

**Scenario 4 — regression suites**: pytest **1802 passed**, ruff clean, `pnpm api:check` clean, api vitest **336 passed** (all after the review + QA fixes).

Gate: PASS

ebarti:
## Accepted residuals / follow-up ledger (none merge-blocking)

1. **Stamp `workflowId` on Temporal-path pipeline events at the writer** — `_PIPELINE_RUN_CONTEXT` (runner.py:54) is never assigned anywhere in the repo, so no Temporal-path event carries a workflow id; the read model now compensates with the deterministic `discover-<tenant>` fallback. Root fix: set the thread-local from `activity.info().workflow_id` inside each discovery activity's blocking callable. (progress-tracer finding)
2. **Reconciler run-pinning** — `_reconcile_one_workflow_run` describes by workflow_id without pinning the run id (cli.py:1605); latent conflation risk when a new run starts mid-reconcile. Impact bounded by the run-scoped fold + read-side re-fold. (ts-projection-tracer finding)
3. **Stored-chimera self-heal on no-op refresh** — `_rebuild_workflow_runs` re-folds correctly on the next `Workflow*` event; an existing chimera row is not rewritten by a plain refresh (ce7ab142's count check only recovers *missing* rows). Reads are already correct via the API re-fold; optional: detect terminal rows whose latest folded run id ≠ row.temporal_run_id.
4. **Fold-query index** — if `loadWorkflowRunLifecycleFolds` ever needs optimizing (8 ms @ 12 k rows today), add `AND stage = 'workflow'` to BOTH folds (never one side only, per reviewer note).
5. **Cosmetic**: `scrape_site_batch` stats can double-count a job if a DB write raises after the ok/partial increment — logging-only, no control-flow impact.

apps/api/src/read-model.ts
```diff
@@ -4910,12 +4920,209 @@ function columnNames(db: SqliteDatabase, tableName: string): Set<string> {
...
+      const eventRunId = workflowLifecycleRunId(payload);
+      const eventStartedAt = nullableString(payload.startedAt) ?? event.occurredAt;
+      const reopens =
+        phase === "none" ||
+        (phase === "terminal" &&
+          startsNewWorkflowExecution({
+            foldedRunId: runId,
+            foldedFinishedAt: finishedAt,
+            eventRunId,
+            eventOccurredAt: eventStartedAt,
```
ebarti: PR Reviewer - pr-reviewer

Severity: Low
Gate: Pass
Evidence: Line 5064 passes `eventOccurredAt: eventStartedAt` into `startsNewWorkflowExecution`, and line 5056 sets `eventStartedAt = nullableString(payload.startedAt) ?? event.occurredAt`. The Python mirror it is meant to match (`projection_builder.py:1889`) passes the raw `event_occurred_at=occurred_at` — the event's occurredAt, never `payload.startedAt`. A `WorkflowStarted` always carries a distinct `payload.startedAt` (workflow.now() at start, finalize.py:227) vs the row's `occurred_at` (DB write time), so in the run-id-absent wall-clock fallback the two folds compare different timestamps against `finished_at`.
Reproduction: Reasoned from code. The new "ignores a late duplicate start" test exercises only the run-id-authoritative branch (same temporal run id); the wall-clock branch's timestamp source is untested.

The guard is an exact mirror of `_starts_new_execution` on the run-id-authoritative branch and on the terminal-preservation / new-execution-reopen structure — I re-verified both against the incident and chimera sequences and they agree. The one residual gap is the wall-clock fallback (reached when a run id is absent on either side, e.g. a reconciler `terminated/reconciled_not_found` terminal that carries no temporal_run_id, followed by a new run's `WorkflowStarted`): TS compares `payload.startedAt` while Python compares `occurred_at`. This is not reachable as a divergent OUTCOME today — a new run's `startedAt` and `occurredAt` differ by milliseconds and both fall long after any stale terminal's `finished_at`, so `started > finished` resolves the same on both sides. It would only flip if `finished_at` landed in the sub-second gap between the new run's workflow.now() start and its event write, which does not happen in practice. Flagging it because the stated goal of this commit was to mirror `_starts_new_execution` exactly, and this is the single spot that does not.

Required fix: feed the raw event occurredAt into the guard — `eventOccurredAt: event.occurredAt` — and keep `eventStartedAt` only for the stored `startedAt` assignment (matching Python's `started_at = payload.get("startedAt") or occurred_at` while `event_occurred_at=occurred_at`). Optionally add a run-id-absent wall-clock TS case to lock the parity.

apps/api/src/read-model.ts
ebarti: PR Reviewer - pr-reviewer

Verified addressed in 1764fafa. `foldWorkflowRunEvents` now tracks `runId` and gates the post-terminal reopen through `startsNewWorkflowExecution`, which mirrors Python's `_starts_new_execution` (run-id difference authoritative). The new test "ignores a late duplicate start for an already-terminalized execution" reproduces the exact WorkflowStarted -> WorkflowFailed -> duplicate-WorkflowStarted (same run id) sequence my node-port flagged and asserts the failed verdict survives. One residual Low parity nit on the wall-clock fallback filed separately (#discussion_r3523914366); not merge-blocking. Resolved.

workers/automation/src/jobhunter/discovery/activities.py
ebarti: PR Reviewer - pr-reviewer

Verified addressed in 1764fafa. `_stage_failure_error` gained an `elif error_class` branch and a bare-"failed"-status branch ("failed with no recorded error detail"), so the "failed: failed" collapse is now unreachable. Covered by test_stage_failure_error_with_class_only_never_collapses and test_stage_failure_error_bare_failed_status_never_collapses (both assert "failed: failed" not in message). Resolved.

docs/local-reliability-qa.md
ebarti: PR Reviewer - pr-reviewer

Verified addressed in 1764fafa (implemented rather than documented-away, the stronger fix). DiscoveryPreparationFanoutInput gained progress_completed/progress_total; the workflow passes len(families)+1 / progress_total; the activity emits legacy-shaped StageStarted/StageCompleted/StageFailed via _record_preparation_progress. Arithmetic verified contiguous: families 0->len, enrichment len->len+1, preparation len+1->len+2==total (100%). Covered by test_fanout_activity_emits_preparation_progress and the failure-path variant. Resolved.

ebarti:
Re-review Low addressed in 9c2e1bfc: the TS reopen fallback now passes the event's `occurredAt` (record time) into `startsNewWorkflowExecution`, matching Python's `_starts_new_execution` exactly. `pnpm api:check` clean, api vitest **336 passed**. All review findings across both passes are now resolved; QA gate passed earlier with direct evidence against the incident DB.

apps/api/src/read-model.ts
ebarti: PR Reviewer - pr-reviewer

Verified addressed in 9c2e1bfc. The wall-clock fallback now passes `eventOccurredAt: event.occurredAt` (read-model.ts:5067), matching Python's `event_occurred_at=occurred_at` in `_starts_new_execution`. `eventStartedAt` (= payload.startedAt ?? occurredAt) is correctly retained only for the stored `startedAt` assignment (line 5076), which mirrors Python's `started_at = payload.get("startedAt") or occurred_at`. Both the reopen comparison and the stored start time are now exact parity with the Python fold; api:check clean on my end. Resolved — no residual parity gap.

workers/automation/src/jobhunter/discovery/workflow.py
```diff
@@ -189,36 +189,87 @@ async def _execute(self, payload: DiscoverWorkflowInput) -> DiscoverWorkflowResu
...
 
-        if failures:
+        # Legacy semantics: per-source failures are tolerated. Enrichment and
+        # preparation ALWAYS run so the healthy sources' jobs still flow through
+        # the pipeline; the workflow only fails as a source failure when EVERY
+        # family failed.
+        enrichment_failure: str | None = None
+        enrichment_error_code: str | None = None
+        try:
+            await workflow.execute_activity(
```
ebarti: [P1] Partial enrichment site errors are dropped here. `discovery_enrichment_activity` can return `status="partial"` with `site_errors`, but this await discards the output. Because `run_discovery_enrichment_stage` treats partial as a generic StageCompleted event (`runner.py:1610`) and preparation later terminalizes the workflow, a site-batch crash can be hidden behind a successful Discover run. Please either carry the partial result into the workflow outcome/read model, or emit a durable partial/warning progress payload with `siteErrors` so the UI/API remains truthful.

workers/automation/src/jobhunter/enrichment/detail.py
```diff
@@ -645,301 +685,308 @@ def scrape_site_batch(
...
+                        )
 
-                conn.commit()
+                    conn.commit()
+                except TransientNetworkError:
+                    raise
+                except Exception as exc:
+                    log.exception("Enrichment job failed: %s", url)
+                    stats["error"] += 1
+                    _record_enrich_job_failure(conn, url, exc)
```
ebarti, line 1027: [P2] This only records the unexpected job crash in `job_stage_states`/`job_events`; it never saves a failed `JobEnrichment` aggregate. I reproduced the fixture shape locally: the bad job ended with `job_stage_states.enrich = failed / ENRICH_INTERNAL_ERROR`, but no `job_enrichments` row. `_ENRICHMENT_PENDING` is blocked by the failed stage state, so this is not an immediate retry loop, but the canonical enrichment attempt history/list_failed/audit surface loses the failed attempt. Please persist a failed aggregate attempt when possible (or document why stage state is intentionally the source of truth here).

ebarti (commented):
Two remaining review findings from the current pass.

workers/automation/src/jobhunter/discovery/workflow.py
ebarti: Fixed in 90c77128. The workflow now preserves `DiscoveryEnrichmentActivityOutput.status` and `site_errors` in `DiscoverWorkflowResult`; `run_discovery_enrichment_stage` emits partial enrichment as a warning-level `StageCompleted` payload with `progress.status = "partial"` and top-level `siteErrors`; and the API read-model preserves partial discover warnings after successful workflow terminalization instead of squashing them to plain `Workflow succeeded`. Covered by `test_stage_progress_emits_partial_site_errors`, `test_discover_workflow_preserves_partial_enrichment_site_errors`, and `preserves partial discover warnings after the workflow succeeds`. Verified with ruff, API typecheck, full API tests, and the worker reliability slice.

workers/automation/src/jobhunter/enrichment/detail.py
ebarti: Fixed in 90c77128. `_record_enrich_job_failure` now persists a canonical failed `JobEnrichment` attempt with `ENRICH_INTERNAL_ERROR` before writing the stage state/event, while leaving an already-enriched aggregate intact so accepted material is not destroyed by an audit-only crash. Covered by `test_scrape_site_batch_isolates_single_job_failure`, which now asserts the bad job has `job_enrichments.current_status = failed` and a failed attempt carrying the internal error. Verified with ruff and the worker reliability slice.

ebarti:
Pushed two commits from the stack-review follow-up:

- 5f3fee78 merges the updated base `reopen/pr250-workflow-recovery` in (carries the two PR #250 review fixes: 8fe011ae stale-terminal recency guard for dashboard progress, 9a0db450 cross-process column-upgrade race).
- 3525c89f adds the **terminal-direction run-id guard to both workflow-run folds**: a late `WorkflowTerminated`/`WorkflowFailed` carrying a different temporal run id than the currently folded execution (reconciler describe racing a new start — the direction the accepted residual #2 "bounded by the run-scoped fold" note did not cover) is skipped; the live run's own terminal still applies. Writing the TS regression exposed a real fold-parity gap: the TS re-fold ignored a `WorkflowStarted` with a new run id while the fold was still open (previous execution died without a recorded terminal), where the Python fold adopts it — the TS fold now adopts it too; same-run duplicate starts keep their existing idempotent handling.

Regressions: `test_late_terminal_from_superseded_run_does_not_clobber_live_run` (Python fold) and "ignores a late terminal from a superseded execution while a newer run is live" (TS re-fold), both covering Started(A) → Started(B) → late Terminated(A) → B's own terminal.

Verified: 177 API vitest (server + projections, under Node 26 matching this worktree's better-sqlite3 ABI), `tsc --noEmit`, 49 + 34 pytest (workflow-runs fold, discover-reliability, pipeline-observability, workflow-discovery), ruff clean.

PR #257, opened 2026-07-05
docs(plans): [R0] defer naming and PyPI publishing to the pre-publication rename train

ebarti:
## What

- §1 locked decisions: the naming bullet now records the owner decision of 2026-07-05 — the product's final name is JobCtl, delivered by an owner-executed rename train (repo name, product strings, console script, PyPI distribution name `jobctl`, docs sweep) that lands after every W-item and before the visibility flip. Until then nothing is renamed and nothing is published to PyPI.
- §4 W2.1 (PyPI distribution rename) is removed from the program and replaced with a tombstone; the section number is retained so cross-references stay stable.
- §0.1 gate table, sequencing diagram, and ordering notes drop W2.1; §0.5 checkpoint 1 is tombstoned with numbering preserved (§0.5.2–.4 references unchanged).
- W0.3 item 7 and W0.5 now point their retirement/gating language at the rename train instead of W2.1.
- §5 release-gate checklist: the W2.1 line becomes an owner-side "rename train executed" line (rename, `jobctl` distribution name, publish.yml re-enable gated on release-check, structural-check retirement).
- `.github/workflows/publish.yml`: header comment updated to point at the rename train instead of W2.1 (comment-only; trigger stays `workflow_dispatch`).

## Why

Owner decision: there is no reason to publish to PyPI under the interim name — the first real publish happens after the JobHunter → JobCtl rename train, which runs last (before the visibility flip) because a rename invalidates the path/symbol anchors the spec navigates by. Removing W2.1 keeps the remaining implementation program (W1.1–W1.8, W2.2–W2.6) free of naming/publishing concerns and leaves the W0.5/W0.3 accidental-publish guards in force until the train lands.

## Validation

- `python3 scripts/release_check.py` → exit 0, zero findings (the three prompt.py tripwire warnings are the expected W1-gated ones).
- `git diff --check` → clean.
- `grep -n "W2\.1" docs/plans/2026-07-03-oss-release-remediation-spec.md` → only the tombstones and historical pointers remain (lines 155, 180, 369, 385, 950).
- Docs-only + workflow comment; no code paths changed, so no test suites apply.

## Deviations

None.

## Deferred

The rename train itself (owner-side, tracked by the §5 checklist line).

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,4 +1,6 @@
-# W2.1 restores tag publishing after the PyPI distribution rename.
+# Tag publishing returns with the pre-publication rename train
+# (docs/plans/2026-07-03-oss-release-remediation-spec.md §1 "Naming").
+# Do not re-enable before it; the release_check structural guard enforces this.
 name: Publish to PyPI
 
 on:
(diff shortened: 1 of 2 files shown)


ebarti:
Superseded by the merged R0 rename train in #349. That branch includes the naming/PyPI deferral decision and then completes the rename, so merging #257 now would reintroduce stale pre-rename release-plan text.

PR #268, opened 2026-07-05
fix(api): reject cross-site fetch metadata

ebarti:
## What

- Adds a `Sec-Fetch-Site` unsafe-method gate after the existing local Host/Origin/Referer checks.
- Keeps non-browser local clients working when fetch metadata is absent.
- Documents the loopback trust boundary in `SECURITY.md` and `docs/local-ts-api.md`.
- Adds API regression coverage for cross-site, same-origin, none, same-site, and headerless local-client cases.

## Why

W2.3 closes the browser fetch-metadata gap without changing the local-first API model. Cross-site browser requests should fail even if they try to spoof other trusted-looking mutation metadata, while local scripts and seeds remain supported.

## Validation

- `corepack pnpm api:check`
- `corepack pnpm api:test` (20 files / 346 tests)
- `corepack pnpm qa:test` (3 tests)
- `corepack pnpm check`
- `corepack pnpm test` (API 346 tests, web build, Python 1806 tests)
- `corepack pnpm docs:build` (4545 references / 230 emitted files)
- `python3 scripts/release_check.py` (passes with the three expected W0 prompt warnings called out in the remediation prompt)
- `git diff --check`
- Review: W2.3 reviewer Gate: PASS
- QA: W2.3 QA Gate: PASS; one Low docs gap was addressed before this PR was opened

## Deviations

None. W2.1 rename/publish work is explicitly out of scope, so this PR keeps all `JobHunter`/`jobhunter` names unchanged.

## Deferred

None for W2.3.

### DIFF
--- a/apps/api/test/server.test.ts
+++ b/apps/api/test/server.test.ts
@@ -578,6 +578,88 @@ describe("local TypeScript API", () => {
     await app.close();
   });
 
+  it.each([
+    {
+      name: "cross-site browser metadata",
+      jobUrl: "https://example.com/jobs/ready",
+      action: "mark-applied",
+      payload: { reason: "cross-site metadata" },
+      headers: { "sec-fetch-site": "cross-site" },
+      expectedStatus: 403,
+      expectedBody: { ok: false, error: "cross_site_request" },
+    },
+    {
+      name: "same-site browser metadata without loopback origin",
+      jobUrl: "https://example.com/jobs/ready",
+      action: "mark-applied",
+      payload: { reason: "same-site metadata" },
+      headers: { "sec-fetch-site": "same-site" },
+      expectedStatus: 403,
+      expectedBody: { ok: false, error: "cross_site_request" },
+    },
+    {
+      name: "loopback same-site browser metadata",
+      jobUrl: "https://example.com/jobs/ready",
+      action: "mark-applied",
+      payload: { reason: "loopback same-site metadata" },
+      headers: { origin: "http://127.0.0.1:5173", "sec-fetch-site": "same-site" },
+      expectedStatus: 200,
+      expectedBody: { action: "mark_applied", stage: { state: "succeeded" } },
+    },
+    {
+      name: "same-origin browser metadata",
+      jobUrl: "https://example.com/jobs/ready",
+      action: "mark-applied",
+      payload: { reason: "same-origin browser" },
+      headers: { origin: "http://127.0.0.1:5173", "sec-fetch-site": "same-origin" },
+      expectedStatus: 200,
+      expectedBody: { action: "mark_applied", stage: { state: "succeeded" } },
+    },
+    {
+      name: "browser navigation metadata",
+      jobUrl: "https://example.com/jobs/blocked-tailor",
+      action: "mark-skipped",
+      payload: { reason: "browser navigation" },
+      headers: { "sec-fetch-site": "none" },
+      expectedStatus: 200,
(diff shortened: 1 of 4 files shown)


docs/local-ts-api.md
```diff
@@ -15,7 +15,11 @@ any request whose `Host` header is not a loopback host (`127.0.0.1`,
 local browser and CLI callers always send a loopback `Host`, so legitimate
 access is unaffected. A second guard rejects cross-site mutations: `DELETE`,
 `PATCH`, `POST`, and `PUT` requests whose `Origin` or `Referer` is present but
-not loopback fail with `403` `cross_site_request` before any handler runs.
+not loopback fail with `403` `cross_site_request` before any handler runs. When
+browser fetch metadata is present, unsafe mutations also reject
+`Sec-Fetch-Site: cross-site`; `same-site` is accepted only after the loopback
+`Origin`/`Referer` check has passed, so the default Vite web app on another
+loopback port still works while foreign browser origins do not.
```
ebarti, line 22: **[Low]** This overstates the protection on a security surface. The boundary the code actually enforces is same-site vs cross-site, not local-app vs foreign. `isTrustedMutationSource` and the `same-site` acceptance both treat *any* loopback Origin as trusted, so another local app on the same host but a different port (e.g. a page on `http://127.0.0.1:9000` calling the API on `:8766`) is `same-site` with a loopback Origin and is still accepted — it “works”, it is not blocked. Only a different loopback *label* (`http://localhost:PORT` -> `http://127.0.0.1:8766`, which is cross-site) or a non-loopback origin is rejected. Suggest rewording to “cross-site browser origins do not” and noting that same-host other-port pages remain accepted, to match the accurate “defense in depth, not authentication” framing already in SECURITY.md.

ebarti (commented):
CSRF hardening logic is correct and introduces no regression: cross-site is rejected in both trusted sets, none/same-origin are always accepted, a missing Sec-Fetch-Site header keeps non-browser clients (curl, Python worker, seeds) working, and all three web modes still pass (dev proxy -> same-origin; preview/direct -> loopback Origin makes same-site accepted). Two Low, in-diff notes below on doc accuracy and test realism; no blockers.

apps/api/test/server.test.ts
```diff
@@ -578,6 +578,88 @@ describe("local TypeScript API", () => {
     await app.close();
   });
 
+  it.each([
+    {
+      name: "cross-site browser metadata",
+      jobUrl: "https://example.com/jobs/ready",
+      action: "mark-applied",
+      payload: { reason: "cross-site metadata" },
+      headers: { "sec-fetch-site": "cross-site" },
```
ebarti, line 587: **[Low]** Missing the realistic cross-site case. This cross-site row sends no Origin/Referer, but a real browser always attaches an Origin to a POST, so this exercises the `allowLoopbackSameSite: false` branch. The realistic attack this PR is meant to close — a page on `http://localhost:PORT` hitting `http://127.0.0.1:8766` — produces a *loopback* Origin plus `Sec-Fetch-Site: cross-site`, exercising the `allowLoopbackSameSite: true` branch (trusted set `{same-origin, same-site, none}`). That branch is exactly the incremental protection added beyond gate 1 + CORS (both already allow that loopback Origin), and no case covers it. Suggest adding a matrix row like `headers: { origin: "http://localhost:5555", "sec-fetch-site": "cross-site" }` expecting 403 and asserting non-persistence.

PR #279, opened 2026-07-05
feat(api): [R5] project career evidence usage

ebarti:
## Summary
- add the evidence_usage_projections read-model table in the TS API and Python projection builder
- project profile achievement/skill evidence into resume, requirement-fit, and skill-coverage usage rows
- expose GET /v1/evidence-map and document the new read model/API surface

## Validation
- pnpm check
- pnpm api:test
- pnpm docs:build
- uv --project workers/automation run --extra dev pytest -q
- uv --project workers/automation run --extra dev pytest -q workers/automation/tests/test_projection_builder.py::test_evidence_usage_projection_inverts_profile_provenance_and_requirement_fit
- uv --project workers/automation run --extra dev ruff check workers/automation/src/jobhunter/infrastructure/projections/projection_builder.py workers/automation/src/jobhunter/infrastructure/projections/sqlite_projection_store.py workers/automation/tests/test_projection_builder.py
- git diff --check


apps/api/src/projections.ts
```diff
@@ -1245,6 +1336,464 @@ function loadProvenanceAuxByArtifact(
...
+  const rows = allRows<BulletProvenanceRow & { job_title: string | null; employer: string | null }>(
+    db,
+    `SELECT provenance.job_url, provenance.artifact_id, provenance.generation,
+            provenance.bullet_id, provenance.section, provenance.source_id,
+            provenance.evidence_ids_json, provenance.requirement_ids_json,
+            provenance.matched_keywords_json, provenance.transform_type,
+            provenance.control, provenance.rationale, provenance.generated_text,
+            provenance.position, provenance.created_at,
+            ${jobMetadata.selectSql}
+       FROM job_bullet_provenance AS provenance
```
ebarti, line 1782: PR Reviewer - pr-reviewer

Severity: High
Gate: Fail
Evidence: `attachResumeUsages` (this query), `attachRequirementUsagesAndGaps` (`FROM job_requirement_fit_items`, ~L1638), and `attachSkillCoverageUsagesAndGaps` (`FROM artifact_list_projections`, ~L1724) all filter only on `tenant_id`. None joins/excludes `jobhunter_deleted_jobs` or `jobhunter_hidden_jobs`. The default delete is a soft delete: `softDeleteJobs` (apps/api/src/write-model.ts:497) only inserts a tombstone row; `job_bullet_provenance`, `job_requirement_fit_items`, and `artifact_list_projections` rows for that job persist. Every other tenant-wide read path filters these (e.g. `rebuildDashboardProjection` at projections.ts:3047-3119, `listArtifacts`/conversion/employer/apply read-models in read-model.ts). The Python builder has the identical omission (`_attach_resume_usages` L1051, `_attach_requirement_usages_and_gaps` L1111, `_attach_skill_coverage_usages_and_gaps` L1192).

Reproduction: Reasoned from code. Score+tailor a job so it produces evidence/requirement/coverage rows, then soft-delete or hide it via `DELETE`/`hide`. Its title, employer, `generatedTextPreview`, and requirement gaps still appear in `GET /v1/evidence-map` because the projection rebuild reads the still-present source rows.

A job the user explicitly deleted or hid reappears in the new Career Evidence Map audit surface, re-exposing its title, employer, and generated-text preview and inflating usage/gap counts. This is the known deleted/hidden regression class the read model is supposed to prevent, and it ships in both runtimes.

Required fix: Exclude soft-deleted and hidden jobs in all three attach queries in BOTH builders, mirroring the existing `LEFT JOIN jobhunter_deleted_jobs d ... WHERE d.job_url IS NULL` + `LEFT JOIN jobhunter_hidden_jobs h ... AND h.unhidden_at IS NULL` pattern used by the dashboard/artifact projections (guarded by `tableExists`). Add a regression fixture (empty/populated + one deleted and one hidden job) asserting their usages and gaps are absent from the projection.

workers/automation/src/jobhunter/infrastructure/projections/projection_builder.py
```diff
@@ -814,6 +828,402 @@ def _rebuild_job(self, job_url: str) -> None:
...
+        if not _table_exists(self._conn, "job_bullet_provenance"):
+            return
+        job_metadata = _job_metadata_join_sql(self._conn, "provenance.job_url")
+        rows = self._conn.execute(
+            f"""
+            SELECT provenance.job_url, provenance.artifact_id, provenance.generation,
+                   provenance.bullet_id, provenance.generated_text, provenance.created_at,
+                   provenance.evidence_ids_json,
+                   {job_metadata['select_sql']}
+              FROM job_bullet_provenance AS provenance
```
ebarti, line 1210: PR Reviewer - pr-reviewer

Severity: High
Gate: Fail
Evidence: Python mirror of the deleted/hidden omission flagged on the TS side (apps/api/src/projections.ts:1576). `_attach_resume_usages` (this query), `_attach_requirement_usages_and_gaps` (L1111), and `_attach_skill_coverage_usages_and_gaps` (L1192) filter only on `tenant_id`, with no `jobhunter_deleted_jobs`/`jobhunter_hidden_jobs` exclusion. The worker `ProjectionBuilder` fully rebuilds this same table, so whichever runtime runs last wins; both must exclude soft-deleted and hidden jobs or the map re-surfaces them.

Reproduction: Reasoned from code (parity with TS finding).

Required fix: Add the deleted/hidden exclusion to all three Python attach queries, matching the pattern already used by the Python builder for other tenant-wide projections, and cover it in the shared parity fixture.

workers/automation/src/jobhunter/infrastructure/projections/projection_builder.py
```diff
@@ -2881,6 +3353,58 @@ def _preview_text(value: str, limit: int) -> str:
     return value if len(value) <= limit else f"{value[:limit]}..."
 
 
+def _requirement_fit_status_to_read_model(value: dict[str, Any]) -> dict[str, Any]:
```
ebarti, line 3875: PR Reviewer - pr-reviewer

Severity: Medium
Gate: Pass
Evidence: This is a new dual-writer projection, but there is no cross-runtime parity test. The repo convention for dual-writer projections is a shared parity fixture that asserts both builders emit identical output (apps/api/test/audit-projection-parity.test.ts + workers/automation/tests/test_audit_projection_parity.py; also test_domain_event_parity.py). Here each runtime has its own hand-written fixture asserting different things (the TS test at apps/api/test/projections.test.ts asserts via `GET /v1/evidence-map`; the Python test asserts projection rows directly), and this file re-implements `_requirement_fit_status_to_read_model`/`_requirement_artifact_coverage_to_read_model` plus string helpers that already exist in TS. Concrete latent divergences: `_preview_text` truncates by Unicode code points while the TS `previewText` slices by UTF-16 units (different `generatedTextPreview`/`title` for astral chars at the 140/240 boundary); `_strings_from_unknown` maps a JSON `null` element to the literal "None" while TS `parseStringList` maps it to "" and drops it. Because both builders fully DELETE+rebuild the same table, whichever ran last wins, so the same `GET /v1/evidence-map` read can flip between two shapes.

Reproduction: Reasoned from code (helper-by-helper comparison of the two builders).

Required fix: Add a shared cross-runtime parity fixture (seeding profile evidence AND skills AND bullet provenance+coverage AND requirement-fit matched/missing/blocked/transferable AND a deleted/hidden job) that both suites bind to and assert byte-identical projection payloads, following the audit-projection-parity pattern. Align `_preview_text` and `_strings_from_unknown` with their TS counterparts.

apps/api/src/projections.ts
```diff
@@ -738,8 +755,16 @@ export function refreshProjections(db: SqliteDatabase, tenantId = "local"): void
...
         sourceQualityDirty = true;
       }
+      evidenceUsageDirty = true;
     }
   }
+  const evidenceProjectionCount =
+    getRow<{ c: number }>(db, "SELECT COUNT(*) AS c FROM evidence_usage_projections WHERE tenant_id = ?", [
+      tenantId,
+    ])?.c ?? 0;
+  if (evidenceProjectionCount === 0) {
```
ebarti, line 780: PR Reviewer - pr-reviewer

Severity: Medium
Gate: Pass
Evidence: `if (evidenceProjectionCount === 0) evidenceUsageDirty = true` (and the Python mirror `if not evidence_usage_exists` at projection_builder.py:544) uses row-count==0 as the dirty signal. When a profile legitimately yields zero entries AND zero gaps (fresh user / no achievement evidence and no skills), the projection stays empty forever, so every `refreshProjections` call forces a full cross-table rebuild scan (profile x provenance x requirement-fit x coverage) that produces nothing — and, because `!evidenceUsageDirty` is part of the shared early-return guard (L819-826), it also defeats the skip that would otherwise avoid ALL projection work. `refreshProjections` runs on the hot read paths: `listJobs`, `getJobDetail`, and `listEvidenceMap` (read-model.ts). Separately, `evidenceUsageDirty = true` is set for EVERY new event (L758), so during any active pipeline run every jobs-list poll triggers a full evidence rebuild even for events that cannot affect the map (discovery/apply/hide).

Reproduction: Reasoned from code. With no profile evidence/skills, each `GET /v1/jobs` call re-enters `rebuildEvidenceUsageProjection` and never early-returns.

Required fix: Gate this projection on a watermark / built-marker like the other projections (dashboard settles to a singleton row so count is never 0; this table can be legitimately empty), instead of row-count==0, so an empty-but-current projection is not perpetually dirty. Scope the dirty trigger to event types that can change the map (profile / scoring / materials) rather than every event, mirroring `SOURCE_QUALITY_EVENT_TYPES`.

apps/api/src/projections.ts
```diff
@@ -1245,6 +1336,464 @@ function loadProvenanceAuxByArtifact(
...
+        fitKind,
+        reason,
+        jobRefs: [usage],
+      };
+      gaps.set(gap.gapId, gap);
+    }
+  }
+}
+
+function attachSkillCoverageUsagesAndGaps(
```
ebarti, line 1911: PR Reviewer - pr-reviewer

Severity: Low
Gate: Pass
Evidence: `attachSkillCoverageUsagesAndGaps` links a profile skill entry to a coverage-audit keyword by lowercased string equality (`skillEntriesByName.get(keyword.toLowerCase())`). The coverage STATE (covered/declared/missing) is legitimately generation-time-recorded against the rendered artifact, so this satisfies the "no read-time recomputation from job text" rule. But the attribution to a specific profile skill is a name match: the covered keyword is present in the shipped text because the JOB demanded it, not necessarily because of that skill line, so `coverageUsages` asserts presence, not recorded provenance. This differs from `resumeUsages`/`requirementUsages`, which come from recorded evidence IDs in `job_bullet_provenance`/`fit_json`.

Required fix: No data-layer change required — the payload already exposes `keyword` + `coverageState` so it is inspectable. Ensure the consuming UI (#283) labels skill coverage as coverage-audit-derived (name match against shipped keywords), distinct from the evidence-ID-backed resume/requirement usages, so it is not presented as recorded per-bullet provenance.

ebarti (commented):
PR Reviewer - pr-reviewer — Summary

**Gate: FAIL** (2 High). Open-decision implemented: **projection** (dual-builder `evidence_usage_projections`), matching the plan recommendation. The choice is recorded in `docs/architecture/read-model.md` and `docs/local-ts-api.md`.

Scope is correct: read-model + API only; no new generation pipeline. The two load-bearing linkages are honestly sourced — `resumeUsages` from recorded `job_bullet_provenance.evidence_ids_json`, `requirementUsages` from recorded `fit_json.evidence_ids`. Backfill for pre-PR jobs is honest (rebuilds from existing canonical provenance/fit rows; empty arrays where nothing was recorded — no fabrication). Contract payloads match `packages/contracts/src/schemas.ts`. Docs updated for route + read-model.

| - | -------- | -------- | ------- |
| 1 | High | projections.ts:1576 (+ .py:1051) | All three attach queries omit `jobhunter_deleted_jobs`/`jobhunter_hidden_jobs` exclusion; soft-delete keeps source rows, so deleted/hidden jobs re-surface in the evidence map. Both runtimes. |
| 2 | Medium | projection_builder.py:3356 | New dual-writer projection with no cross-runtime parity test; two independent fixtures + re-implemented normalizers/helpers; concrete latent divergences (`_preview_text` code-points vs UTF-16, `_strings_from_unknown` null→"None"). |
| 3 | Medium | projections.ts:765 (+ .py:544) | `count==0` dirty signal makes a legitimately-empty projection rebuild on every `refreshProjections` (hot `listJobs`/`getJobDetail` path) and defeats the shared early-return; dirty-on-every-event is over-broad. |
| 4 | Low | projections.ts:1704 | `skill_coverage` links profile skill→coverage keyword by name match; state is generation-time-recorded (OK) but attribution is presence not provenance — UI (#283) must label it. |
| 5 | Low | docs/architecture/storage.md:59 | Projection-table inventory not updated with `evidence_usage_projections` (list already omits workflow_run/source_quality, so pre-existing staleness). |

Verification (in `.claude/worktrees/review-pr279` at head 3a8d72a5): `pnpm api:check` pass; `pnpm api:test` 351 pass; `uv ... pytest -q` 1821 pass; `uv ... ruff check <touched>` pass.

Findings 2/3 also apply to the Python builder and should be fixed in lockstep. Recommend resolving the two High items (deleted/hidden exclusion in both builders + regression fixture) before merge.

apps/api/src/projections.ts
ebarti: PR Fixer - pr-fixer

Fixed in c146f19a.

Added a `tableExists`-guarded `jobLifecycleExclusionSql(db, jobUrlExpression)` helper (mirrors the neighbouring `jobMetadataJoinSql` and the `rebuildDashboardProjection` idiom) and applied it to all three attach queries: `attachResumeUsages` (`provenance.job_url`), `attachRequirementUsagesAndGaps` (`items.job_url`), and `attachSkillCoverageUsagesAndGaps` (`job_id`). Each now emits `LEFT JOIN jobhunter_deleted_jobs d ... WHERE d.job_url IS NULL` (restored_at-guarded) + `LEFT JOIN jobhunter_hidden_jobs h ... AND h.unhidden_at IS NULL`. The Python builder got the identical `_job_lifecycle_exclusion_sql` change so both runtimes stay in lockstep. Contract/API shape is unchanged (pure row filtering), so the descendant PRs are unaffected.

Regression fixture `excludes soft-deleted and hidden jobs from the career evidence map` (apps/api/test/projections.test.ts) seeds one active + one soft-deleted + one hidden job from canonical data and asserts that neither removed job's jobKey, title, employer, or generated-text preview appears in `GET /v1/evidence-map`, while the live job still does. Confirmed it fails on the unpatched code and passes with the fix.

Verified: `pnpm api:check` (clean), `pnpm api:test` (352 passed).

workers/automation/src/jobhunter/infrastructure/projections/projection_builder.py
ebarti: PR Fixer - pr-fixer

Fixed in c146f19a.

Added a `_table_exists`-guarded `_job_lifecycle_exclusion_sql(conn, job_url_expression)` helper (mirrors the TS `jobLifecycleExclusionSql` and the builder's existing deleted/hidden idioms) and applied it to all three attach queries: `_attach_resume_usages` (`provenance.job_url`), `_attach_requirement_usages_and_gaps` (`items.job_url`), and `_attach_skill_coverage_usages_and_gaps` (`job_id`). Each now excludes soft-deleted (`d.job_url IS NULL`, restored_at-guarded) and hidden (`h.unhidden_at IS NULL`) jobs, identical semantics and join keys to the TS builder so whichever runtime rebuilds last produces the same result.

Regression fixture `test_evidence_map_excludes_soft_deleted_and_hidden_jobs` (workers/automation/tests/test_projection_builder.py) seeds active + soft-deleted + hidden jobs and asserts the built `evidence_usage_projections` never reference the removed jobs' keys, titles, employers, or generated-text previews, while the live job still does. Confirmed it fails on the unpatched code and passes with the fix.

Verified: `uv --project workers/automation run --extra dev pytest -q` (1822 passed), `uv --project workers/automation run --extra dev ruff check .` (clean).

PR #277, opened 2026-07-05
feat(api): [R3] add extension capability token substrate

ebarti:
## What

Delivers the browser-extension plan's §3.3/P0 substrate because merged W2.3 only added the `Sec-Fetch-Site` CSRF gate and did not include capability-token or extension-origin trust.

- Adds a local browser-extension capability token under the JobHunter app dir (`~/.jobhunter/extension-capability-token` by default), with Settings pairing/rotation routes and a compact Settings UI.
- Adds route-scoped `chrome-extension://` CORS and bearer-token trust for authenticated `/v1/extension/*` routes while preserving the loopback Host gate.
- Keeps ordinary loopback web/CLI API behavior additive and unchanged; extension tokens do not authorize non-extension routes or application submission.
- Updates API/client contracts, web ports/hooks, regression tests, and owning security/API/QA docs.

## Why

PR #265 assumes §3.3 was delivered by OSS W2.3, but merged `main` only had the CSRF `Sec-Fetch-Site` hardening. Phase 1 capture needs a paired local-extension trust substrate before accepting browser-extension writes.

## Validation

Final stack validation after rebasing onto `origin/main` `7451055b`:

- `corepack pnpm check` — passed.
- `corepack pnpm test` — passed: API 20 files / 359 tests; web build passed; extension 5 files / 24 tests; extension e2e 2 files / 4 tests; Python 1818 passed.
- `corepack pnpm --filter @jobhunter/web test` — passed: 148 files / 912 tests.
- `corepack pnpm docs:build` — passed; 4548 refs across 230 emitted files.
- `python scripts/release_check.py` — passed with existing warnings in `workers/automation/src/jobhunter/apply/prompt.py`.
- `git diff --check` — passed.
- GitHub checks on this PR: release check and dead-link build passed; Cloudflare Pages skipped; DCO skipped under the repository-owner exemption workflow.

PR #281, opened 2026-07-05
feat(extension): [R3] add browser capture ingest

ebarti:
## What

Delivers Phase 1 capture from the browser-extension plan on top of the P0 auth-substrate PR.

- Adds a Manifest V3 extension package under `apps/extension` with popup pairing, active-tab capture, bounded local queueing, and loopback-only API calls.
- Adds authenticated `POST /v1/extension/captures`, which seeds `manual_capture_queue` with `browser_extension_capture` / `manual_capture:extension` provenance and then delegates to the existing manual-capture worker importer.
- Uses stable `captureId` retry keys so pending retries update the same queue row, imported replays return the original import response, and dismissed replays return a terminal no-op response for extension queue cleanup.
- Extends manual-capture provenance with `captureClient` / `extensionVersion`, including the Python importer persistence path and web Discovery label.
- Adds localhost-only extension privacy tests and folds extension checks into the root `check` / `test` aggregates.
- Updates README, user docs, local API docs, runtime/security docs, requirements, and QA docs for install, pairing, capture, and privacy boundaries.

## Why

The extension plan requires capture to reuse the existing Job Discovery manual-capture/import path rather than fork import, dedupe, snapshot, quarantine, or projection logic. This keeps the extension as a local client surface while preserving the existing loopback posture and user-mediated capture provenance.

## Validation

Final stack validation after rebasing onto `origin/main` `7451055b`:

- `corepack pnpm check` — passed.
- `corepack pnpm test` — passed: API 20 files / 359 tests; web build passed; extension 5 files / 24 tests; extension e2e 2 files / 4 tests; Python 1818 passed.
- `corepack pnpm --filter @jobhunter/web test` — passed: 148 files / 912 tests.
- `corepack pnpm docs:build` — passed; 4548 refs across 230 emitted files.
- `python scripts/release_check.py` — passed with existing warnings in `workers/automation/src/jobhunter/apply/prompt.py`.
- `git diff --check` — passed.
- GitHub checks on this PR: release check and dead-link build passed; Cloudflare Pages skipped; DCO skipped under the repository-owner exemption workflow.

Focused P1 checks during finalization:

- `corepack pnpm --filter @jobhunter/contracts check` — passed.
- `corepack pnpm --filter @jobhunter/api check` — passed.
- `corepack pnpm --filter @jobhunter/api exec vitest run test/discovery-controls.test.ts -t "extension capture"` — passed: 4 tests.
- `corepack pnpm --filter @jobhunter/extension check` — passed.
- `corepack pnpm --filter @jobhunter/extension test` — passed: 3 files / 8 tests at the P1 layer.

## Notes

- No P2b LLM answer drafts and no Phase 3 submission behavior are included.


apps/api/src/server.ts
```diff
@@ -373,6 +377,37 @@ export function buildApp(options: BuildAppOptions): FastifyInstance {
...
+    if (!seeded || "ok" in seeded) {
+      return seeded;
+    }
+    try {
+      return await manualCaptureImporter(seeded.itemId, extensionCaptureToManualCapture(body), {
+        appDir,
+        dbPath: options.dbPath,
+      });
+    } catch (error) {
+      if (error instanceof ManualCaptureImportError) {
```
ebarti, line 402: PR Reviewer - Codex

Severity: Medium
Gate: Pass
Evidence: `apps/api/src/server.ts:389` persists a pending extension `manual_capture_queue` row before `apps/api/src/server.ts:393-405` calls the importer and returns `manual_capture_import_failed` on importer errors. The extension then treats any non-auth API failure as queueable at `apps/extension/src/background.ts:91-105`, so the same capture can be posted again later with a new item id while the first server-side row remains pending.

If the worker/importer fails after the capture row is seeded, the browser extension keeps a local retry copy even though the API already accepted and persisted a pending manual-capture item. The next successful retry seeds a second pending row for the same capture; the original row remains visible/importable from Discovery, which creates duplicate retry surfaces for one user action.

Required fix: Make server-accepted/import-failed captures idempotent from the extension's perspective: either return an accepted/pending response with the seeded `itemId` so the extension does not enqueue a duplicate, or clean up/mark the seeded row before returning a retryable failure. Add a regression where `manualCaptureImporter` throws after seeding and assert a later retry cannot create a second pending row for the same capture.

ebarti (commented):
Code-review gate finding for P1 capture.

apps/api/src/server.ts
ebarti: PR Reviewer - Codex

Severity: Medium
Gate: Pass
Evidence: The new stable `captureId` hashing makes importer-failure retries reuse one `manual_capture_queue` id (`apps/api/src/discovery-controls.ts:876-881`), but `seedExtensionManualCapture` preserves an existing imported/dismissed status on conflict (`apps/api/src/discovery-controls.ts:845-855`) and the route still calls `manualCaptureImporter(...)` unconditionally after seeding (`apps/api/src/server.ts:397-405`). The Python importer only accepts `status = 'pending'` rows (`workers/automation/src/jobhunter/infrastructure/discovery/production_wiring.py:945-955`).

The amendment closes the duplicate-row case when import fails after seeding, but it is still not idempotent when the server imports successfully and the browser loses/times out the response. The queued retry replays the same `captureId`, preserves the already-imported row, then calls the pending-only importer and gets a not-found failure, so `flushCaptureQueue` keeps the local capture stuck even though the job was already imported.

Required fix: Detect an existing `imported` row for the stable extension item id and return the prior successful import response, or otherwise make replay after successful import a 2xx idempotent no-op that lets the extension clear the queued capture. Add a regression that imports once, retries the same `captureId`, and asserts the second call succeeds without invoking the pending-only importer.

apps/api/src/discovery-controls.ts
```diff
@@ -816,6 +830,124 @@ export function listManualCaptureQueue(db: SqliteDatabase): ManualCaptureListRes
...
+        source: "browser_extension",
+        capture_client: input.captureClient,
+        extension_version: input.extensionVersion,
+      },
+      null,
+      0,
+    ),
+    now,
+  );
+  const replay = importedExtensionManualCaptureResponse(db, itemId);
```
ebarti, line 902: PR Reviewer - Codex

Severity: Medium
Gate: Pass
Evidence: `apps/api/src/discovery-controls.ts:853-881` preserves an existing `dismissed` row when the same stable extension `captureId` is replayed, but `apps/api/src/discovery-controls.ts:900-904` only short-circuits imported rows. The route then calls the pending-only importer at `apps/api/src/server.ts:393-397`, and the extension retains failed queued sends in `apps/extension/src/queue.ts:63-72`.

A queued extension capture can still be stranded if the API seeds the row, the importer fails, and the user dismisses the pending manual-capture row before the local extension queue retries. The retry reuses the same item id, preserves `status = dismissed`, does not return a terminal/idempotent 2xx response, and then calls the importer for a non-pending row. That makes the extension keep retrying a capture the user already dismissed.

Required fix: Treat dismissed stable-id replays as terminal from the extension perspective, for example by returning a 2xx dismissed/no-op response that lets the local queue clear, or by allowing an explicit re-open policy. Add a regression that seeds, dismisses, replays the same `captureId`, and asserts the extension-facing response cannot strand the queue.

workers/automation/tests/test_discovery_production_wiring.py
```diff
@@ -1420,6 +1420,66 @@ def test_manual_capture_import_runs_discovery_enrichment_and_snapshot_pipeline(
     assert provenance["source_kind"] == "user_mediated_capture"
 
 
+def test_manual_capture_import_preserves_extension_capture_provenance(
```
ebarti, line 1423: PR Reviewer - pr-reviewer

Severity: Medium
Gate: Pass
Evidence: This new test covers extension-capture provenance (source_id/capture_client/extension_version) only. The plan §4.5.6 acceptance fixture and the Phase 1 DoD require the extension capture path to prove (b) dedupe/attach against an already-known URL *including a tracking-param variant* and (c) quarantine on an ambiguous capture, with "no silent duplicates (BR-021)". The API happy-path test (apps/api/test/discovery-controls.test.ts:916) does send a `?utm_source=newsletter` URL, but it injects a *stubbed* `manualCaptureImporter`, so `normalize_observed_url` / `find_canonical_owner` never run — the tracking-param dedupe is asserted nowhere end to end. No quarantine-on-ambiguous-capture test exists for the extension entry point either.
Reproduction: Reasoned from code — grep shows the only utm_source reference is in the stubbed-importer API test; the quarantine tests at lines 1201/1382 are for non-extension paths.

The dedupe/quarantine logic is correctly reused unchanged from DiscoverJobsUseCase, so regression risk is low, but the plan explicitly enumerated these fixtures as Phase 1 acceptance and the DoD checkbox ("Dedupe attaches ... incl. tracking-param variants; ambiguous captures quarantine") is claimed without a focused proof on the seedExtensionManualCapture -> import path.

Required fix: Add a Python fixture that seeds an extension capture (source_id manual_capture:extension) for a URL already known to JobHunter with a tracking-param variant (e.g. utm_source/gh_src) through import_manual_capture_item and asserts it attaches to the existing Job (no duplicate), plus a case that quarantines an ambiguous capture. This exercises the real dedupe/quarantine, not a stub.

apps/api/test/discovery-controls.test.ts
```diff
@@ -812,6 +814,358 @@ describe("discovery product controls API", () => {
...
+          "sec-fetch-site": "cross-site",
+        },
+        payload: {
+          originatingUrl: "https://example.com/jobs/extension",
+          captureMode: "current_page",
+          capturedUrl: "https://example.com/jobs/extension",
+          contentText: "Visible user-provided posting text.",
+        },
+      });
+      expect(unauthorized.statusCode, unauthorized.body).toBe(403);
```
ebarti, line 902: PR Reviewer - pr-reviewer

Severity: Low
Gate: Pass
Evidence: This assertion proves *a* rejection, but with `origin: chrome-extension://…` + `sec-fetch-site: cross-site` and no token, the onRequest hook trips the mutation-origin gate first (server.ts:312 -> 403 `cross_site_request`), never reaching the token check (server.ts:331 -> 401 `invalid_extension_capability_token`). So this test does not actually lock in that the *capability token* gates the endpoint — it locks in the cross-site-origin gate. A caller presenting a trusted loopback Origin but no/invalid token would be rejected 401 by isAuthorizedExtensionApiRequest (verified in code), but nothing tests that path. Separately, there is no test for an invalid payload (400 from ExtensionCaptureIngestSchema/parseBody), which the reviewer checklist calls for.
Reproduction: Reasoned from code — traced onRequest ordering in apps/api/src/server.ts:295-335; grep found no 401 invalid_extension_capability_token or 400 assertion for /v1/extension/captures.

Required fix: Add (1) a case with a valid loopback Origin (or the extension origin) but a missing/garbage Authorization header asserting 401 `invalid_extension_capability_token`, and (2) a malformed-payload case (e.g. missing originatingUrl or all-content-fields-absent) asserting 400 and that the importer is not called.

apps/extension/src/queue.ts
```diff
@@ -0,0 +1,98 @@
...
+): Promise<FlushResult> {
+  const queue = await loadCaptureQueue(storage, now);
+  const remaining: QueuedCapture[] = [];
+  let flushed = 0;
+  for (const item of queue) {
+    try {
+      await sendCapture(token, item.capture);
+      flushed += 1;
+    } catch {
+      remaining.push(item, ...queue.slice(flushed + remaining.length + 1));
```
ebarti, line 68: PR Reviewer - pr-reviewer

Severity: Low
Gate: Pass
Evidence: flushCaptureQueue stops draining at the first send that throws and re-queues that item plus all following it (head-of-line). background.ts:100 enqueues on *any* error except 401/403. A capture the server permanently rejects with a non-retryable 4xx (e.g. 400 payload/413 too-large) is therefore retried on every flush and blocks every later queued capture until it either ages out (CAPTURE_QUEUE_RETENTION_MS = 7 days) or is evicted by the 25-item cap. Transient failures (stack down / 5xx) are handled correctly; the gap is that non-retryable failures are treated as retryable.
Reproduction: Reasoned from code — a queued item whose payload the server 400s will fail on each flushCaptureQueue pass and break the loop before later items are attempted.

Impact is bounded (single-user, 25-item cap, 7-day expiry) and reaching a server 400 requires client/server schema drift since the client parses the schema before enqueue, hence Low.

Required fix: Distinguish retryable from non-retryable outcomes — on a non-retryable 4xx (client error, not 401/403 which are already special-cased) drop the poison item and continue flushing the rest instead of re-queueing and breaking. Re-queue only on network/5xx/timeout.

ebarti (commented):
PR Reviewer - pr-reviewer — Summary review (R3 P1 capture)

**Gate: PASS** (Blocker=0, High=0). Reviewed the diff of `feat/r3-extension-capture` @ 3d8540ea against its correct base `feat/r3-extension-auth-substrate` (#277 head dd99e293; merge-base equals the base, so no P0 commits leak into this diff).

## What is solid
- **Auth reuses the P0 substrate, not reimplemented.** `POST /v1/extension/captures` matches `isExtensionAuthenticatedApiPath` (starts with `/v1/extension/`, excludes the pairing paths), so the shared `onRequest` hook enforces `isAuthorizedLocalCapabilityToken` on top of the mandatory loopback `Host` gate. The token is required regardless of Origin; extension-auth.ts / local-origin.ts / config.ts are untouched.
- **Import path reused unchanged.** The endpoint seeds a pending `manual_capture_queue` row then delegates to the existing `createWorkerManualCaptureImporter` -> `import_manual_capture_item` -> `DiscoverJobsUseCase`/`CapturePostingSnapshotUseCase`. No import/dedupe/snapshot/quarantine/projection logic is forked.
- **Provenance (D-3) end to end.** `source_id = manual_capture:extension`, reason `browser_extension_capture`, `capture_client`/`extension_version` are seeded flat and re-nested by the Python importer under `manual_capture_provenance` (persisted with sort_keys), and both the first-response and imported-replay paths read the same nested shape. `source_kind = user_mediated_capture` preserved.
- **Idempotency/replay.** Stable `captureId` -> deterministic `extension:<sha256>` item id; pending retries update the same row; imported replay returns the original `ManualCaptureImportResponse`; dismissed replay returns a terminal 2xx no-op. Covered by four API tests.
- **Privacy I-1 (D-5).** Manifest is loopback-only (host_permissions + CSP connect-src), no `<all_urls>`/`https://*`; `scripting` is justified (page snapshot). Privacy tests assert the source and the built bundle. Offline queue is bounded (25 items / 7 days) and cleared on pair change (`saveToken`).
- **Frontend.** Only adds exhaustive switch arms for the new reason in the existing `ManualCapturePanel`; no new query keys/mutations/anti-patterns. No new `DomainEvent.eventType`, so SSE parity is unaffected (no casts/any in prod).
- **Docs.** README, user security/data-safety/getting-started/normal-flows, SECURITY.md, developer/security.md, runtime.md, local-ts-api.md, requirements (BR-019/TR-021), package.json aggregates all updated.

## Findings

| Severity | Location | Finding |
| --- | --- | --- |
| Medium | test_discovery_production_wiring.py:1423 | Extension-capture dedupe/tracking-param attach + quarantine fixtures (plan §4.5.6 b/c, DoD BR-021) missing; the utm_source API test uses a stubbed importer so normalization/dedupe never runs. Low regression risk (dedupe code reused) but the DoD claim is unproven for this entry point. |
| Low | discovery-controls.test.ts:902 | The "unauthenticated" 403 trips the cross-site-origin gate, not the token (401) gate; no test isolates that a missing/invalid token alone rejects, and no invalid-payload (400) test exists. Token gate is enforced in code. |
| Low | queue.ts:68 | Offline-queue head-of-line blocking: a non-retryable 4xx capture is re-queued and blocks later captures until expiry/eviction. Bounded (single-user, 25/7d) and hard to hit (client validates before enqueue). |

## Verification (in isolated worktree, no ports bound)
- `pnpm extension:check` pass; `pnpm extension:test` 3 files / 8 pass; `pnpm extension:e2e` build + bundle-scan pass
- `pnpm api:check` pass; `pnpm api:test` 20 files / 358 pass
- `pnpm web:check` pass; `pnpm --filter @jobhunter/web test` 148 files / 912 pass (incl. parity tests)
- `pytest tests/test_discovery_production_wiring.py` 23 pass; `ruff check .` pass

No Blocker/High. Medium + Low are non-blocking; recommend addressing the Medium dedupe/quarantine fixture before Phase 2 builds on this path.

PR #287, opened 2026-07-05
feat(materials): [R4] add artifact comparison

ebarti:
## Scope

Plan phase: R4 artifact comparison from PR #264 Delivery Model.

This PR adds:
- artifact comparison contracts and a materials selector over existing artifact detail rows
- an ArtifactComparison component with recorded, missing-coverage, and empty-selection states
- Apply Review composition between the accepted artifact and a rendered draft artifact
- Artifacts drawer comparison picker scoped to same job and artifact type
- fixtures, MSW rows, unit/a11y/story/e2e coverage, and narrow docs updates

Post-review amendments:
- artifact detail projections attach coverage, voice, and bullet provenance rows by exact artifact generation in both the TypeScript and Python builders
- comparison deltas include declared-only coverage transitions
- Apply Review pins the accepted artifact id from the draft base artifact across render/refetch and repeated renders

## Data Boundary

Displayed values come from artifact detail read rows, stored tailoring coverage audits, validator and judge fields, rendered draft artifact ids, and review comment risk labels. Missing coverage audit rows stay visible as coverage-not-recorded states.

No additional threshold was added.

## Validation

- PASS `pnpm check`
- PASS `pnpm test` (API 352 tests, web build, Python 1819 tests)
- PASS `pnpm --filter @jobhunter/web test` (153 files, 933 tests)
- PASS `pnpm --filter @jobhunter/web test-d` (9 type tests)
- PASS `pnpm --filter @jobhunter/web e2e -- tests/artifact-comparison.spec.ts` (2 tests)
- PASS `pnpm --filter @jobhunter/api exec vitest run test/projections.test.ts` (29 tests)
- PASS `uv --project workers/automation run --extra dev pytest -q workers/automation/tests/test_audit_projection_parity.py` (3 tests)
- PASS `pnpm --filter @jobhunter/web exec vitest run src/views/apply-review/ApplyReviewView.test.tsx src/contexts/materials/components/ArtifactComparison.test.tsx src/contexts/materials/selectors/compareCoverage.test.ts` (46 tests after baseline pinning)
- PASS `pnpm docs:build`
- PASS `pnpm web:storybook:build`
- PASS built Storybook render check for Recorded Coverage, Coverage Not Recorded, and Empty Selection stories: no browser warnings/errors and no failing responses
- PASS `git diff --check feat/r4-outcome-analytics-view...HEAD`
- PASS prohibited-phrasing diff scan against `feat/r4-outcome-analytics-view...HEAD`: no matches

## Known Existing Runner Stop

Full `pnpm web:storybook:test` still stops on `apps/web/src/contexts/pipeline/components/JobActions.stories.tsx` with `Cannot read properties of null (reading 'stores')`. The new ArtifactComparison stories render cleanly in the built Storybook check above.


apps/web/src/views/apply-review/ApplyReviewView.tsx
```diff
@@ -1152,8 +1208,20 @@ function SelectedReview({ item }: { readonly item: ApplyReviewQueueItem }) {
           </header>
           <div className="apply-review-pane-scroll apply-review-materials-scroll">
             {resumeAuditArtifactId ? <ArtifactGroundingRiskPanel artifactId={resumeAuditArtifactId} /> : null}
+            <ArtifactComparison
+              emptyRightMessage="Render a saved draft to compare it with the accepted artifact."
+              leftArtifactId={resumeAuditArtifactId}
```
ebarti: PR Reviewer - Codex

Severity: High
Gate: Fail
Evidence: This new comparison uses `resumeAuditArtifactId` from the live queue item as the "Accepted" side (`apps/web/src/views/apply-review/ApplyReviewView.tsx:1043`, `:1211-1218`). Rendering a draft already invalidates `applyReviewKeys.queue` on settle (`apps/web/src/contexts/apply/hooks/useApplyReviewMutations.ts:234-235`), while the API persists the rendered draft as a newer approved material/artifact (`apps/api/src/resume-review-drafts.ts:860-882`) and the Apply Review queue selects the highest-generation approved material preview (`apps/api/src/application-feedback.ts:1800-1845`, sorted by generation at `:2451-2458`). After the queue refetches, `item.materialsPreview.resumeTextArtifactId` can become the rendered draft id, so the comparison's Accepted side and Rendered draft side collapse to the same artifact and the last accepted artifact/editor is no longer the stable baseline. The added tests stub `applyReviewQueue` as a static response, so they do not exercise this render/refetch boundary.

The phase requires Apply Review to keep the accepted artifact/editor visible while showing the comparison. Because the accepted side is derived from mutable queue data after render, this implementation can replace the baseline before approval.

Required fix: Pin the accepted baseline for the review session/comparison (for example from the draft `baseResumeTextArtifactId`/`baseResumePdfArtifactId` or an explicit pre-render snapshot) and keep the editor/comparison left side on that baseline until an approval decision replaces it. Add a regression where `renderResumeReviewDraft` is followed by an Apply Review queue refetch returning the new rendered artifact as the latest material, and assert the comparison still shows the old accepted artifact on the left and the rendered artifact on the right.

apps/web/src/contexts/materials/components/ArtifactComparison.tsx
```diff
@@ -0,0 +1,224 @@
...
+      </div>
+    </div>
+  );
+}
+
+function coverageCountLabel(side: ArtifactComparisonSide): string {
+  if (!side.coverageRecorded || !side.coverageCounts) {
+    return "coverage not recorded";
+  }
+  return `${side.coverageCounts.covered}/${side.coverageCounts.planned} covered; ${side.coverageCounts.missing} missing`;
```
ebarti: PR Reviewer - Codex

Severity: Medium
Gate: Pass
Evidence: The stored `BulletCoverageAudit` read model has three outcome buckets (`covered`, `declared`, `missing`) plus `counts.declared` (`packages/contracts/src/schemas.ts:2583-2595` in this branch), but the comparison side label only renders covered/planned and missing counts here, and the delta panel only renders newly covered/lost/still missing (`apps/web/src/contexts/materials/components/ArtifactComparison.tsx:140-143`). The selector likewise compares `right.covered`, `left.covered`, and missing intersection only (`apps/web/src/contexts/materials/selectors/compareCoverage.ts:90-92`). A keyword that moves from missing to declared, or covered to declared, is stored in the audit row but disappears from the comparison instead of staying traceable.

The phase says comparison should use the stored coverage audit rows. Dropping the declared bucket makes the displayed comparison incomplete for current Phase 3 audit shape.

Required fix: Surface `declared` in the side summary and add declared-aware delta output or an explicit declared group, then add selector/component tests where a keyword moves between missing/declared/covered so the stored audit bucket remains visible.

ebarti (commented):
Found one High phase-compliance issue: the comparison does not pin the accepted artifact across the render/refetch boundary, so the rendered draft can become both sides of the comparison before approval. Also left one Medium coverage-audit completeness issue.

apps/web/src/views/apply-review/ApplyReviewView.tsx
ebarti: PR Reviewer - Codex

Severity: High
Gate: Fail
Evidence: The current fix snapshots the comparison baseline immediately before each render by assigning `renderBaselineArtifactId.current = auditArtifactId` in `apps/web/src/views/apply-review/ApplyReviewView.tsx:849`, but `auditArtifactId` is recomputed from the live Apply Review queue row at `apps/web/src/views/apply-review/ApplyReviewView.tsx:815-817`. Rendering still invalidates/refetches the queue on settle in `apps/web/src/contexts/apply/hooks/useApplyReviewMutations.ts:234-235`. After the first render/refetch, the queue can point `materialsPreview.resumeTextArtifactId` at the rendered draft artifact; a second render before approval then overwrites the baseline ref with that draft id, and `comparisonDraftTarget` passes the overwritten value into the left side at `apps/web/src/views/apply-review/ApplyReviewView.tsx:859-865`. The added regression only covers one render followed by a queue refetch (`apps/web/src/views/apply-review/ApplyReviewView.test.tsx:1601-1616`, `:1662-1679`), so it does not catch baseline drift across a second render. The new architecture note also promises that the last accepted artifact remains visible until approval (`docs/architecture/materials.md:281-289`).

The accepted-side pin is still not stable for a normal edit/render retry sequence. Once the queue has refreshed to the candidate artifact, any later render can make the previous draft the “Accepted” side instead of the last approved artifact.

Required fix: Capture the accepted baseline once per review session/draft from a stable source such as `draft.baseResumeTextArtifactId` or the pre-render queue snapshot, and do not overwrite it on later renders until approval/job switch replaces the accepted artifact. Add a regression that renders once, refetches the queue with the rendered artifact as latest, renders again before approval, and asserts the left side remains the original accepted artifact while the right side is the newest rendered draft.

apps/web/src/views/apply-review/ApplyReviewView.tsx
```diff
@@ -842,7 +856,23 @@ function ResumeLineReview({
...
+        renderBaselineArtifactId.current ?? draftBaseArtifactId(draft) ?? auditArtifactId,
+      ),
+    [auditArtifactId, draft, renderedDraftResult],
+  );
+
+  useEffect(() => {
+    onComparisonTargetChange?.(comparisonTarget);
+  }, [comparisonTarget, onComparisonTargetChange]);
+  useEffect(() => {
+    renderBaselineArtifactId.current = null;
```
ebarti, line 874: PR Reviewer - code-review

Severity: Medium
Gate: Pass
Evidence: `renderBaselineArtifactId` is pinned in `handleRenderDraft` (ApplyReviewView.tsx:849-850), read in `comparisonTarget` (864-866), and reset only by the effect on this line keyed to `[item.jobKey]` (873-875). Changing the resume template in-place runs `handleTemplateChange` → `setJobResumeTemplate` + `useEnsureCurrentResumeMaterialsMutation({force:true})` (1121-1136), which regenerates a new accepted generation WITHOUT changing `item.jobKey`.
Reproduction: Reasoned from code. Existing tests cover repeated renders (test 1551) and job-switch reset (test 1688), but not render → template-change → render.

After the user renders a draft, then switches the resume template (regenerating the accepted generation), then renders again, the pinned "Accepted" baseline still points at the pre-regeneration accepted artifact. The comparison then pairs the NEW rendered draft against a superseded generation, labels that stale generation "Accepted", and folds the template-change coverage difference into the draft-vs-accepted delta. This is display-only (no persisted attribution), but it misrepresents "what changed" on an audit surface.

Required fix: Reset/recompute the pinned baseline when the accepted generation changes in place — e.g., key the reset on the draft base artifact id or the current materials generation, not only `item.jobKey` — and add a regression test for render → template-change → render.

apps/web/src/views/artifacts/ArtifactDetailPanel.tsx
```diff
@@ -30,6 +34,37 @@ function isSuppressed(status: string): boolean {
   return status.toLowerCase() === "suppressed";
 }
 
+function artifactComparisonListInput(type: string | null | undefined): ArtifactsListInput {
+  const input: ArtifactsListInput = {
+    pageSize: 200,
```
ebarti, line 39: PR Reviewer - code-review

Severity: Medium
Gate: Pass
Evidence: `artifactComparisonListInput` sets `pageSize: 200` and filters only by `type` (this function). `ArtifactListQuerySchema` (packages/contracts/src/schemas.ts:1638-1652) has no job/jobKey filter and caps `pageSize` at 200. `comparableArtifacts` then filters `candidate.jobKey === artifact.jobKey` client-side (ArtifactDetailPanel.tsx:47-56).
Reproduction: Reasoned from code. The added test only seeds one same-job candidate; it does not seed >200 other-job same-type artifacts ahead of the viewed job.

The Artifacts-view comparison picker fetches at most the 200 most-recent artifacts of the selected type across ALL jobs, then narrows to the same job in the client. Once a user accumulates more than 200 same-type artifacts and the viewed job’s siblings fall outside that most-recent window, the picker renders "No other artifact for this job and type was found in the current artifact list" even though comparable generations exist — silently defeating the plan’s "pick two generations / templates for a job" capability for older jobs at scale. It degrades to an honest empty state (never wrong data), but a real comparand is dropped.

Required fix: Scope the candidate fetch to the job (add a job filter to the artifacts list query, or a per-job artifacts read) instead of a cross-job 200-row cap filtered client-side. Add a test with same-type artifacts from other jobs ordered ahead of the viewed job’s siblings.

apps/web/src/contexts/materials/selectors/compareCoverage.ts
```diff
@@ -0,0 +1,138 @@
...
+  if (left && right) return "recorded";
+  if (!left && !right) return "not_recorded";
+  return left ? "right_not_recorded" : "left_not_recorded";
+}
+
+function computedAgainstLabel(left: BulletCoverageAudit, right: BulletCoverageAudit): string {
+  if (left.computedAgainst === right.computedAgainst) {
+    return left.computedAgainst;
+  }
+  return `${left.computedAgainst} / ${right.computedAgainst}`;
```
ebarti, line 115: PR Reviewer - code-review

Severity: Low
Gate: Pass
Evidence: `computedAgainstLabel` concatenates differing bases as `${left.computedAgainst} / ${right.computedAgainst}` (this function); `coverageDelta` computes set differences regardless of basis (compareCoverage.ts:89-99). Coverage bases differ across generations — `grounded_shipped_text_v1` vs `judge_claimed_legacy` (docs/architecture/tailoring.md; the same basis split is surfaced explicitly in ApplyReviewView revision/shippedFit labeling).

When the two artifacts’ coverage was computed against different bases (e.g., a legacy accepted artifact recorded as `judge_claimed_legacy` vs a fresh draft recorded as `grounded_shipped_text_v1`), the `+covered` / `lost` / `declared` chips are set differences taken across incompatible measurement methods. An entry can reflect the measurement-method change rather than a real coverage change. The only disclosure is the concatenated `computedAgainst` label next to the heading, which is easy to miss and does not tell the reader the delta is not directly comparable.

Required fix: When `left.computedAgainst !== right.computedAgainst`, surface an explicit "coverage bases differ — delta not directly comparable" caveat (or suppress the numeric delta) rather than only concatenating the two basis labels.

apps/api/src/projections.ts
```diff
@@ -1177,28 +1179,14 @@ function loadBulletProvenanceByArtifact(
...
-      ORDER BY position, bullet_id`,
-    [jobUrl, tenantId, Number(generation)],
+      WHERE job_url = ? AND tenant_id = ?
+      ORDER BY generation, position, bullet_id`,
+    [jobUrl, tenantId],
   );
   if (rows.length === 0) return result;
-  // All rows of one generation share the artifact they explain (the writer binds
-  // the whole set to one artifact_id). Group defensively by artifact_id anyway.
   const byArtifact = new Map<string, Record<string, unknown>[]>();
```
ebarti, line 1195: PR Reviewer - code-review

Severity: Low
Gate: Pass
Evidence: TS `loadBulletProvenanceByArtifact` selects every generation’s rows (`ORDER BY generation, position, bullet_id`, projections.ts:1182-1187) and groups them into the `byArtifact` map on this line, concatenating rows from any generations that share an `artifact_id`. Python `_load_bullet_provenance_by_artifact` loops `DISTINCT generation` and assigns `provenance[record.artifact_id] = ...` per generation (projection_builder.py), so for a shared `artifact_id` a later generation OVERWRITES the earlier one (last-wins).
Reproduction: Reasoned from code. The shared parity fixture (`audit_projection_parity.json`) does not include an `artifact_id` reused across generations.

If two generations ever shared one `artifact_id`, the two runtimes would diverge — TS merges both generations’ bullets into one set; Python keeps only the last generation’s — a silent read-model parity break on a dual-written surface. Today the writer binds one `artifact_id` per generation, so this is not reachable; it is a latent, unguarded asymmetry rather than a live bug, but it sits exactly where this PR widened the query from latest-only to all-generations.

Required fix: Make the runtimes converge on one tie-break (e.g., TS also last-generation-wins per `artifact_id`), or add a parity-fixture row that reuses an `artifact_id` across generations to lock the behavior. Low priority given the current one-artifact-per-generation invariant.

apps/web/src/views/apply-review/ApplyReviewView.tsx
ebarti: PR Reviewer - code-review (follow-up on this thread)

Both baseline-drift Highs look addressed in the current head (d0e3b949), so this thread reads as resolved-in-code (just not marked resolved):

- Accepted side no longer derives from the mutable queue: `handleRenderDraft` now pins `renderBaselineArtifactId.current = renderBaselineArtifactId.current ?? draftBaseArtifactId(draft) ?? auditArtifactId` (ApplyReviewView.tsx:849-850) and the comparison reads the same expression (864-866), where `draftBaseArtifactId` returns the draft-creation base `baseResumeTextArtifactId`/`baseResumePdfArtifactId` (998-1000) — a stable value that does not move on queue refetch.
- Set-once via `??` means a second render does not overwrite the pin.
- Regression added: the render-twice test seeds `refreshedQueue` with `resumeTextArtifactId = <rendered artifact>` after render, then renders twice, and asserts the left side stays "Accepted resume" (the draft base) both times (ApplyReviewView.test.tsx:1601-1686). I re-ran `pnpm --filter @jobhunter/web test` (933 passed) and this file specifically — green.

One RESIDUAL edge the sticky pin does not cover: it resets only on `item.jobKey`, so an in-place regeneration via a template change (handleTemplateChange → ensureCurrentMaterials force) leaves the pin on the pre-regeneration accepted artifact. I filed that separately as Medium on ApplyReviewView.tsx:874.

apps/web/src/contexts/materials/components/ArtifactComparison.tsx
ebarti: PR Reviewer - code-review (follow-up on this thread)

This Medium (dropped `declared` bucket) looks addressed in the current head (d0e3b949), so it reads as resolved-in-code:

- Selector now emits `newlyDeclared` / `declaredLost` / `stillDeclared` in addition to covered/lost/missing (compareCoverage.ts:95-98), and the component renders `+declared` / `declared lost` / `declared` groups plus a `; N declared` side count (ArtifactComparison.tsx:143-146, 185-187).
- Regression added: "keeps declared-only coverage in its own comparison bucket" asserts a keyword moving between missing/declared/covered stays visible (compareCoverage.test.ts:44-80), and the component test asserts the declared counts render (ArtifactComparison.test.tsx:91-99). Verified green via `pnpm --filter @jobhunter/web test`.

ebarti (commented):
PR Reviewer - code-review — Summary (R4 Phase 3: artifact comparison)

**Scope reviewed:** diff of `feat/r4-artifact-comparison` @ `d0e3b949` against its base `feat/r4-outcome-analytics-view` (merge-base == base tip, so the diff is clean with no foreign commits). This delivers plan Phase 3 (side-by-side artifact comparison in review). No outcome joins are introduced here (that is Phase 4), so the outcome-attribution / small-sample-rate / causal-framing concerns do not apply to this surface.

**What it does, and what holds up well:**
- Pure `compareArtifactCoverage` selector over stored `ArtifactTailoringExplanation.coverageAudit` buckets (covered/declared/missing) — honest set differences, no JD re-inference. The "coverage not recorded" invariant (no zero-coverage inference when a side lacks a coverage row) is enforced and directly tested.
- Projection change (TS + Python) to materialize provenance/coverage/voice for EVERY generation keyed by `artifact_id` (previously latest-only) is the right fix so a historical accepted artifact shows its own coverage. Dual-runtime parity is covered by paired new tests, and the read-model PDF sibling resolution stays generation-scoped (a gen-1 PDF resolves gen-1 coverage, not the newest).
- Frontend conventions honored: materials-context component, reads via Operations hooks (gated with `enabled`), views compose, render mutation invalidates the rendered artifact detail + lists. Stories (3 states) + a11y tests + e2e present. Docs updates are narrow and correct.

**Verification (all green, run in an isolated worktree):**
- `pnpm api:check` PASS; `pnpm web:check` PASS
- `pnpm --filter @jobhunter/api exec vitest run test/projections.test.ts` 29 passed; `test/audit-projection-parity.test.ts` 2 passed
- `pnpm --filter @jobhunter/web test` 153 files / 933 passed (confirms the global MSW `*/v1/artifacts` fixture change is safe); comparison-specific files 53 passed
- `uv --project workers/automation run --extra dev pytest -q .../test_audit_projection_parity.py` 3 passed; `ruff check` on changed files clean

**Findings (this review):**

| Severity | Location | Issue |
|---|---|---|
| Medium | ApplyReviewView.tsx:874 | Pinned "Accepted" baseline resets only on `jobKey`; render → in-place template regeneration → render again pairs the new draft against the superseded accepted generation |
| Medium | ArtifactDetailPanel.tsx:39 | Comparison picker fetches ≤200 same-type artifacts across ALL jobs then filters `jobKey` client-side (no job filter in `ArtifactListQuery`); older jobs beyond the window silently show no comparand |
| Low | compareCoverage.ts:115 | Delta computed across differing coverage bases (grounded vs judge-claimed-legacy) is only disclosed via a concatenated label |
| Low | apps/api/src/projections.ts:1190 | TS merges rows per `artifact_id` across generations; Python takes last-generation — latent parity divergence if an `artifact_id` is ever reused (not reachable under current invariant) |

**Prior review threads (Codex, from first commit f9834949) — verified resolved-in-code at d0e3b949 (threads left open, not marked resolved):**
- 2× High baseline-drift (accepted side from mutable queue / pin overwritten on 2nd render): fixed by pinning to the draft-creation base with set-once `??` semantics; covered by the render-twice regression test. Replied on the thread with evidence. My Medium above is a distinct residual (template regeneration) the sticky pin does not cover.
- 1× Medium dropped `declared` bucket: fixed — selector + component now surface declared transitions, with tests. Replied on the thread.

**Gate: PASS** (Blocker=0, High=0). 2 Medium + 2 Low remain, none merge-blocking; the previously-filed High findings are addressed in the current head.

PR #288, opened 2026-07-05
feat(web): [R8] add saved jobs table views

ebarti:
## Summary
- Adds reusable saved table view contracts for the Jobs table.
- Extends `FilterableDataGrid` with additive presentation controls for column order, visibility, widths, density, and saved-view actions.
- Wires Jobs view snapshots while keeping active filters and sort URL-owned.

## Stack
- Base: #267 (`docs/plan-saved-views-digest`)
- Next: `feat/r8-p2-saved-view-rules`

## Validation
Full stack validation was run from `feat/r8-p5-digest-cli`:
- `pnpm check`
- `pnpm test`
- `pnpm --filter @jobhunter/web test-d`
- `pnpm --filter @jobhunter/web test`
- `pnpm --filter @jobhunter/web e2e -- tests/dashboard.spec.ts`
- `pnpm web:storybook:build`
- `pnpm docs:build`
- `git diff --check`

Review gates:
- pr-reviewer: Gate: PASS
- QA: Gate: PASS

PR #289, opened 2026-07-05
feat(web): [R8] add saved view grouping and rules

ebarti:
## Summary
- Adds saved Jobs table grouping, density, color-rule, and presentation controls.
- Persists templates in the versioned saved table view Zustand store.
- Restores active saved-view grid filters on Jobs view remount without moving URL-owned filters or sort out of the route state.

## Stack
- Previous: #288 (`feat/r8-p1-saved-views-core`)
- Next: `feat/r8-p3-digest-read-model`

## Validation
Full stack validation was run from `feat/r8-p5-digest-cli`:
- `pnpm check`
- `pnpm test`
- `pnpm web:storybook:build`
- `pnpm docs:build`
- `git diff --check`

Review gates:
- pr-reviewer: Gate: PASS
- QA: Gate: PASS

### DIFF
--- a/apps/web/src/shared/ui/filterable-data-grid.tsx
+++ b/apps/web/src/shared/ui/filterable-data-grid.tsx
@@ -6,6 +6,7 @@ import {
   IconX,
 } from "@tabler/icons-react";
 import {
+  Fragment,
   type KeyboardEvent,
   type MouseEvent,
   type PointerEvent as ReactPointerEvent,
@@ -69,6 +70,20 @@ export interface DataGridSortState {
 export type DataGridDensity = "compact" | "regular" | "comfy";
 export type DataGridColumnVisibilityState = Record<string, boolean>;
 export type DataGridColumnWidthsState = Record<string, number>;
+export type DataGridColorTone = "success" | "warning" | "danger" | "info";
+
+export interface DataGridGroupingState {
+  columnId: string;
+}
+
+export interface DataGridColorRule {
+  columnId: string;
+  predicate: {
+    op: "eq" | "neq" | "gte" | "lte" | "contains";
+    value: string | number;
+  };
+  tone: DataGridColorTone;
+}
 
 export interface DataGridHeaderContext<TData> {
   pageRows: readonly TData[];
@@ -121,6 +136,8 @@ export interface FilterableDataGridProps<TData> {
   columnWidths?: DataGridColumnWidthsState;
   onColumnWidthsChange?: (next: DataGridColumnWidthsState) => void;
   density?: DataGridDensity | null;
+  grouping?: DataGridGroupingState | null;
+  colorRules?: readonly DataGridColorRule[];
   rowClassName?: (row: TData) => string | undefined;
   rowAriaSelected?: (row: TData) => boolean;
   onRowActivate?: (row: TData) => void;
@@ -230,6 +247,72 @@ function sameStrings(left: readonly string[], right: readonly string[]): boolean
     left.length === right.length &&
     left.every((value, index) => value === right[index])
   );
+}
+
+function columnRuleValue<TData>(
+  row: TData,
+  column: DataGridColumn<TData>,
+): string | number {
+  const filterValue =
+    column.getFilterSearchValue?.(row) ?? column.getFilterValue?.(row);
+  if (filterValue !== undefined) return filterValue;
+  return column.getSortValue?.(row) ?? "";
+}
+
+function groupLabelFor<TData>(
+  row: TData,
+  column: DataGridColumn<TData>,
+): string {
+  const value = column.getFilterValue?.(row) ?? column.getSortValue?.(row);
+  const label = String(value ?? "").trim();
+  return label || "Unspecified";
+}
+
+function matchesColorRule<TData>(
+  row: TData,
+  column: DataGridColumn<TData>,
+  rule: DataGridColorRule,
+): boolean {
+  const left = columnRuleValue(row, column);
+  const right = rule.predicate.value;
+  if (rule.predicate.op === "gte" || rule.predicate.op === "lte") {
+    const leftNumber = Number(left);
+    const rightNumber = Number(right);
+    if (!Number.isFinite(leftNumber) || !Number.isFinite(rightNumber)) {
+      return false;
+    }
+    return rule.predicate.op === "gte"
+      ? leftNumber >= rightNumber
+      : leftNumber <= rightNumber;
+  }
+  const leftText = normalize(String(left));
+  const rightText = normalize(String(right));
+  if (rule.predicate.op === "eq") return leftText === rightText;
+  if (rule.predicate.op === "neq") return leftText !== rightText;
+  return leftText.includes(rightText);
+}
+
+function firstMatchingTone<TData>(
+  row: TData,
+  columnsById: ReadonlyMap<string, DataGridColumn<TData>>,
+  colorRules: readonly DataGridColorRule[],
+  columnId?: string,
+): DataGridColorTone | null {
+  for (const rule of colorRules) {
+    if (columnId && rule.columnId !== columnId) continue;
+    const column = columnsById.get(rule.columnId);
+    if (column && matchesColorRule(row, column, rule)) {
+      return rule.tone;
+    }
+  }
+  return null;
+}
+
+function toneClass(
+  target: "row" | "cell",
+  tone: DataGridColorTone | null,
+): string | undefined {
+  return tone ? `data-grid-${target}-tone-${tone}` : undefined;
 }
 
 export function isActiveDataGridFilter(
@@ -410,6 +493,8 @@ export function FilterableDataGrid<TData>({
   columnWidths: controlledColumnWidths,
   onColumnWidthsChange,
   density,
+  grouping,
+  colorRules = [],
   rowClassName,
   rowAriaSelected,
   onRowActivate,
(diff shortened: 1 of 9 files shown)


apps/web/src/shared/ui/filterable-data-grid.tsx
```diff
@@ -232,6 +249,72 @@ function sameStrings(left: readonly string[], right: readonly string[]): boolean
   );
 }
 
+function columnRuleValue<TData>(
+  row: TData,
+  column: DataGridColumn<TData>,
+): string | number {
+  const filterValue =
+    column.getFilterSearchValue?.(row) ?? column.getFilterValue?.(row);
```
ebarti, line 257: PR Reviewer - pr-reviewer

Severity: Medium
Gate: Pass
Evidence: `columnRuleValue` (filterable-data-grid.tsx:252-260) resolves a row's value for a color rule as `getFilterSearchValue ?? getFilterValue ?? getSortValue`. The Jobs numeric columns return formatted strings from `getFilterValue`/`getFilterSearchValue`: `compensation_min_eur`/`compensation_max_eur` return `salaryAmountLabel(...)` e.g. "120,000 euros per year" (columns.tsx:509-510, 523-524); `compensation_market`, `compensation_confidence`, and `compensation_warnings` ("3 warnings") likewise. For `gte`/`lte`, `matchesColorRule` (lines 278-286) does `Number(left)` on that string, which is `NaN`, so `!Number.isFinite` short-circuits and the rule returns false. A numeric color rule on any compensation column therefore never matches; it works only for `fit_score`, whose `getFilterValue` is `String(row.fitScore)` (a bare number). The new unit test (filterable-data-grid.test.tsx:40-45) uses a synthetic `observed` column that has ONLY `getSortValue` and no `getFilterValue`, so the numeric path is validated only for that easy case and the real formatted-column failure is uncovered.
Reproduction: Reasoned from code — `Number("120,000 euros per year")` is `NaN`; value resolution prefers the formatted filter string over the numeric `getSortValue`.

A user who adds "Salary min (euro/year) >= 100000, tone success" sees zero rows highlighted with no error, while the same rule on Fit score works — the feature silently fails for its most natural numeric use case and behaves inconsistently across columns.

Required fix: For `gte`/`lte`, resolve the left operand from the numeric `getSortValue` first, falling back to parsing the filter value only when no numeric sort value exists. Add a table-driven test exercising a numeric rule against a column whose `getFilterValue`/`getFilterSearchValue` returns a formatted string (mirroring the compensation columns) so the numeric path is proven on realistic data.

apps/web/src/shared/ui/filterable-data-grid.tsx
```diff
@@ -926,65 +1046,95 @@ export function FilterableDataGrid<TData>({
...
+                              if (shouldIgnoreRowActivation(event)) {
+                                return;
+                              }
+                              onRowActivate(row);
+                            }
+                          : undefined
+                      }
+                    >
+                      {displayColumns.map((column, columnIndex) => {
+                        const content = column.render(row, { pageRows, rowId, rowIndex });
```
ebarti, line 1091: PR Reviewer - pr-reviewer

Severity: Medium
Gate: Pass
Evidence: With grouping active the grid renders rows reordered into groups (pageRowGroups at lines 626-646, rendered at 1049-1137), but each cell still receives the flat, un-grouped page order via `column.render(row, { pageRows, ... })` on this line — `pageRows` here is the top-level page array in server/visible order, not `group.rows`. The Jobs select column drives shift-range selection from `context.pageRows`: `updateSelectedRow(..., context.pageRows)` then `pageRows.slice(min, max)` (columns.tsx:324-338). When the group column differs from the sort column, the visual order no longer matches `pageRows`, so shift-clicking between two visually adjacent rows selects a range computed over the original order — a different, visually non-contiguous set than what appears between anchor and target.
Reproduction: Reasoned from code — group by Stage while sorted by Discovered, then shift-select between two rows shown adjacently under one group header; the selection is computed from their pre-group index positions.

The mis-selected set then feeds the bulk delete/hide/retry actions, so the operator can act on rows they did not intend to select.

Required fix: When grouping is active, drive shift-range selection from the displayed (grouped) row order — pass the group-flattened order as `context.pageRows` (or otherwise expose it to the selection logic) so `slice(min,max)` matches what the user sees. Add a test for shift-select under grouping.

apps/web/src/shared/ui/filterable-data-grid.tsx
```diff
@@ -926,65 +1046,95 @@ export function FilterableDataGrid<TData>({
...
-                </tr>
-              );
-            })}
+            {pageRowGroups.map((group) => (
+              <Fragment key={group.key}>
+                {group.label ? (
+                  <tr className="data-grid-group-row">
+                    <th colSpan={Math.max(1, displayColumns.length)}>
+                      <span>{group.label}</span>
+                      <span>{group.rows.length}</span>
```
ebarti, line 1055: PR Reviewer - pr-reviewer

Severity: Low
Gate: Pass
Evidence: The group header renders `group.rows.length` as the count beside each group label. For the Jobs table pagination is server-side (JobsTable passes the `pagination` prop, so `pageRows === visibleRows` = only the current fetched page — filterable-data-grid.tsx:614-618), and grouping is computed over that page (pageRowGroups, lines 626-646). The count is therefore the number of rows in that group on the current page only, not a dataset total, and the same group reappears with different counts on other pages.
Reproduction: Reasoned from code — group by Stage with more matching jobs than one page holds; each page shows its own per-page group counts.

An operator reading "apply 8" beside a group header will likely read it as a total rather than a per-page subtotal, which is misleading on a triage surface.

Required fix: Label the count as page-scoped (or omit it under server-side pagination); a true total must come from the read model, not the client page. At minimum, document the per-page semantics.

ebarti (commented):
## PR Reviewer summary — pr-reviewer

**Gate: PASS** (Blocker=0, High=0). Two Medium and three Low findings; none block merge, but the two Mediums are real feature defects worth fixing before this ships as "grouping + color rules work."

Reviewed #289 `feat/r8-p2-saved-view-rules` @ `8adf0c28` against base #288 `feat/r8-p1-saved-views-core` @ `2c086aa8` (verified: `2c086aa8` is the exact merge-base and an ancestor of head — clean diff, no base commits pulled into review). This is plan phase **P2 (grouping + color rules)**; the P1 store already reserved the `grouping`/`colorRules` schema and normalizers, and this PR wires the same persisted-client tier — state-discipline consistent with #288.

### Findings

|---|---|---|---|
| 1 | Medium | filterable-data-grid.tsx:257 (inline) | Numeric color rules (`gte`/`lte`) silently never match on compensation columns — value resolution prefers the formatted `getFilterValue` string over numeric `getSortValue`, so `Number("120,000 euros per year")` = NaN. Works only for `fit_score`. Unit test uses a synthetic column with only `getSortValue`, so the real failure is uncovered. |
| 2 | Medium | filterable-data-grid.tsx:1091 (inline) | Grouping breaks shift-range selection — cells receive the flat `pageRows` order while rows are displayed grouped, so `pageRows.slice(min,max)` (columns.tsx:324-338) selects a visually non-contiguous set that then feeds bulk delete/hide/retry. |
| 3 | Low | filterable-data-grid.tsx:1055 (inline) | Group-header count is per-page under server-side pagination, not a dataset total; likely misread as a total on a triage surface. |
| 4 | Low | (see below) | No Storybook story / a11y coverage for the new grouping + color-rule control UI or the group-header / toned-row grid states. |
| 5 | Low | (see below) | No end-user documentation for saved views (incl. grouping/color rules); PR description does not explain the omission. |

**Finding 4 (Low) — stories/a11y coverage.** `apps/web/src/shared/ui/filterable-data-grid.stories.tsx` covers Populated/Loading/Empty/FilteredNoMatch/Paginated/ActivatableRows but adds no grouped or color-toned story, and `saved-table-views-control.tsx` has no stories file at all — so the storybook a11y bar never exercises the new interactive control (four selects + value input + rule list) or the new grid states. The group-header row is `<tr class="data-grid-group-row"><th colSpan>` with no `scope`, which is exactly the kind of state a per-discriminant story would catch. Functional coverage exists (grid grouping+color test and the JobsView integration test), so this is coverage/convention, not a proven defect. Per the plan's verification (`pnpm web:storybook:test`, zero critical/serious) and the frontend conventions (per-state stories), add stories for the grouped grid, each color tone, and the control's rule editor.

**Finding 5 (Low) — user docs.** `docs/local-reliability-qa.md` (QA smoke) and `docs/backlog.md` (reconciled the delivered entries) were updated, but no `docs/user/*.md` describes saved views at all — grep for "saved view"/"group by"/"color rule" under `docs/user/` returns nothing (this is inherited from #288, which introduced the feature). Grouping + color rules is user-facing product behavior; per CLAUDE.md, either document it in `docs/user/normal-flows.md` or state in the PR description why no user doc is warranted.

### What's correct / verified

- **Scope + architecture:** additive engine props (all optional, existing call sites unchanged — no-strangler); store/control live in `shared/` (correct — saved views are a cross-table shared concern, not a backend context); `JobsView` composes (navigates + reads the shared store) with no `useQuery`/`useMutation`/`apiClient.*` in the view. Consistent with the view-vs-context dichotomy.
- **State discipline:** grouping/colorRules persist in the same Zustand+persist tier as #288's saved views; active filters/sort remain URL-owned. The new `localTableFilters` hydration effect (JobsView.tsx:288-292) restores a saved view's grid-local filters on remount without duplicating URL state.
- **Color tokens:** row/cell tones map to semantic tokens only (`--success`/`--warning`/`--destructive`/`--status-info` via `color-mix(in oklch, …)`); all verified defined in `tokens.css` for both themes — Token Foundation gate passes. No raw colors, no class/CSS injection (tone is a fixed literal, values render as escaped text).
- **Commit/title:** Conventional Commits with `[R8]` tag.

### Verification (run in isolated worktree)

- `pnpm install --frozen-lockfile` — OK
- `pnpm web:check` (tsc --noEmit) — PASS, clean
- `pnpm --filter @jobhunter/web test` — PASS (147 files, 914 tests)

Note: I could not run `code-review` / `security-review` / `archscan` skills or `pnpm web:storybook:test` in this environment; the review above is manual. No security-relevant surface in this diff (frontend rendering + client store only).

PR #290, opened 2026-07-05
feat(api): [R8] add local digest read model

ebarti:
## Summary
- Adds the local daily digest contract, `digest_state`, and passive `GET /v1/digest` read model.
- Adds shared TS/Python parity coverage for digest counts, UTC timestamp boundary behavior, failed apply-run readiness, eligibility-blocked readiness, and hidden/closed active job filtering.
- Adds digest deep links and documentation for the local API/storage surfaces.

## Stack
- Previous: #289 (`feat/r8-p2-saved-view-rules`)
- Next: `feat/r8-p4-digest-panel`

## Validation
Full stack validation was run from `feat/r8-p5-digest-cli`:
- `pnpm check`
- `pnpm test`
- `pnpm web:storybook:build`
- `pnpm docs:build`
- `git diff --check`

Focused digest validation:
- `pnpm --filter @jobhunter/api exec vitest run test/digest.test.ts`
- `uv --project workers/automation run --extra dev pytest -q workers/automation/tests/test_digest_parity.py workers/automation/tests/test_digest_cli.py`

Review gates:
- pr-reviewer: Gate: PASS
- QA: Gate: PASS

Known follow-up
- pr-reviewer reported a non-blocking Medium: follow-up digest counts are outcome-derived while the current deep link targets the existing applied Jobs view. No Blocker or High findings remain.


workers/automation/src/jobhunter/digest.py
```diff
@@ -0,0 +1,578 @@
...
+    count = _count_rows(conn, f"SELECT COUNT(*) AS count FROM job_list_projections{where}", params)
+    high_fit = _count_rows(
+        conn,
+        f"SELECT COUNT(*) AS count FROM job_list_projections{where} AND COALESCE(fit_score, -1) >= ?",
+        [*params, high_fit_threshold],
+    )
+    return {"count": count, "highFitCount": high_fit}
+
+
+def _blocked_sources(conn: sqlite3.Connection) -> dict[str, Any]:
```
ebarti, line 192: PR Reviewer - pr-reviewer

Severity: High
Gate: Fail
Evidence: `_blocked_sources` (digest.py:192) counts only `source_quality_stats` rows. The TS counterpart `digestBlockedSources` (apps/api/src/read-model.ts:524) filters `listSourceHealth(db)`, which additionally appends operational-only sources from `operational_attempt_metrics` (`operationalSourceRollups` -> `operationalMetricRows`, read-model.ts:3912 — no source_role/stage filter, only `outcome != 'started'`) via `sourceRollupToHealth`, which sets `consecutiveFailures = (lastOutcome === "failed" ? failures : 0)`. A source with >=3 failed attempts recorded in `operational_attempt_metrics` but with no `source_quality_stats` row therefore passes the TS `consecutiveFailures >= 3` filter and is counted, while Python never sees it. `blockedSources.count` then differs between the dashboard panel (TS) and the CLI (Python) on the same DB, violating plan gate Feature 2 #6 ("identical numbers on identical SQLite state"). The parity fixture only seeds `source_quality_stats` sources, so both `test_digest_parity.py` and `digest.test.ts` pass with this blind spot. The plan (digest data-source map, datum 2) cites `listSourceHealth` as the canonical source, so the Python side is the one that diverges.
Reproduction: Reasoned from code (parity tests pass but never seed an operational-only source).

The two runtimes must count the same source set for `blockedSources.count`, which the feature makes a hard parity gate.

Required fix: Make `_blocked_sources` fold in `operational_attempt_metrics` rollups the way `listSourceHealth` does (matching the plan's cited source), OR restrict TS `digestBlockedSources` to `source_quality_stats` only — then add a parity-fixture case with an operational-only source (>=3 failures, no `source_quality_stats` row) so the divergence is guarded.

workers/automation/src/jobhunter/digest.py
```diff
@@ -0,0 +1,578 @@
...
+            """
+            SELECT DISTINCT job_url
+            FROM job_score_staleness
+            WHERE tenant_id = ?
+              AND resolved = 0
+            """,
+            [TENANT_ID],
+        ).fetchall()
+        return sum(1 for row in rows if _row_get(row, "job_url") in active_jobs)
+    if _table_exists(conn, "job_list_projections"):
```
ebarti, line 543: PR Reviewer - pr-reviewer

Severity: Medium
Gate: Pass
Evidence: When `job_score_staleness` does not exist, `_count_stale_scores` (digest.py:500-511) falls back to `job_list_projections` rows with `current_stage='score' AND current_state='stale'`. The TS value it must match — `preparation.outdatedScoreCount` via `countOutdatedScores` (apps/api/src/read-model.ts:701) — instead falls back to a policy-version comparison over `job_scores` (latest version whose trace `scoring_policy_version` is null or `< currentPolicyVersion`, no correction). These are different computations and can produce very different counts in the no-staleness-table state. The parity fixture always creates `job_score_staleness`, so this branch is never exercised by `test_digest_parity.py` / `digest.test.ts`.

Lower reachability than the blockedSources divergence because `job_score_staleness` is normally present, but it is a second place where the Python digest and the TS read model compute the same contract number by different rules.

Required fix: Align the Python fallback with the TS policy-version fallback (read latest `job_scores` + trace policy version), or if the no-staleness-table state is genuinely unreachable, drop the divergent fallback and document why. If kept, add a parity case without `job_score_staleness`.

workers/automation/src/jobhunter/digest.py
```diff
@@ -0,0 +1,578 @@
...
+
+def _has_review_material(row: Any) -> bool:
+    return bool(
+        _row_get(row, "has_resume")
+        or _row_get(row, "has_cover_letter")
+        or _row_get(row, "has_pdf")
+    )
+
+
+def _digest_materials_ready(row: Any) -> bool:
```
ebarti, line 386: PR Reviewer - pr-reviewer

Severity: Medium
Gate: Pass
Evidence: `_apply_queue_rows` (digest.py:236) + `_digest_materials_ready` (digest.py:343) hand-reimplement the canonical TS apply-review-queue query (`listApplyReviewQueue`, apps/api/src/application-feedback.ts:236) and the readiness state machine (`buildApplyAudit`, apps/api/src/apply-audit.ts:45) that TS uses via `materials.ready = applyAudit.state === "ready"`. `reviewNeededMaterials`/`pendingApprovals` parity is asserted only at the count level against the shared fixture, so a future change to `buildApplyAudit` (a new hard-blocker, prerequisite, or state rule) silently desyncs the CLI from the panel. Concrete latent divergence today: TS normalizes the substage with `stage(row.current_substage ?? row.current_stage)` (nullish coalescing, application-feedback.ts:752) while Python uses `_stage(current_substage or current_stage)` (digest.py:349, falsy coalescing). An empty-string `current_substage` therefore normalizes to `"apply"` in TS but to the parent stage in Python, flipping the `current_stage != "apply"` readiness check and thus the `reviewNeededMaterials` count. The fixture always sets `current_substage == current_stage` (a valid stage), so it never exercises this.

Required fix: Match TS nullish semantics in Python (treat only `None`/missing as absent, not `""`). Add fixture coverage that drives readiness through a non-`apply` substage and an empty substage. Longer term, assert the full readiness-driving fields (not just the counts) so the two implementations cannot drift undetected.

workers/automation/tests/test_digest_parity.py
```diff
@@ -0,0 +1,407 @@
...
+        now=_parse_utc(fixture["now"]),
+    )
+
+    expected = fixture["expected"]
+    assert digest["ok"] is True
+    assert digest["generatedAt"] == expected["generatedAt"]
+    assert digest["since"] == expected["since"]
+    assert digest["highFitThreshold"] == expected["highFitThreshold"]
+    assert digest["newMatches"] == expected["newMatches"]
+    assert digest["blockedSources"]["count"] == expected["blockedSources"]["count"]
```
ebarti: PR Reviewer - pr-reviewer

Severity: Low
Gate: Pass
Evidence: The parity tests assert only `blockedSources["count"]` (test_digest_parity.py:51; digest.test.ts:126), not the `blockedSources.sources` array, which is part of the `DailyDigest` contract. TS orders sources by `recommended_state DESC, observed_jobs DESC, source_id ASC` (listSourceHealth) while Python orders by `recommended_state DESC, source_id ASC` (_blocked_sources, digest.py:204), so the rendered order can differ between the panel and the CLI even when counts match. Separately, every fixture job has `deleted_at = NULL`, so deleted-job exclusion is never exercised — both runtimes do filter `deleted_at IS NULL`, but this is the "deleted jobs reappearing" regression class the plan explicitly calls out, so a fixture row would lock it in.

Required fix: Assert the full `blockedSources.sources` array and align the two orderings; add at least one `deleted_at`-set fixture job so deleted-job exclusion is guarded in both runtimes.

ebarti (commented):
## PR Reviewer summary — R8 P3 local digest read model

Reviewed the diff of #290 against its base `feat/r8-p2-saved-view-rules` (merge-base = base tip `8adf0c28`, clean three-dot diff). This is P3 (digest read model + contract + watermark + TS/Python parity); no panel (P4) or CLI (P5) yet, consistent with the stack. Findings posted inline.

### What's solid
- **Day boundary is deliberate and consistent.** UTC is fixed in the contract (`DIGEST_DAY_BOUNDARY`), documented in `docs/local-ts-api.md`, and new-matches/follow-ups are timestamp-based (watermark), so the plan's flagged local-vs-UTC inconsistency is genuinely resolved, not repeated.
- **Deleted/hidden/closed exclusion** is applied in both runtimes for new matches, review-needed materials, and stale scores, and the fixture exercises hidden + closed exclusion. The route test asserts the digest new-matches count equals the `/v1/jobs` deep-link result (good auditability).
- **Passive read.** `GET /v1/digest` never advances the watermark (asserted in both tests), is read-only, and stays unauthenticated-loopback like its neighbors. `digest_state` is a watermark row, not a projection; the digest is a live read-time aggregation over existing projections (matches `read-model.md`), so there is no dual-writer projection to keep in lockstep.
- **Budget parity is sound** (both clamp `remainingUsd` to >=0; both derive `unlimited`/`status` consistently). Contract constants, docs (`local-ts-api.md`, `storage.md`), and client/port wiring are correct. `CLOSED_ACTIVE_STATES`, `STAGES`, `STAGE_STATES` match across runtimes.

### Findings

| # | Severity | Location | Issue |
|---|----------|----------|-------|
| 1 | **High** | digest.py:192 `_blocked_sources` | `blockedSources.count` diverges TS vs Python: TS `listSourceHealth` counts operational-only sources (`consecutiveFailures>=3`) that Python's `source_quality_stats`-only query never sees. Parity fixture only seeds quality-stats sources, so the blind spot passes CI. Violates the feature's hard "identical numbers" gate. |
| 2 | Medium | digest.py:500 `_count_stale_scores` | No-`job_score_staleness`-table fallback (`current_state='stale'`) differs from the TS policy-version fallback in `countOutdatedScores`; uncovered by fixture. |
| 3 | Medium | digest.py:343 `_digest_materials_ready` | Reimplements canonical `listApplyReviewQueue` + `buildApplyAudit` with count-only parity; future readiness changes silently desync the CLI. Concrete example: `??` (TS) vs `or` (Python) substage normalization flips readiness for an empty-string substage. |
| 4 | Low | test_digest_parity.py:51 | Parity asserts only `blockedSources.count`, not the `sources` array (TS/Python orderings differ); no `deleted_at`-set fixture row. |
| 5 | Low (acknowledged) | deepLinks | `followUpsDue` deep link `/jobs?applyStatus=applied` is a superset of the derived "due" count. Already noted in the PR description and matches the plan's deep-link table; tighten when precise follow-up filtering lands. |

### Verification (all pass)
- `pnpm install --frozen-lockfile` — ok
- `pnpm api:check` — ok
- `pnpm api:test` — 21 files, 352 tests pass
- `pnpm --filter @jobhunter/api exec vitest run test/digest.test.ts` — 2 pass
- `uv --project workers/automation run --extra dev pytest -q .../test_digest_parity.py` — 1 pass
- `ruff check` (digest.py, test, database.py) — All checks passed

### Gate: FAIL
One High: the cross-runtime `blockedSources.count` divergence is a latent parity break the fixture cannot catch (not a test failure — everything green). Resolve #1 (align the two source sets + add a fixture case) to reach PASS. The Mediums are drift/maintainability risks worth fixing before P4/P5 build the panel and CLI on this read model.

PR #297, opened 2026-07-05
feat(discovery): [R10] P0 politeness domain model, ports, and honest-UA policy

ebarti:
## Phase

**P0** of the crawl-politeness train — `docs/plans/2026-07-05-crawl-politeness-plan.md`. First PR in the stack; base is the plan branch `docs/plan-crawl-politeness`.

## What

Purely additive types + Protocols + TS mirror. **No fetch behavior changes and no fetch path imports the new ports** — this phase only gives the enforcement layer (P1+) its vocabulary.

- **Extend `SourcePolicy`** (Python `domain/discovery/source_registry.py` + TS mirror `packages/domain-types/src/discovery/source.ts`) with:
  - `robots_policy` (`RobotsPolicy`: `HONOR` default / `EXEMPT_DOCUMENTED_API`)
  - `min_request_interval_seconds` (per-host pacing)
  - `max_concurrent_requests_per_host` (per-host concurrency cap)
  - `max_requests_per_run` — a **distinct** outbound-request budget. `max_pages_per_run` keeps its existing result-volume meaning (not silently repurposed).
  - The fail-closed `third_party_control_bypass` invariant and its `__post_init__` guard / TS literal-`false` type are unchanged.
- **Mark documented public JSON API policies** (`WORKDAY_API_POLICY`, `ATS_API_POLICY`) as `EXEMPT_DOCUMENTED_API`, with request budgets that do not throttle documented pagination.
- **Add `ManualActionReason.ROBOTS_DISALLOWED`** and include it in the default `ManualInterventionPolicy.triggers` (triage-classified like `RATE_LIMIT`/`BOT_DETECTION`, but distinct from an error).
- **Declare the driven ports** in `domain/ports/politeness.py`: `RobotsPort`, `RateLimiterPort`, `PolitenessGatewayPort`, plus the contract value objects `HonestUserAgent` (+ `default_honest_user_agent`), `PolitenessDecision`, `RunBudget`, `RobotsVerdict`, `PolitenessOutcome`.

## Why

The declarative half of politeness exists and is fail-closed, but enforcement is absent on all 10 fetch surfaces, and live crawlers spoof a desktop browser UA. P0 establishes the shared vocabulary so P1 can build the single enforcement gateway.

## Owner decisions recorded (resolved defaults; ADR lands in P5)

[...]

PR #300, opened 2026-07-05
feat(discovery): [R10] P1 shared politeness gateway (enforcement core)

ebarti:
## Phase

**P1** of the crawl-politeness train. Stacks on **P0** (#297). Base: `feat/r10-p0-politeness-model`.

## What

The single enforcement core every outbound fetch will route through, in `infrastructure/network/` (the sanctioned cross-context home). **No fetch surface is rerouted yet** — P1 delivers the component and proves it in isolation with the two mandated fixtures. P2/P3 reroute the 10 surfaces.

- **`RobotsCache`** (`RobotsPort`) — fetch `/robots.txt` politely (short timeout, honest UA, per-host TTL cache), parse with stdlib `urllib.robotparser`, evaluate `can_fetch`. **No new dependency.**
- **`HostRateLimiter`** (`RateLimiterPort`) — process-shared, host-keyed, **thread-safe** min-interval spacing + bounded-semaphore concurrency + `Retry-After`. Shared via `get_shared_rate_limiter()` so `ThreadPoolExecutor` fan-out cannot re-introduce the parallel-mode bypass.
- **`RunBudgetCounter`** (`RunBudget`) — thread-safe per-run outbound-request counter, distinct from `max_pages_per_run`.
- **`PolitenessGateway`** (`PolitenessGatewayPort`) — `check()` is a pure verdict (peek); `guard()` is the fetch/navigation path that consumes one budget unit and holds the per-host slot. Stamps the honest UA. Robots-exempt (documented-API) sources skip robots but still pace + budget.
- **`record_politeness_outcome` + `PolitenessSession`** — robots-deny / rate-limit / budget-exhaustion recorded via the existing `record_operational_attempt_metric` with `outcome="blocked"` and `is_operational_failure`/`is_scrape_failure` both **false**. A zero-yield source proves *why* without being counted as a scrape error (RCA discipline). UI surfacing is P4.

## Owner decisions recorded (defaults per the plan)

- **D5 — robots cache TTL.** 1h for parsed robots; 5min short TTL for fail-closed (unreachable) results so they re-check.
- **D6 — unreachable robots.** `4xx` → allow (RFC 9309 unavailable). `5xx`/timeout → **fail-closed** (`UNKNOWN`, treated as disallowed, retried next run via short TTL, warned). DNS failure / connection refused → **fail-open with warning** (definitive absence; a truly-down host fails harmlessly on the content fetch).

## Design notes

- The run budget bounds **content** fetches; `robots.txt` fetches are bounded separately by per-host caching + TTL and are not double-charged, keeping the budget meaning crisp.
- `check()` never records (peek); `guard()` records blocked outcomes — avoids double-recording when a caller peeks before committing.

## Validation (loopback / stubbed only — no live board traffic, nothing spendful)

- `uv --project workers/automation run --extra dev pytest -q workers/automation/tests/test_politeness_gateway.py` -> **13 passed** (robots-deny fixture, rate-limit fixture, cache TTL, 5xx fail-closed / 404 allow / connection-refused fail-open, budget-exhaustion recording, exempt-skips-robots, honest UA)
- `uv --project workers/automation run --extra dev pytest -q` -> **1830 passed**

## Docs

None warranted in P1 (internal component, no user-facing behavior yet). Config/safety/architecture docs + ADR land in P5.


workers/automation/src/jobhunter/infrastructure/network/rate_limiter.py
```diff
@@ -0,0 +1,110 @@
...
+                    return
+                wait = earliest - now
+            self._sleep(wait)
+
+    def note_retry_after(self, host: str, retry_after_seconds: float) -> None:
+        slot = self._slot_for(host, 1)
+        with slot.lock:
+            slot.retry_after_until = max(
+                slot.retry_after_until,
+                self._clock() + max(0.0, retry_after_seconds),
```
ebarti, line 96: PR Reviewer - reviewer

Severity: Medium
Gate: Pass
Evidence: `note_retry_after` (L91-96) clamps only the lower bound (`max(0.0, ...)`) — no upper cap. `_wait_for_turn` (L79-89) then does a single `self._sleep(earliest - now)` with no ceiling. The limiter is a process-lifetime singleton (`get_shared_rate_limiter`), so `retry_after_until` persists for the whole worker process.
Reproduction: Reasoned from code; `test_rate_limiter_honors_retry_after` confirms the virtual clock advances to the full value. With real `time.sleep`, `note_retry_after(host, 31536000)` makes the next `slot(host)` sleep ~1 year while holding the bounded semaphore, freezing that host and the pooled worker thread.

A broken or hostile server can return an absurd `Retry-After` (huge seconds, or a far-future HTTP-date once callers parse dates) and stall a worker thread for the full duration with the per-host slot held. The review brief explicitly called out "cap on absurd values" for this core. This is latent in P1 (no fetch surface calls `note_retry_after` yet) and escalates to High the moment P2/P3 wire a real Retry-After.

Required fix: clamp `retry_after_seconds` to a sane maximum in `note_retry_after` (e.g. `min(value, MAX_RETRY_AFTER_SECONDS)`), and/or cap the single `_wait_for_turn` sleep and convert an over-cap wait into a `RATE_LIMITED` blocked outcome (defined in `PolitenessOutcome` but never emitted today) instead of an unbounded thread sleep. Add a test asserting the cap.

workers/automation/src/jobhunter/infrastructure/network/robots.py
```diff
@@ -0,0 +1,148 @@
...
+        assert entry.parser is not None  # override is None => parser present
+        return RobotsVerdict.ALLOW if entry.parser.can_fetch(user_agent, url) else RobotsVerdict.DISALLOW
+
+    def _entry_for(self, host_key: str, user_agent: str) -> _RobotsEntry:
+        now = self._clock()
+        with self._lock:
+            cached = self._cache.get(host_key)
+            if cached is not None and cached.expires_at > now:
+                return cached
+        entry = self._fetch(host_key, user_agent)
```
ebarti, line 91: PR Reviewer - reviewer

Severity: Low
Gate: Pass
Evidence: `_entry_for` (L85-91) releases `self._lock` before calling `self._fetch` (the network GET), and the robots fetch runs OUTSIDE the per-host rate-limit slot (`PolitenessGateway._robots_block` runs before `self._rate_limiter.slot` in `guard()`). So N threads entering the gateway for the same uncached host concurrently each issue a simultaneous `GET /robots.txt`.
Reproduction: Reasoned from code. Bounded: first-touch only; once the result caches (1h TTL) subsequent calls hit cache.

Small hole in the train's per-host-concurrency promise, specific to the robots endpoint, at run start when multiple threads target one host.

Required fix (hardening, not merge-blocking): add a per-host single-flight lock so only one thread fetches a given host's robots.txt while others await the result, or route the robots fetch through the rate limiter.

workers/automation/src/jobhunter/infrastructure/network/robots.py
```diff
@@ -0,0 +1,148 @@
...
+
+    def evaluate(self, url: str, user_agent: str) -> RobotsVerdict:
+        host_key = self._host_key(url)
+        if host_key is None:
+            return RobotsVerdict.UNKNOWN
+        entry = self._entry_for(host_key, user_agent)
+        if entry.verdict_override is not None:
+            return entry.verdict_override
+        assert entry.parser is not None  # override is None => parser present
+        return RobotsVerdict.ALLOW if entry.parser.can_fetch(user_agent, url) else RobotsVerdict.DISALLOW
```
ebarti, line 83: PR Reviewer - reviewer

Severity: Low
Gate: Pass
Evidence: `evaluate` (L83) delegates to stdlib `RobotFileParser.can_fetch`. Confirmed empirically on the project runtime (CPython 3.14.4): robots `Disallow: /` then `Allow: /jobs` returns `can_fetch('/jobs/1') = False`. stdlib uses order-based first-match, NOT RFC 9309 longest-match (most-specific wins).
Reproduction: Reproduced with a standalone `RobotFileParser` snippet under `uv run python` (3.14.4).

Consequence: exception-carving robots files (`Disallow: /` then `Allow: /public`) are over-blocked. This errs toward politeness (safe direction) and each affected source records a visible `robots_disallowed` outcome, so it is observable rather than a silent wrong-direction allow. But a legitimately-allowed board could yield nothing.

Required fix: none required for merge; document the stdlib simplification in a comment here and in the P5 config/architecture docs so the owner knows Allow-exception boards may need the EXEMPT path or a stricter parser later.

workers/automation/src/jobhunter/infrastructure/network/politeness.py
```diff
@@ -0,0 +1,285 @@
...
+    ) -> None:
+        self._gateway = gateway
+        self._policy = policy
+        self._budget = budget
+        self._context = context
+        self._recorder_conn = recorder_conn
+
+    @property
+    def user_agent(self) -> str:
+        return self._gateway.user_agent  # type: ignore[attr-defined]
```
ebarti, line 257: PR Reviewer - reviewer

Severity: Low
Gate: Pass
Evidence: `PolitenessSession.user_agent` (L257) reads `self._gateway.user_agent` behind `# type: ignore[attr-defined]`, and `note_retry_after` (L272-273) reaches the gateway via a defensive `getattr`, because `PolitenessGatewayPort` (`domain/ports/politeness.py`) declares neither `user_agent` nor `note_retry_after`. P2/P3 callers need both (stamp the honest UA; forward server Retry-After).

Required fix (maintainability, not merge-blocking): declare `user_agent` (property) and `note_retry_after` on `PolitenessGatewayPort` so the contract is honest and the type-ignore/getattr workarounds go away.

ebarti (commented):
PR Reviewer - reviewer (COMMENT only — not an approval or change request)

Gate: PASS — Blocker=0, High=0. Reviewed PR #300 head 12006390 vs #297 branch `feat/r10-p0-politeness-model` (merge-base = #297 HEAD e09ad4c0, clean stack). Verification: 13 gateway tests + full `pytest -q -p no:randomly` = 1830 passed; `ruff check` clean.

Verified correct:
- D6 unreachable-robots branch table implemented EXACTLY as recorded (per-flavor table below).
- Robots cache keyed by scheme+netloc; TTL 1h success / 5min unreachable (D5). Honest UA stamped and passed to `can_fetch`; no browser impersonation.
- Recorder writes `outcome="blocked"`, `failure_category` in {robots_disallowed, budget_exhausted}, `is_scrape_failure=0`, `is_operational_failure=0`.
- Budget enforced before the request via atomic `try_consume`; exhaustion recorded as its own non-error outcome.
- Rate limiter process-shared, thread-safe (BoundedSemaphore + per-host lock), paces by min-interval, no busy-wait.

D6 per-flavor verdict (`robots.py` `_fetch`):
- 4xx incl 404 -> allow (normal TTL). CORRECT.
- 2xx -> parse + evaluate. CORRECT.
- 5xx -> UNKNOWN fail-closed, 5min TTL, warned; gateway maps UNKNOWN -> ROBOTS_DISALLOWED. CORRECT.
- timeout (direct or URLError-wrapped) -> UNKNOWN fail-closed, warned. CORRECT.
- DNS failure / connection refused -> allow, warned (fail-open). CORRECT.
- any other exception -> fail-closed. CORRECT (conservative).

consecutiveFailures interplay answer (P4 foundation):
- Primary path (`source_quality_stats` exists): `listSourceHealth` reads `consecutive_failures` straight from the projection, which is computed purely from `DiscoveryRunCompleted`/`DiscoveryRunFailed` domain events (`projections.ts`) — it never reads `operational_attempt_metrics`, so blocked rows cannot touch it or the recommended-state gate (>=3 quarantined / >=5 disabled).
- Fallback path (`sourceRollupToHealth`): every failure counter is gated by `isFailureMetric` (`read-model.ts` L3753 = outcome in {failed, partial_failed}); a blocked row is not a failure, so failures/consecutive/operational/scrape/retryable all exclude it, and lastOutcome="blocked" yields consecutiveFailures=0.
- Net: P4 can rely on blocked rows NOT poisoning health. Two caveats for P4: in the fallback path `runCount`(=attempts) does include blocked rows; and `PolitenessOutcome.RATE_LIMITED` is never emitted in P1 (limiter enforces by waiting), so no `rate_limited` rows exist yet.

Findings (all Medium/Low; none block the gate) — see inline:
1. Medium — Retry-After has no upper cap; unbounded `_wait_for_turn` sleep can freeze a host + worker thread for the full attacker-chosen duration. Latent in P1, High once P2/P3 wire real headers.
2. Low — robots.txt fetch bypasses the limiter and has no single-flight; concurrent first-touch bursts a host's /robots.txt.
3. Low — stdlib robotparser is order-based first-match, not RFC 9309 longest-match; over-blocks Allow-exception boards (safe direction). Document it.
4. Low — `PolitenessGatewayPort` omits `user_agent`/`note_retry_after` that `PolitenessSession` relies on.

Scope: minimal diff, correct base, [R10] title, no fetch surface rerouted (P1 component in isolation) — matches the phase goal. No secrets/PII. COMMENT-only from me.

workers/automation/src/jobhunter/infrastructure/network/rate_limiter.py
ebarti: Addressed in the active train: **#314 @ 37561813** (P3b). Clamped at the sink (`HostRateLimiter.note_retry_after` + a defensive per-nap ceiling in `_wait_for_turn`) to `max_retry_after_seconds` (default 300s), logging when it clamps — so every current and future caller is covered, not per-call-site. Over-clamp hosts are flagged and the gateway's `guard` records a `RATE_LIMITED` outcome and skips (via `hard_rate_limit_remaining`) rather than parking a worker for the clamped cooldown. Tests: absurd-header clamp, within-cap pacing vs hard-skip, cooldown expiry, single-nap ceiling, gateway skip-without-slot-or-budget, and an end-to-end `GatewayHttpClient` absurd-429 that clamps + records + pre-empts the next fetch.

Note re "`RATE_LIMITED` never emitted": it's already emitted on every 429/503 via `PolitenessSession.record_server_rate_limit` (landed in #303), so P4's counter is non-zero today; #314 adds the gateway pre-emption emission point and bounds the freeze.

PR #303, opened 2026-07-06
feat(discovery): [R10] P2 route ATS adapters through the politeness gateway

ebarti:
## Phase

**P2**, part 1 of the non-browser routing — surface **#3 (ATS board APIs)**. Stacks on **P1** (#300). Base: `feat/r10-p1-gateway`.

P2 is split by surface for reviewability (the plan enumerates the surfaces separately and marks them internally parallelizable): this PR routes ATS; follow-on stacked PRs route compensation (#9), Workday (#2), and gate the jobspy boundary (#1).

## What

Rip-and-replace: the ATS adapters no longer build their own `urllib` transport.

- **`GatewayHttpClient`** (`infrastructure/network/http_client.py`) — the single `urllib` transport for non-browser surfaces. Every request is wrapped in a `PolitenessSession` guard (robots + per-host rate/concurrency + per-run budget + honest UA). A **blocked decision returns `None`** (recorded as a first-class outcome by the session, never raised as a scrape error). A server **429/503** is honored as a rate-limit outcome (`Retry-After` fed back to the limiter) rather than surfaced as a failure.
- **Delete `default_http_fetcher`** and its spoofed browser UA from `ats_adapters.py`; make the injected `http` fetcher **required**. The composition root (`run_scheduled_ats_sources` → `_adapter_for_source`) builds a **per-source** `GatewayHttpClient` from the source's `SourcePolicy`. ATS API sources are robots-exempt (D2) but still paced, budgeted, and honestly identified. Tests may still inject `http` to bypass the network.
- `PolitenessSession.record_server_rate_limit` for server-issued 429/503.

## Enforcement proof (the defect class this train kills)

- `test_fetch_surface_enforcement.py` — AST assertion that `ats_adapters.py` imports **no raw `urllib` transport**, and that `_adapter_for_source` with no injected `http` builds a `GatewayHttpClient`-backed fetcher. This file grows one surface at a time as each follow-on PR reroutes it.

## Validation

- `uv --project workers/automation run --extra dev pytest -q` -> **1838 passed** (+8)
- Targeted: `pytest -q test_gateway_http_client.py test_fetch_surface_enforcement.py test_ats_adapters.py test_discovery_production_wiring.py` -> **45 passed**

No live board traffic: fixtures use stub openers / injected fetchers.

## Docs

Deferred to P5 (config reference, safety notes, architecture, ADR) with the phases that finalize user-facing config. This PR is internal routing with no new user-facing surface.


workers/automation/tests/test_fetch_surface_enforcement.py
```diff
@@ -0,0 +1,67 @@
...
+    tree = ast.parse(source)
+    offenders: list[str] = []
+    for node in ast.walk(tree):
+        if isinstance(node, ast.Import):
+            for alias in node.names:
+                if alias.name.split(".")[0] == "urllib" and alias.name != "urllib.parse":
+                    offenders.append(alias.name)
+        elif isinstance(node, ast.ImportFrom):
+            if (node.module or "").split(".")[0] == "urllib" and node.module != "urllib.parse":
+                offenders.append(node.module or "")
```
ebarti, line 38: PR Reviewer - pr-reviewer

Severity: Medium
Gate: Fail
Evidence: `_imports_raw_http` only flags imports whose top-level package is `urllib` (lines 34, 37). Running this exact function against injected snippets: it catches `import urllib.request` inside a function body and aliased `import urllib.request as _u` (both -> flagged), but returns `[]` for `import http.client`, `import socket`, `import requests`, `import httpx`, and `from urllib3 import PoolManager`. The assert on line 46 claims "still imports raw HTTP transport" -- broader than what is actually checked.
Reproduction: Reproduced with python3 running the function body verbatim against the six snippets.

This is the sole automated regression guard for the whole R10 "no ad-hoc transport" invariant. It does correctly kill the specific defect the deleted `default_http_fetcher` embodied (raw `urllib`), including lazily-imported/aliased re-adds. But a future edit to any routed surface could reintroduce an ad-hoc transport via stdlib `http.client`/`socket` (both importable with zero new deps) and this test would still pass green, defeating the guard for the exact class it advertises.

Required fix: Broaden the AST check to also flag `http.client`, `socket`, and the known third-party HTTP libs (`requests`, `httpx`, `urllib3`, `aiohttp`) -- or invert it to an allowlist of permitted transport imports -- and add a self-test proving a raw non-urllib transport is detected.

workers/automation/src/jobhunter/infrastructure/network/http_client.py
```diff
@@ -0,0 +1,128 @@
...
+                with self._opener.open(request, timeout=timeout or self._default_timeout) as response:
+                    return response.read()
+            except urllib.error.HTTPError as exc:
+                if exc.code in (429, 503):
+                    retry_after = parse_retry_after(
+                        exc.headers.get("Retry-After") if exc.headers else None
+                    )
+                    if retry_after:
+                        self._session.note_retry_after(url, retry_after)
+                    self._session.record_server_rate_limit(url, retry_after)
```
ebarti, line 126: PR Reviewer - pr-reviewer

Severity: Low
Gate: Pass
Evidence: On 429/503 the server's `Retry-After` is parsed and fed straight into the shared per-host limiter via `note_retry_after` (line 125) with no upper bound. `HostRateLimiter._wait_for_turn` (rate_limiter.py:79-89) sleeps `earliest - now` where `earliest = max(next_start_allowed, retry_after_until)`, so a large value delays the next slot for that host for its full duration. Because the limiter is per-host and shared across sources, one Greenhouse board returning e.g. `Retry-After: 999999999` stalls every `boards-api.greenhouse.io` source's next request for the rest of the run.
Reproduction: Reasoned from code (parse_retry_after -> note_retry_after -> _wait_for_turn sleep).

Honoring server backoff is intended and the sleep mechanism itself landed in P1; this PR is the first caller to feed server-controlled data into it, which makes the missing ceiling newly reachable. Impact is self-throttling only (no security/data risk) and the sources are trusted documented APIs -- hence Low.

Required fix: Clamp `retry_after` to a sane maximum (e.g. a few minutes) before calling `note_retry_after`, so a malformed/hostile header cannot park a host for an unbounded time.

workers/automation/src/jobhunter/infrastructure/network/http_client.py
```diff
@@ -0,0 +1,128 @@
...
+    ) -> str | None:
+        raw = self._request(
+            url,
+            method="GET",
+            json_body=None,
+            timeout=timeout,
+            accept="application/json, text/csv;q=0.9, */*;q=0.8",
+            extra_headers=extra_headers,
+        )
+        return None if raw is None else raw.decode("utf-8")
```
ebarti, line 93: PR Reviewer - pr-reviewer

Severity: Low
Gate: Pass
Evidence: `GatewayHttpClient.fetch_text` is added here but has no caller in this PR (`grep -rn fetch_text workers/automation` shows only this definition; the compensation surface still uses its own raw `_fetch_text` in `sqlite_market_repository.py:675`, rerouted in a later PR) and no test in `test_gateway_http_client.py` (which covers only `fetch_json` + `parse_retry_after`). Its distinct `Accept` header and its None-on-block decode path are unexercised.
Reproduction: Reproduced with grep (no callers, no tests).

Not a correctness problem for P2, but shipping an untested transport method ahead of its consumer risks a latent bug that surfaces only when a later PR wires it.

Required fix: Either add a focused test for `fetch_text` (allowed decode + blocked -> None) now, or defer the method to the PR that first calls it.

ebarti (commented):
PR Reviewer - pr-reviewer -- R10 P2 (route ATS adapters through the politeness gateway)

Gate: PASS (Blocker=0, High=0). Findings: Medium=1, Low=2.
Reviewed head 188681337 vs base feat/r10-p1-gateway (merge-base == base tip; clean stack).

Bypass audit (priority 1): PASS. ats_adapters.py imports no raw transport (json / urllib.error / urllib.request all removed; only logging/html/bs4/domain imports remain). default_http_fetcher and its spoofed browser UA are deleted with zero lingering references repo-wide. Every adapter fetch flows through the injected HttpFetcher, built as a per-source GatewayHttpClient.fetch_json at the composition root. Enforcement-test strength: it catches a re-added `import urllib.request` inside a function body and aliased imports (empirically verified) but misses non-urllib transports -- see the Medium inline.

Behavior preservation (priority 2): PASS. Parsing/filtering/pagination unchanged; timeout preserved (20s default, adapters don't override); JSON decode identical. Intended changes only: spoofed UA -> honest UA, and 429/503 -> None (graceful rate-limit outcome) instead of a raised scrape error. All four adapters treat None as empty (`or {}` / `or []`), so a block yields empty results, never a crash.

Injection correctness (priority 3): PASS. Per-source PolitenessSession => per-source budget isolation; gateway (hence rate limiter + robots cache) shared across the run. `getattr(source, "policy", None) or ATS_API_POLICY`: ScheduledSource has no .policy, so ATS sources get ATS_API_POLICY (robots-EXEMPT per D2, 1s interval, 2000-req budget). That fallback can only reach ATS sources because _gateway_ats_fetcher runs only after the `ats_kind in {GREENHOUSE,LEVER,ASHBY}` guard. Recording dims read .source_kind/.priority, which ScheduledSource does expose.

Recording accuracy (priority 4): PASS. guard()._record records only blocked (non-allowed) decisions; a 429 follows an allowed decision, so record_server_rate_limit emits exactly one rate_limited row with no double-count. Outcomes carry the correct source_id/run_id/adapter. Budget exhaustion keeps prior sources' results and records budget_exhausted (no explosion).

Tests (priority 5): `pytest -q -p no:randomly` => 1838 passed; `ruff check .` => All checks passed. New unit tests stub the opener (no live network). Gaps noted inline (enforcement scope; fetch_text untested; production http=None path covered by the enforcement + unit seams rather than an end-to-end wiring test).

Conventions (priority 6): PASS. [R10] conventional title, base correct, diff minimal/scoped to ATS.

Note: the code-review / security-review / archscan skills were not invocable by this agent in this environment (no Skill tool exposed); I applied the equivalent review lenses manually.

workers/automation/src/jobhunter/infrastructure/network/http_client.py
ebarti: Addressed in the active train: **#314 @ 37561813** (P3b). Clamped at the sink (`HostRateLimiter.note_retry_after` + a defensive per-nap ceiling in `_wait_for_turn`) to `max_retry_after_seconds` (default 300s), logging when it clamps — so every current and future caller is covered, not per-call-site. Over-clamp hosts are flagged and the gateway's `guard` records a `RATE_LIMITED` outcome and skips (via `hard_rate_limit_remaining`) rather than parking a worker for the clamped cooldown. Tests: absurd-header clamp, within-cap pacing vs hard-skip, cooldown expiry, single-nap ceiling, gateway skip-without-slot-or-budget, and an end-to-end `GatewayHttpClient` absurd-429 that clamps + records + pre-empts the next fetch.

Note re "`RATE_LIMITED` never emitted": it's already emitted on every 429/503 via `PolitenessSession.record_server_rate_limit` (landed in #303), so P4's counter is non-zero today; #314 adds the gateway pre-emption emission point and bounds the freeze.

PR #301, opened 2026-07-05
feat(discovery): [R9] per-family streaming fan-out (Phase 1)

ebarti:
## What (R9 Phase 1 — score-as-you-discover)

`DiscoverWorkflow` now drains enrichment and fans out per-job preparation **after each family completes**, instead of once at the very end. A job discovered by the first family is scored while later families are still crawling — Time To First Score (TTFS) drops from "after the whole run" to "after the first family".

Stacked on the plan branch `docs/plan-streaming-latency` (PR #260). Delivers **Phase 1** of `docs/plans/2026-07-05-streaming-pipeline-latency-plan.md`.

## Why

Today a job discovered early cannot be scored until the entire discovery+enrichment phase finishes: families run sequentially, then one global enrichment, then one end-of-run fan-out. Early jobs sit unscored. This streams scoring while preserving every safety/correctness invariant.

## How it works

- **Per-family streaming + terminal reconcile (plan option b).** After each *completed* family: `discovery_enrichment` (drains that family's fresh jobs) + `discovery_preparation_fanout` (starts their `JobPreparationWorkflow`s). A **terminal reconcile** enrichment + fan-out still runs after the loop and stays authoritative for the tolerated-partial-failure folding and progress finalization. Streaming passes are additive/best-effort: any non-cancellation failure is left for the terminal pass to sweep up (deduped by the deterministic prep id). This keeps the existing folding + progress semantics essentially unchanged.
- **Progress model (decision recorded).** Fixed denominator `len(families) + 2`, known at plan time; the counter is monotonic on the family + terminal spine. Streaming passes are **progress-silent** (`progress_total=0`) so the Runs bar never oscillates or shrinks. Incremental scores reach the UI through the independent `JobScored` → projections → SSE path, not the progress bar. No new `discovery_runs` columns → projection parity preserved.
- **Race-free repeated fan-out (decision recorded).** `prep-{idempotency_key}` + `USE_EXISTING` make N invocations start exactly one workflow per job. Because the idempotency key includes `kind`, a fresh job crossing `pending_score` → `pending_tailor` mid-tailor would otherwise get a second racing `TAILOR_RESUME` workflow. New `include_pending_tailor` flag: the **first** fan-out sweeps pre-existing `pending_tailor` stragglers once (the only moment `pending_tailor` can't hold a this-run `SCORE_JOB` job); every later pass is score-only. No double-tailor, no double spend (I1/I4).

## Invariants (all hold)

[...]

PR #306, opened 2026-07-06
feat(discovery): [R9] per-job preparation handoff (Phase 2)

ebarti:
## What (R9 Phase 2 — per-job handoff)

Tightens score-as-you-discover from **per-family** (Phase 1) to **per-job**. As each job is individually enriched (committed to `pending_score`), the enrichment worker starts that job's `SCORE_JOB` preparation workflow **immediately** — before its siblings in the same family are even scraped — so Time To First Score drops to per-job granularity.

Stacked on **#301** (Phase 1). Delivers **Phase 2** of `docs/plans/2026-07-05-streaming-pipeline-latency-plan.md`.

## Mechanism (decision recorded)

Temporal-native, event-driven, per the coordinator's "prefer Temporal-native over ad-hoc polling" guidance and the plan's Phase 2 option. An opaque `on_job_enriched` callback is threaded from the enrichment activity to `enrichment/detail.py`'s per-job seam (`run_discovery_enrichment_stage` → `_run_discovery_enrichment_until_idle` → `_run_enrich` → `run_enrichment` → `_run_detail_scraper` → `scrape_site_batch`) and fired after each job's commit. The start is a **side effect inside the enrichment activity**, so `DiscoverWorkflow`'s command history is unchanged.

## Guarantees (identical to Phase 1)

- **Determinism/replay (I3):** handoff is inside the activity → workflow history unchanged.
- **Idempotence (I1):** per-job start uses the same deterministic `prep-{idempotency_key}` `SCORE_JOB` id as the fan-out, so the handoff and the per-family/terminal reconciling fan-outs converge on exactly one execution per job via `USE_EXISTING`. A re-enrichment that changes `source_event_id` legitimately forks a new workflow (material change only).
- **Spend (I4):** per-job `JobPreparationWorkflow` preflight unchanged (Phase 1 budget fixture covers it).
- **Safety:** per-job starts are serialized by a lock (site enrichment can run in parallel threads when `workers>1`) and are best-effort — a start failure is logged and left for the fan-out backstop, never mistaken for an enrichment failure.

## New surface

`per_job_handoff` flag + prep params on `DiscoveryEnrichmentActivityInput` (enabled on streaming enrichment passes, off for the terminal reconcile), a `start_job_preparation_workflow` single-job starter in `pipeline/preparation.py`, and `_build_per_job_handoff` in the activity.

## Validation

```
uv --project workers/automation run --extra dev pytest -q      # 1815 passed
uv --project workers/automation run --extra dev ruff check .    # clean
git diff --check                                                # clean
```

Fixtures:
- **Promptness** — `test_scrape_site_batch_hands_off_each_job_as_it_is_enriched` asserts each job is handed off right after it is enriched, before the next sibling is scraped (a failed job produces no handoff). Plus `test_scrape_site_batch_handoff_error_does_not_break_enrichment` (isolation).
- **Idempotence at per-job granularity** — `test_per_job_handoff_id_converges_with_fanout_and_forks_on_reenrichment` (same id as the fan-out; re-enrichment forks).
- **Wiring** — `test_build_per_job_handoff_starts_scored_prep_with_params` + `test_build_per_job_handoff_disabled_returns_none`.
- **Budget (I4)** — covered by Phase 1's `test_preparation_workflow_fails_fast_when_budget_exceeded_and_spends_nothing` (the per-job handoff starts the same `JobPreparationWorkflow`).

Existing enrichment/observability/queue-selector test fakes were updated to accept the new opaque `on_job_enriched` kwarg (signature-only; no assertions weakened). Python + docs only; TS/read-model untouched. `pnpm docs:build` not run locally (fresh worktree, no node deps); no new doc anchors were introduced in this PR.

## Docs

`docs/architecture/pipeline/index.md`, `concurrency.md` (per-job handoff bullet), `operations.md` (per-job score granularity note), `docs/decisions.md` (ADR Phase 2 mechanism), `docs/local-reliability-qa.md` (Phase 2 fixture coverage).

## Retarget note

Retarget to `main` (or to the merged Phase 1 branch's target) before merge if Phase 1 lands first.


ebarti:
Follow-up fix commit `a2db2035`: Phase 2's per-job handoff scores jobs the instant they're enriched, which exposed a double-tailor race in the Phase 1 "first fan-out does the full derive" approach — a fresh job already in `pending_tailor` would be re-derived as a second racing `TAILOR_RESUME` workflow. Fixed by moving the one-time `pending_tailor` straggler sweep to **before** the family loop (the only moment `pending_tailor` holds only pre-existing work) and making every family/terminal fan-out score-only. New fixture `test_only_the_preloop_sweep_derives_pending_tailor` pins the invariant (`include_pending_tailor` sequence `[True, False, False, False, False]`). Also sets up Phase 3 concurrency correctly.

workers/automation/src/jobhunter/discovery/activities.py
```diff
@@ -220,6 +236,43 @@ async def discovery_enrichment_activity(
...
+        return None
+
+    from jobhunter.domain.tenant import TenantId
+    from jobhunter.pipeline.preparation import start_job_preparation_workflow
+
+    handoff_lock = threading.Lock()
+
+    def _handoff(job_url: str) -> None:
+        with handoff_lock:
+            start_job_preparation_workflow(
```
ebarti, line 261: PR Reviewer - pr-reviewer

Severity: Medium
Gate: Pass
Evidence: `_handoff` holds a shared `handoff_lock` around `start_job_preparation_workflow(...)` (discovery/activities.py:260-261). That call chains `_run_start_batch` -> `_run_coroutine` -> `asyncio.run` -> `default_workflow_starter`, which opens a brand-new Temporal client every time (`get_temporal_client()` in infrastructure/temporal/client.py calls `Client.connect(...)` with no cache) and then runs `_record_dispatch_started` -> `ProjectionBuilder(...).refresh()` (infrastructure/rpc/workflow_starter.py:112). So every enriched job pays a fresh TCP+TLS+namespace-describe round-trip plus an incremental projection refresh, and under `workers>1` the parallel `_run_detail_scraper` threads all serialize on `handoff_lock` for each start.
Reproduction: Reasoned from code (call chain traced end to end; not executed against a live Temporal).

The per-job handoff is on the enrichment hot path and its whole purpose is lower latency, but doing a fresh connect + projection refresh under a global lock once per job partially works against that goal and can throttle enrichment throughput when `workers>1`. This is a performance/scalability concern, not a correctness bug — the workflow_starter docstring already accepts the per-call reconnect as "a few ms against localhost" for the single-user local case, so it is acceptable to merge.

Required fix: None required for merge. Consider a follow-up that reuses one client/event-loop for handoff starts (or drains starts from a bounded queue off the scrape thread) so per-job starts do not serialize a fresh connect + full refresh on the hot path.

workers/automation/src/jobhunter/enrichment/detail.py
```diff
@@ -1019,6 +1026,22 @@ def scrape_site_batch(
...
+                    # mistaken for an enrichment failure below.
+                    if (
+                        on_job_enriched is not None
+                        and status in ("ok", "partial")
+                        and desc_len > 0
+                    ):
+                        try:
+                            on_job_enriched(url)
+                        except Exception:  # noqa: BLE001 - handoff is best-effort
+                            log.warning(
```
ebarti, line 1042: PR Reviewer - pr-reviewer

Severity: Low
Gate: Pass
Evidence: A per-job handoff failure is caught and surfaced only via `log.warning(..., exc_info=True)` (enrichment/detail.py:1042-1043). Unlike the abandoned-thread path in infrastructure/temporal/run_in_activity.py (which calls `record_operational_attempt_metric`), a persistently failing handoff (e.g., Temporal unreachable) leaves no `job_events` row or operational metric, so it is invisible in the Runs/registry surfaces and appears only in worker logs.
Reproduction: Reasoned from code; `test_scrape_site_batch_handoff_error_does_not_break_enrichment` confirms the failure is caught but asserts nothing about it being recorded anywhere.

This is correct as best-effort behavior — the failure is genuinely non-fatal, is backstopped by the score-only fan-out + terminal reconcile, and is NOT swallowed silently ("failed: failed"), so it does not gate the merge. Flagging only because R9 explicitly wants handoff failures visible in the failure taxonomy rather than logs alone.

Required fix: None required for merge. Optional follow-up: record an `operational_attempt_metric` (or a job event) on handoff-start failure for observability parity with the abandoned-thread path, so repeated handoff failures are diagnosable without log access.

workers/automation/tests/test_discovery_preparation_orchestration.py
```diff
@@ -182,6 +182,51 @@ def test_streaming_fanout_dedups_repeated_passes_via_deterministic_ids(
...
+    monkeypatch: pytest.MonkeyPatch,
+) -> None:
+    """R9 Phase 2 idempotence (I1) at per-job granularity: the per-job handoff
+    starts the SAME deterministic id a SCORE_JOB fan-out would derive for the
+    job, so `USE_EXISTING` collapses the handoff + reconciling fan-out into one
+    execution. A re-enrichment that changes ``source_event_id`` — a genuine
+    material change — legitimately forks a new prep workflow."""
+    monkeypatch.setattr(preparation, "get_connection", lambda: conn)
+    monkeypatch.setattr(preparation, "current_scoring_policy_version", lambda *_a, **_k: 4)
+    monkeypatch.setattr(preparation, "_latest_source_event_id", lambda _conn, _url: "event-A")
```
ebarti, line 196: PR Reviewer - pr-reviewer

Severity: Low
Gate: Pass
Evidence: `test_per_job_handoff_id_converges_with_fanout_and_forks_on_reenrichment` monkeypatches `_latest_source_event_id` to a constant ("event-A") for both the handoff start and the `build_preparation_workflow_spec` fan-out comparison (lines 196, 225). It therefore proves "given an identical source event, handoff id == fan-out id; a different source event forks" and "a repeated handoff for the same source event reuses the id" — which is exactly what it names.
Reproduction: Verified by reading the fixture; targeted suite green (`pytest test_discovery_preparation_orchestration.py` passes).

The load-bearing part of the I1 convergence guarantee — that a still-`pending_score` job's latest source event does not drift between the handoff (computed at enrichment commit) and a later fan-out — is argued in the PR body but not exercised here, since the source event is stubbed. I separately confirmed by code that re-observation of an existing job publishes `JobSourceObserved` (not in `_latest_source_event_id`'s match set), so multi-family re-discovery does NOT drift the key and only a genuine posting-content change forks — so the mechanism is sound. This is a test-realism gap, not a defect; it does not gate the merge.

Required fix: None required for merge. Optional: add a fixture that writes real `job_events` (a discovered+enriched job, then re-observed by a second source) and asserts the derived `prep-` id is stable across the handoff and a subsequent score-only fan-out, so the "no drift for pending_score" half of I1 is pinned rather than stubbed.

ebarti (commented):
PR Reviewer - pr-reviewer — R9 Phase 2 (per-job handoff)

Gate: PASS. No Blocker/High. 1 Medium + 2 Low, all inline, none blocking.

Base verified: `feat/r9-phase2` @ a2db2035 sits directly on `feat/r9-phase1` (merge-base == phase1 tip); only the two expected commits, no foreign history.

What I verified (all pass):
- Determinism/replay (I3): the handoff is a side effect INSIDE the enrichment activity (`on_job_enriched` fired post-commit in `scrape_site_batch`), so `DiscoverWorkflow` command history is unchanged.
- Async/heartbeat discipline: the callback runs in the worker `ThreadPoolExecutor` thread (`run_blocking_with_heartbeat` offloads the sync body), so `_run_coroutine` correctly takes the `asyncio.run` branch and the heartbeat loop on the event loop is never blocked. A raising handoff is caught and cannot fail enrichment (pinned by `test_scrape_site_batch_handoff_error_does_not_break_enrichment`).
- Spend (I4): the handoff starts the same `JobPreparationWorkflow`, which runs `_check_spend` before any spending step (`preparation/workflow.py:91`), so a job enriched after the ceiling is hit starts a workflow that fails fast and spends nothing. Ceiling applies mid-stream.
- Promptness (I named-invariant): `test_scrape_site_batch_hands_off_each_job_as_it_is_enriched` pins the exact interleave (handoff fires before the next sibling is scraped; a failed job produces no handoff).

Retry-refire adjudication (coordinator priority #1): NOT a High. Enrichment commits per job and the enrichment selection only re-picks jobs lacking a description, so on an activity retry `on_job_enriched` does NOT refire for already-enriched jobs. Even if it did, the deterministic `prep-{idempotency_key}` id + `WorkflowIDConflictPolicy.USE_EXISTING` collapses duplicate concurrent starts to one execution; `test_per_job_handoff_id_converges_with_fanout_and_forks_on_reenrichment` pins both the repeated-handoff-same-id (dedup precondition) and the material-change fork. Residual (Low, inline): no end-to-end activity-retry simulation, and the convergence fixture stubs `_latest_source_event_id`.

Double-tailor race fix (a2db2035) — placement adjudication (coordinator priority #3): the fix BELONGS in #306 and must NOT be backported to #301. The race is Phase-2-specific: per-job handoff starts SCORE_JOB workflows DURING enrichment, so a fresh job can reach `pending_tailor` before the first family fan-out; under the old "first-fan-out does the full derive" it would be re-derived as a racing TAILOR_RESUME. In #301 standalone, fan-outs are the ONLY starters, so at the first (full-derive) fan-out no this-run job is yet in `pending_tailor` — Phase 1 is safe on its own. #306 correctly moves the one-time `include_pending_tailor=True` sweep BEFORE the family loop and makes every family/terminal fan-out score-only. `test_only_the_preloop_sweep_derives_pending_tailor` pins the flag sequence `[True, False, False, False, False]`, which genuinely fails under the old 4-flag arrangement. So #301 stays standalone-mergeable without this commit.

Verification:
- `uv --project workers/automation run --extra dev pytest -q` -> 1816 passed.
- `uv --project workers/automation run --extra dev ruff check .` -> clean.
- `git diff --check` -> clean.
- targeted: `pytest test_workflow_discovery.py test_discover_reliability.py test_discovery_preparation_orchestration.py` -> 67 passed.
- `pnpm install --frozen-lockfile` + `pnpm docs:build` -> build complete; "link check passed: 4547 references resolve across 230 emitted files" (implementer could not run this; now confirmed green).

Docs accurately reflect per-job handoff (pipeline index/concurrency/operations, decisions ADR §Phase 2, reliability-qa matrix row). Security/architecture: no new external-input or secret surface; the discovery-activity -> pipeline/preparation call and lazy import match the existing fan-out pattern (no new boundary violation). Minor doc nit (non-gating): the reliability-qa row cites the per-job handoff fixtures but not `test_only_the_preloop_sweep_derives_pending_tailor`, the load-bearing race-fix test.

PR #310, opened 2026-07-06
feat(discovery): [R10] P2 gate the jobspy invocation boundary (budget + pacing)

ebarti:
## Phase

**P2**, part 4 (final non-browser routing) — surface **#1 (broad-board aggregator, python-jobspy)**. Stacks on **P2c-workday** (#308). Base: `feat/r10-p2c-workday`.

## What

`python-jobspy` owns its `tls-client`/`requests` transport, so per-request robots and honest-UA **cannot** be enforced inside it (owner decision **D3**). We enforce politeness at **our call boundary** instead:

- **`_full_crawl`**: a per-run request budget (`BROAD_BOARD_LEAD_POLICY`) is consumed per search, and inter-search pacing runs through a per-crawl host limiter (the crawl loop is sequential; the process-shared limiter guards the parallel fan-out paths in workday/enrichment). When the budget stops a crawl, a **budget-exhausted** outcome is recorded via `record_operational_attempt_metric` as a first-class non-error (`source_id="jobspy"`, `is_scrape_failure=0`).
- **`search_jobs`** (manual single search) also paces via the shared limiter.
- **Documented residual (D3)**: jobspy's internal per-board requests remain unpoliced; broad boards stay opt-in / scheduler-gated upstream. Called out in a code comment and here.

## Enforcement proof

- `test_jobspy_gateway_boundary.py`: `_full_crawl` with a 2-request budget stops after 2 searches and records a `budget_exhausted` non-error outcome.

## Validation

- `uv --project workers/automation run --extra dev pytest -q -p no:randomly` -> **1844 passed**
- A pre-existing random-order flake in `test_llm_client` (real backoff sleep; unrelated to R10 — `llm.py` untouched) passes in isolation and under deterministic ordering.

## Non-browser routing complete (surfaces #1, #2, #3, #9)

With P2/P2b/P2c/P2d, every non-browser fetch surface routes through the gateway or is gated at its boundary. Browser surfaces (#4–#8, #10) are P3.

## Docs

Deferred to P5 (safety notes incl. the jobspy residual, architecture, ADR).


workers/automation/src/jobhunter/discovery/jobspy.py
```diff
@@ -1106,6 +1132,22 @@ def _full_crawl(
     # Ensure DB schema is ready
     init_db()
 
+    # Politeness invocation boundary (R10, D3): pace searches + bound the run's
+    # outbound request budget. jobspy's internal per-board transport is unpoliced.
+    # A per-crawl limiter is correct here (the crawl loop is sequential); the
+    # process-shared limiter guards the parallel ThreadPoolExecutor fan-out paths.
+    politeness_ua = PolitenessGateway().user_agent
+    limiter = HostRateLimiter()
+    request_budget = RunBudgetCounter(BROAD_BOARD_LEAD_POLICY.max_requests_per_run)
```
ebarti, line 1141: PR Reviewer - pr-reviewer

Severity: Medium
Gate: Pass
Evidence: jobspy.py:1141 creates `RunBudgetCounter(BROAD_BOARD_LEAD_POLICY.max_requests_per_run)` (default 500), and line 1183 consumes one unit per SEARCH (query x location). Each search runs `_run_one_search`, which issues up to TWO `scrape_jobs` calls (non-Glassdoor batch at line 908 + Glassdoor at line 931), and each `scrape_jobs` internally fans out across boards x pages x description fetches. The `RunBudget` port contract (domain/ports/politeness.py:130-132) defines `limit` as "Maximum outbound requests allowed for this run", and every sibling R10 surface consumes one unit per actual fetch via `session.guard(url)` (workday.py:178, compensation/sqlite_market_repository.py:106, discovery/production_wiring.py:1859, network/http_client.py:121). jobspy is the only surface where 1 budget unit != 1 outbound request.
Reproduction: Reasoned from code (sibling consumption confirmed via grep + read).

The per-run budget is consumed once per search, so `max_requests_per_run=500` permits up to 500 searches -- each up to 2 jobspy invocations and many internal HTTP requests -- not 500 outbound requests. The code comments (lines 65-71, 1135-1138) correctly disclose that jobspy's internal transport is unpoliced (no "jobspy is polite" overstatement), but they still call this an "outbound request budget", and it reuses the same `max_requests_per_run` field that means "requests" everywhere else. An operator reading/tuning this knob will mis-estimate real outbound volume by 1-2 orders of magnitude. Also note that for typical configs (e.g. 10 queries x 5 locations = 50 searches) the 500 default never binds, so the real throttle is the 1s inter-search pacing, not this budget.

Required fix: Make the unit explicit where the maintainer sees it -- either (a) relabel the comment at 1135-1138 to state this is a per-SEARCH invocation cap (counts searches, not requests), or (b) consume a count that better reflects invocations (e.g. charge per `scrape_jobs` call so Glassdoor-enabled searches cost 2). Do not police jobspy internals (correctly out of scope per D3); this is purely honest labeling of the enforced ceiling.

workers/automation/src/jobhunter/discovery/jobspy.py
```diff
@@ -1106,6 +1132,22 @@ def _full_crawl(
     # Ensure DB schema is ready
     init_db()
 
+    # Politeness invocation boundary (R10, D3): pace searches + bound the run's
+    # outbound request budget. jobspy's internal per-board transport is unpoliced.
+    # A per-crawl limiter is correct here (the crawl loop is sequential); the
+    # process-shared limiter guards the parallel ThreadPoolExecutor fan-out paths.
+    politeness_ua = PolitenessGateway().user_agent
+    limiter = HostRateLimiter()
```
ebarti, line 1140: PR Reviewer - pr-reviewer

Severity: Low
Gate: Pass
Evidence: jobspy.py:1140 constructs a fresh per-crawl `HostRateLimiter()` for `_full_crawl`, while `search_jobs` (line 1046) and every other R10 surface use `get_shared_rate_limiter()`. rate_limiter.py:1-15 and `RateLimiterPort` (domain/ports/politeness.py:164-169) state the limiter MUST be process-shared because "a per-thread or per-call limiter would re-introduce the parallel-mode bypass this train exists to kill." Both jobspy entry points key on the same host string "jobspy" (`_JOBSPY_HOST_KEY`, line 72).
Reproduction: Reasoned from code. Not reachable today: jobspy family = one sequential activity (runner.py:1399-1443) and `search_jobs` has zero callers repo-wide.

The private limiter paces searches correctly within one crawl, but it does not coordinate with the shared "jobspy" slot used by `search_jobs`, nor with any concurrent `_full_crawl` in the same worker process (e.g. two overlapping discovery runs). In those cases the intended `max_concurrency=1` and min-interval for the "jobspy" host are not jointly enforced. The justification comment ("the crawl loop is sequential") only addresses intra-crawl concurrency, not cross-path. Switching here is free: `_run_one_search` does not re-acquire the limiter, so there is no re-entrancy/deadlock risk.

Required fix: Use `get_shared_rate_limiter()` in `_full_crawl` (as `search_jobs` already does), or document explicitly why the jobspy host is intentionally exempt from the process-shared guarantee.

workers/automation/tests/test_jobspy_gateway_boundary.py
```diff
@@ -0,0 +1,61 @@
...
+def test_jobspy_budget_exhaustion_stops_crawl_and_records(monkeypatch) -> None:
+    conn = sqlite3.connect(":memory:")
+    conn.row_factory = sqlite3.Row
+    conn.execute("CREATE TABLE jobs (url TEXT)")
+    ensure_operational_metric_tables(conn)
+
+    tiny_policy = SourcePolicy(
+        policy_id="broad-board-tiny-budget",
+        allowed_methods=(SourcePolicyMethod.RENDERED_LISTING,),
+        min_request_interval_seconds=0.0,
```
ebarti, line 26: PR Reviewer - pr-reviewer

Severity: Low
Gate: Pass
Evidence: test_jobspy_gateway_boundary.py:26 sets `min_request_interval_seconds=0.0`, and the test asserts only budget-stop + recording. No test asserts that inter-search pacing is applied (that `limiter.slot` is entered per search with the policy's `min_request_interval_seconds` / `max_concurrent_requests_per_host`). Inter-search pacing is a headline deliverable of this PR and -- per the separate budget-unit finding -- the real throttle for typical configs where the 500 budget never binds.
Reproduction: Reproduced -- `uv --project workers/automation run --extra dev pytest workers/automation/tests/test_jobspy_gateway_boundary.py -q -p no:randomly` -> 1 passed; that pass does not exercise pacing.

The budget/exhaustion path is well covered, but the pacing half is only executed (with interval 0), never asserted. A regression that dropped the `with limiter.slot(...)` wrapper or passed the wrong interval would pass every test.

Required fix: Add a focused test that injects a spy/fake limiter (or a fake clock) and asserts `slot()` is invoked once per search with `BROAD_BOARD_LEAD_POLICY.min_request_interval_seconds` and `max_concurrent_requests_per_host`. `HostRateLimiter` accepts `clock`/`sleep` callables (rate_limiter.py:44-53) for deterministic timing assertions.

ebarti (commented):
PR Reviewer - pr-reviewer

**Gate: PASS** (Blocker=0, High=0). Reviewed PR #310 (`feat/r10-p2d-jobspy` @ d87e9cc) against base `feat/r10-p2c-workday`.

Verification (isolated worktree, no live network -- jobspy stubbed):
- `uv --project workers/automation run --extra dev pytest -q -p no:randomly` -> **1844 passed** in 57.15s

Confirmed:
- **Budget stops FURTHER invocations, not mid-library**: the `try_consume` check + `break` (jobspy.py:1183-1195) run before `_run_one_search`, so no partial jobspy call. Off-by-one correct (exactly `max_requests_per_run` searches run; the next records `budget_exhausted`).
- **Recording is a first-class non-error**: `is_operational_failure=0`, `is_scrape_failure=0`, `outcome="blocked"`, `failure_category="budget_exhausted"`, `source_id="jobspy"`; `budget_exhausted` is not terminal so it is never reclassified as a scrape failure. Exhaustion does not raise -> the run completes gracefully (DiscoveryRunCompleted, not Failed).
- **Persistence**: `record_operational_attempt_metric` does not commit, but on the integrated path the pending INSERT is flushed by `_record_discovery_run_event(..."DiscoveryRunCompleted")` -> `conn.commit()` (runner.py:1060) on the same thread-local connection. See residual note.
- **LinkedIn / D3 posture**: no authenticated-session code, no UA spoofing, no stealth added. The honest UA (`PolitenessGateway().user_agent`) is only recorded in the outcome, never stamped onto jobspy (correct -- D3 says we cannot). Matches the recorded decision.
- **Behavior preservation + `resolve_jobspy_boards` fallback**: search params, board resolution, filtering, storage unchanged. Budget/pacing live unconditionally in `_full_crawl`, which `run_discovery` always reaches; the DEFAULT_JOBSPY_BOARDS fallback only changes `sites`, so the fallback path is still budgeted/paced.

Findings (3 inline): 1 Medium (budget unit is per-search, not per-request -- boundary-honesty labeling of `max_requests_per_run`), 2 Low (per-crawl vs shared limiter; pacing not asserted in tests). None gate-blocking.

Residual (not inline): the `budget_exhausted` row's durability relies on a later commit on the same thread-local connection. It works in the family path (verified), but `record_politeness_outcome` is a shared no-commit helper and the test asserts in-memory presence on the same connection without a commit, so it does not prove durable persistence. A defensive `politeness_conn.commit()` after recording would make it self-contained.

COMMENT-only review (no approve / request-changes).

workers/automation/src/jobhunter/discovery/jobspy.py
ebarti: Folded into the R10 gate-hardening PR **#315 @ e27bf81e**. Relabelled the jobspy budget honestly as a per-**search-invocation** cap (rename `search_budget`, comment, outcome reason, and log all now say so) since jobspy owns its internal per-board transport we can't count. Also from the same review: `_full_crawl` now paces on `get_shared_rate_limiter()` (shared "jobspy" bucket), a deterministic inter-search pacing assertion was added, and `record_politeness_outcome`'s no-commit contract is now documented with a defensive commit at the budget-exhausted break site.

PR #313, opened 2026-07-06
feat(web): [R10] surface politeness outcomes in source health

ebarti:
## What & why

R10 phase **P4** (read/aggregate/display only). Surfaces the crawl-politeness outcomes the R10 gateway records — **robots-disallowed / rate-limited / budget-exhausted** — per source in the discovery UI, so a user can see *why* a source produced nothing instead of a silent zero. Stacked on `feat/r10-p3-browser`.

The write side (P1, #300) already records these as first-class **NON-error** outcomes in `operational_attempt_metrics`: `outcome="blocked"`, `attempt_kind="politeness_gate"`, `failure_category` set to the reason, `is_scrape_failure=0`. This PR is purely the read side.

## What the read model exposes

Additive `politeness` object on both `SourceHealthSummary` (dashboard) and `SourceRegistryEntrySummary` (discovery registry):

- `robotsDisallowedCount`, `rateLimitedCount`, `budgetExhaustedCount`
- `lastBlockedReason` (`robots_disallowed` \| `rate_limited` \| `budget_exhausted` \| `null`)
- `lastBlockedAt` (ISO, or `null`)

All zeros + `null` = no outcome recorded (honest empty state, nothing implied).

## Scope decision (deviation from the plan's literal wording — please read)

Plan P4 item #1 says to extend `source_quality_stats` + `project_source_quality`. I did **not** do that, on purpose:

- P1 landed the politeness outcomes **only** in `operational_attempt_metrics`, not as `job_events`. `project_source_quality` is built from `job_events`, so it structurally cannot source these — extending it there would force it to read a second table it doesn't own, or (worse) infer outcomes from empty results, which the RCA discipline forbids.
- The repo **already** surfaces operational outcomes (`operationalFailureCount` / `scrapeFailureCount` / `lastFailureCategory`) via a **read-time rollup** over `operational_attempt_metrics` in `listSourceHealth`, *not* via `source_quality_stats`. Politeness outcomes are the same kind of fact from the same table/writer, so the new rollup (`apps/api/src/source-politeness.ts`) sits in exactly that place and is merged into both `listSourceHealth` and `listSourceRegistry`. Every displayed reason has an explicit source of truth (the recorded gateway outcome).

**No Python read-model change / no dual-runtime parity fixture:** the only surface that aggregates `operational_attempt_metrics` for display is the TS read model. The Python side only reads `source_quality_stats` (scheduler demotion), which is untouched. Full Python suite run to prove nothing broke.

[...]

PR #312, opened 2026-07-06
feat(enrichment): [R10] P3 honest user-agent on all browser fetch surfaces

ebarti:
## Phase

**P3**, part 1 of browser routing — the **honest-UA** swap across every anonymous Playwright crawl path (surfaces **#4 smart-extract, #5 enrichment detail live batch, #6 detail port ref impl, #7 WTTJ**). Stacks on **P2d-jobspy** (#310). Base: `feat/r10-p2d-jobspy`.

## What

Rip-and-replace: deletes the spoofed `Mozilla/5.0 … AppleWebKit/537.36` desktop-browser UA from the browser paths; the single honest-UA policy sources every browser context's identity.

- `enrichment/detail.py` `UA`, `discovery/smartextract.py` `UA`, `infrastructure/enrichment/playwright_fetcher.py` `_USER_AGENT` all resolve to `default_honest_user_agent().header_value()` (`JobHunter/<version> (+repo)`). Covers `new_page`/`new_context` at detail, WTTJ, smart-extract, and the detail port fetcher.
- **Authenticated LinkedIn persistent context (#8)** — a real, user-logged-in Chrome session, semantically distinct from anonymous crawling — now presents **its own real browser identity** (`user_agent=None`) instead of a spoofed literal. Owner-scoped posture recorded per **D1/D3**: the product does not evade controls, but must not break a user-authorized flow.

## Scope note

Pre-navigation robots/budget gating and the `SITE_DELAYS` → shared-limiter swap (the parallel-mode pacing fix) land in **P3 part 2** (next stacked PR). This PR is the isolated, fully-verifiable honest-UA change — the single most-cited browser defect ("live crawlers spoof a desktop-browser UA").

## Enforcement proof

- `test_fetch_surface_enforcement.py`: no `Mozilla`/`AppleWebKit` literal remains in the browser surfaces, and the UA constants resolve to the honest identity.

## Validation

- `uv --project workers/automation run --extra dev pytest -q -p no:randomly` -> **1846 passed**
- `uv --project workers/automation run --extra dev ruff check .` -> **All checks passed!**

## Docs

Deferred to P5 (safety notes + the LinkedIn/UA posture + ADR).

PR #314, opened 2026-07-06
feat(enrichment): [R10] P3b gate every browser navigation through the politeness gateway

ebarti:
## What & why

**P3b of the R10 crawl-politeness train.** Closes the last browser-fetch bypass: every Playwright `page.goto` now runs behind a politeness-gateway verdict (robots.txt + per-host rate limit + per-run request budget + honest UA). Ground truth going in was that browser navigation obeyed no robots/rate/budget at all — a full bypass of the enforcement the rest of R10 built.

This PR also carries two review-driven items folded into the active train (details below): the **Retry-After clamp (High)** at the limiter sink, and the **#312 LinkedIn-UA fix merged forward** so P3b sits on the corrected base.

Stacked on **#312** (P3a, honest UA). **Do NOT merge** — Ready for review only; retarget to `main` at merge time.

### Surfaces now gated (browser)
| Surface | Where | Enforcement |
| --- | --- | --- |
| Enrichment detail crawl (live path) | `scrape_detail_page` via `scrape_site_batch` (sequential + parallel `_run_detail_scraper`) | robots + rate + budget; robots-block folds into the JobEnrichment stage lifecycle |
| WTTJ Algolia bootstrap | `resolve_wttj_urls` | robots + rate + budget; blocked ⇒ 0 URLs, no nav |
| `DetailPageFetcherPort` reference impl | `PlaywrightDetailPageFetcher.fetch` | robots + rate + budget; blocked ⇒ empty `DetailPage`, no nav |
| Smart Extract listing crawl | `collect_page_intelligence` | robots + rate + budget; blocked ⇒ empty intel, no nav |
| Authenticated LinkedIn session | `scrape_site_batch` resolver branch | **owner-scoped carve-out (D1/D3):** robots NOT enforced on the owner's logged-in account; rate + budget still apply; `user_agent=None` |

### How the robots-block folds into the lifecycle
`scrape_site_batch` **peeks** the gate (`session.check`) before entering `running`, because `Pending→Blocked` is a valid stage transition but `Running→Blocked` is not. A robots-deny:
- performs **zero** `page.goto`,
- sets the enrich stage to `blocked` (`ENRICH_ROBOTS_DISALLOWED`, `StageBlocked` event, `next_action` = manual import),
- records a first-class `robots_disallowed` outcome in `operational_attempt_metrics` (`is_scrape_failure=0`, `is_operational_failure=0`) — never a scrape failure,
- leaves the job enrichment-pending so a later run re-evaluates robots.

Budget exhaustion stops the batch and defers remaining jobs. The authoritative budget-consume + slot-hold happens in `scrape_detail_page`'s `guard`; the rare peek-allowed/guard-blocked parallel race is folded as a retryable deferral.

### Interaction with R9 streaming (per-job `on_job_enriched` handoff)
R9 (merged to main) fires a per-job `on_job_enriched` handoff the moment a job's enrichment commits (kicking off its scoring/preparation workflow). The politeness gate sits **upstream of the enrichment commit**, so a robots-blocked or budget-deferred job simply **never enriches and never hands off** — the correct interaction, no coupling needed. Because blocked/deferred rows are recorded with `is_scrape_failure=0` / `is_operational_failure=0` (same discipline as P1), the streamed runs on main do **not** misreport blocked pages as scrape errors once the trains combine. Reviewer: please confirm this `is_scrape_failure=0` invariant holds on the blocked and deferred paths.

### SITE_DELAYS removed
The fixed per-site `SITE_DELAYS` sleep is deleted in favor of the P1 process-wide host limiter, so parallel `ThreadPoolExecutor` mode can no longer crawl N hosts with no shared pacing. `_run_detail_scraper` shares one gateway + one run budget across all site batches (sequential and parallel). `ENRICHMENT_CRAWL_POLICY` (robots HONOR, 2.0s min-interval ≈ the old default, concurrency 1, 1000-nav budget) is the conservative default; per-host overrides are a P5/config concern (D4).

## Retry-After clamp at the limiter sink (High — folded from #300/#303 review)
`HostRateLimiter.note_retry_after` stored the server `Retry-After` deadline **uncapped**, and `_wait_for_turn` slept the full delta while holding the host semaphore slot. Since the limiter is a process-lifetime singleton fed real 429/503 `Retry-After` values by `GatewayHttpClient` (#303), a hostile/absurd header would freeze that host — and the pooled worker thread that next reaches it — for an attacker-chosen duration.

Fixed **at the sink so every current and future caller is covered**:
- `note_retry_after` clamps the stored deadline to `max_retry_after_seconds` (default **300s**), logs when it clamps, returns the effective seconds.
- `_wait_for_turn` caps any single nap at the ceiling (defensive: the loop re-checks between naps rather than sleeping an unbounded delta at once).
- **Over-clamp** hosts (server asked for longer than the cap) are flagged; the gateway's `guard` consults `hard_rate_limit_remaining` and **records a `RATE_LIMITED` outcome + skips** the fetch instead of parking a worker for the clamped cooldown ("recorded outcome instead of sleeping"). Within-cap `Retry-After` values are paced normally.
- Adds `hard_rate_limit_remaining` to `RateLimiterPort`.

Note on "`RATE_LIMITED` never emitted": it is **already** emitted on every 429/503 via `PolitenessSession.record_server_rate_limit` (landed in #303), so P4's rate-limited counter is non-zero today; this PR adds a second emission point (gateway pre-emption) and, crucially, bounds the freeze.

## #312 LinkedIn-UA fix merged forward
#312's review FAILed on a High: the authenticated LinkedIn context was built with `user_agent=UA` at the `scrape_site_batch` site while the factory used `user_agent=None`. Fixed on the #312 branch (commit `b4516e26`, unifying surface #8) with a both-sites AST test, then merged forward here so P3b builds on the corrected base.

## Decisions recorded here
- **D1/D3 (LinkedIn):** authenticated persistent context is owner-scoped — robots is an owner decision (not enforced on the logged-in account), rate + budget apply, `user_agent=None`. Documented in `_default_linkedin_apply_resolver_factory`, `_apply_authenticated_linkedin_apply_url`, and `_OwnerAuthenticatedRobots`.
- **D4 (defaults/overrides):** conservative `ENRICHMENT_CRAWL_POLICY` defaults; per-host override wiring is P5.
- **Honest UA (owner review):** the honest UA / contact string remains owner-tunable — no impersonation on surfaces we control. **The owner should review the effective UA string** (finalized in P5's config reference).

## Scope notes
- No frontend/contract files touched (P4 = #313 owns those, in parallel).
- Comprehensive user-facing docs (README safety, `data-and-safety`, `security`, pipeline docs, the ADR, config/env reference, `doctor` notices, G1 record) are **P5's** explicit scope; this PR ships only the P3-acceptance QA regression entry in `docs/local-reliability-qa.md`.

## Validation
- `uv --project workers/automation run --extra dev pytest -q -p no:randomly` → **1862 passed**
- New: `test_enrichment_politeness_gate.py` (7 tests incl. the mandated robots-zero-`goto` loopback fixture, budget stop, guard skip, LinkedIn carve-out, two-host parallel pacing); `test_fetch_surface_enforcement.py` browser-nav tripwire + both-sites LinkedIn-UA AST test; `test_politeness_gateway.py` + `test_gateway_http_client.py` Retry-After clamp tests (absurd header, within-cap pacing vs hard-skip, cooldown expiry, single-nap ceiling, gateway skip-without-slot-or-budget, end-to-end absurd-429 clamp+record+pre-empt).
- No TS touched, so `api:check`/`api:test` not applicable.

## Safety
No live crawling — all fixtures are loopback (127.0.0.1) or stubbed Playwright; no real board traffic, no submissions, no user profile/data used. The honest UA (`JobHunter/<version> (+<repo url>)`) never impersonates a browser on surfaces we control; owner review of the final UA/contact string is flagged for P5's config reference.

---

## #314 review fixes (701d6f7e)

Resolved the two gating findings from review 4633635820.

**High — robots-blocked enrich stranding (was Gate: Fail).** A robots-blocked enrich stage folds into `blocked` while its aggregate stays enrichment-pending, so a later run re-selects it. The stage state machine has no `Blocked -> Running` edge, so `set_stage_state('running')` raised `ValueError` -> caught as `ENRICH_INTERNAL_ERROR` -> job flipped `failed` -> the pending selector excluded it forever, contradicting the "a later run re-evaluates robots" promise. Fix: `scrape_site_batch` now calls `_unblock_enrich_stage_if_blocked` (state-machine `Blocked -> Pending` Unblock, recorded as an auditable `StageReset` with reason `robots_recheck`) before the running transition, so `Pending -> Running` is valid and the job re-enriches. Regressions: `test_robots_blocked_job_re_enriches_after_robots_allows` (block -> allow -> re-enriched end-to-end) and `test_repeatedly_robots_blocked_job_stays_blocked_never_failed` (still-blocked re-eval loop stays non-failed).

**Medium — authenticated LinkedIn recovery pass bypassed the gateway (functional fix).** The default-on apply-URL recovery pre-pass navigated via `resolver.resolve()` without the gateway, so neither the per-host rate limit nor the per-run budget applied — making the PR title's "every page.goto" and the carve-out's "rate + budget still apply" overclaims. Fix (the preferred functional option): the recovery pass now runs `resolver.resolve()` inside `session.guard(url)` on an owner-authenticated gateway session (`_OwnerAuthenticatedRobots` -> robots-off per D1/D3, shared host limiter + the run's single budget ON), and `_run_detail_scraper` hoists the one `run_budget` so the recovery pre-pass and every site batch share it. A budget/rate deferral skips the navigation without burning a retry attempt. This makes the "every navigation" / "rate + budget" claims true by construction, so the #316 carve-out docs stay accurate (no narrowing). Regressions: `test_recovery_pass_navigation_consumes_run_budget` and `test_recovery_pass_defers_when_run_budget_exhausted`.

**Low (folded, offered) — tightened the Retry-After regression** from `count >= 2` to `count == 2` so it bites if one request is ever double-recorded across the server-429 and guard pre-empt paths. The other three #314 Lows remain as review comments per the standing directive.

**Verification:** `uv --project workers/automation run --extra dev pytest -q -p no:randomly` -> 1866 passed; `ruff check .` -> clean. QA regression matrix row updated in `docs/local-reliability-qa.md`.


workers/automation/src/jobhunter/enrichment/detail.py
```diff
@@ -557,15 +657,108 @@ def _make_llm_extractor() -> LlmExtractor:
...
+def _record_enrich_robots_blocked(
+    conn: sqlite3.Connection, url: str, decision: PolitenessDecision
+) -> None:
+    """Fold a robots-disallowed navigation into the enrichment lifecycle.
+
+    Robots-blocked is a first-class, non-terminal outcome — never a scrape
+    failure. The enrich stage moves to ``blocked`` (``Pending -> Blocked`` is the
+    valid transition; entering ``running`` first would strand it, so this runs
+    before any attempt starts) and the job stays enrichment-pending so a later
+    run re-evaluates robots (or the owner imports the posting manually). The
```
ebarti, line 686: PR Reviewer - pr-reviewer

Severity: High
Gate: Fail
Evidence: state_machine.py:246-265 has no ('Blocked','Running') transition (only ('Blocked','Pending')). _record_enrich_robots_blocked writes enrich stage='blocked'; on a later run the job is re-selected (je stays pending/NULL — the block never touches the aggregate) and scrape_site_batch calls set_stage_state(..., 'enrich', 'running') at detail.py:1000 with the default validate_transition=True, which raises ValueError and is caught at detail.py:1274 -> _record_enrich_job_failure records ENRICH_INTERNAL_ERROR (and writes je.current_status='failed', which _ENRICHMENT_PENDING then excludes, so the job is stranded).
Reproduction: Reproduced with command/steps: seeded a temp DB, set enrich='blocked' (validate_transition=False) exactly as _record_enrich_robots_blocked does, then set_stage_state(..., 'running') -> 'ValueError: Invalid state transition for enrich: blocked -> running'.

A robots-blocked enrichment job that later becomes allowed does NOT re-enrich — it fails with a spurious ENRICH_INTERNAL_ERROR. This is reachable via the common transient case: a robots.txt 5xx/timeout fail-closes to UNKNOWN->blocked (5-min fail cache), and the next run (robots now reachable/allow) throws on the running transition. This directly contradicts this docstring's promise ('a later run re-evaluates robots') and the test comment at test_enrichment_politeness_gate.py:269-271; no test exercises the block->allow recovery.

Required fix: Before starting a previously-blocked job, transition Blocked->Pending first (the state machine's Unblock, e.g. reset_job_stage('enrich') or a set_stage_state(...,'pending', validate_transition=False)) so re-evaluation runs Pending->Running; add a regression test that blocks a job, flips robots to allow, and asserts it enriches (not ENRICH_INTERNAL_ERROR).

workers/automation/src/jobhunter/enrichment/detail.py
```diff
@@ -502,8 +594,16 @@ def _apply_authenticated_linkedin_apply_url(
...
         if page is not None and hasattr(resolver, "resolve_loaded_page"):
+            # Reuses the enrichment page, whose navigation was already gated by
+            # scrape_detail_page — no additional fetch happens here.
             resolution = resolver.resolve_loaded_page(page, url)  # type: ignore[attr-defined]
         elif hasattr(resolver, "resolve"):
+            # Fresh navigation in the owner's authenticated persistent Chrome
+            # session (retry pre-pass). This is an owner-scoped, user-authorized
+            # session (§Owner decisions D1/D3): robots is an owner decision and
+            # is not enforced on the owner's own account, and pacing/budget for
+            # the authenticated session are deferred to owner config (P5). It is
```
ebarti: PR Reviewer - pr-reviewer

Severity: Medium
Gate: Pass
Evidence: _reset_authenticated_linkedin_retry_candidates (default-on: reset_linkedin_candidates=True at runner.py:2205 and linkedin_apply_resolver_enabled() defaults True) calls _apply_authenticated_linkedin_apply_url(..., page=None) at detail.py:1719, which takes the elif branch (line 607) resolver.resolve(url) -> page.goto at linkedin_apply_resolver.py:206. That navigation does not pass through PolitenessSession/guard, so it applies neither the shared per-host rate limit nor the per-run budget.
Reproduction: Reasoned from code: grep of all Playwright navigation call sites; traced both call sites of _apply_authenticated_linkedin_apply_url (page=page at 1023 uses the gated resolve_loaded_page; page=None at 1719 uses the ungated resolve()).

This is a live, default-on authenticated LinkedIn navigation that escapes rate + budget, contradicting the PR title ('every Playwright page.goto now runs behind a politeness-gateway verdict') and the review mandate's invariant that rate + budget still apply on the owner's authenticated session. robots-off is correct here (D1/D3); the gap is pacing + budget only. Pre-existing path, transparently deferred to P5 by this comment.

Required fix: Either route the recovery-pass navigation through the owner-authenticated gateway (robots-off, rate + budget on) like the batch path, or narrow the PR title/claim and explicitly document the recovery-pass carve-out so the 'every navigation' claim is precise.

workers/automation/src/jobhunter/enrichment/detail.py
```diff
@@ -557,15 +657,108 @@ def _make_llm_extractor() -> LlmExtractor:
...
+    conn: sqlite3.Connection,
+    repo: SqliteEnrichmentRepository,
+    aggregate: JobEnrichment,
+    url: str,
+    started_at: str,
+    cascade_result: dict,
+) -> None:
+    """Fold a mid-flight budget block (peek allowed, guard denied) as a deferral.
+
+    Only reachable in the rare parallel race where the shared run budget drains
```
ebarti, line 776: PR Reviewer - pr-reviewer

Severity: Low
Gate: Pass
Evidence: check() (politeness.py:109-115) does not consult hard_rate_limit_remaining; only guard() does (politeness.py:130-136). So a peek can pass while the guard yields RATE_LIMITED for an over-clamp host, landing here with cascade_result['status']=='blocked'. The handler then records ENRICH_POLITENESS_DEFERRED as a failed attempt (attempt_count++) and breaks the whole site batch.

The docstring says this is 'Only reachable in the rare parallel race where the shared run budget drains between the pre-navigation peek and the guard', but it is also reachable via a shared-limiter over-clamp RATE_LIMITED cooldown (and, in principle, a mid-flight robots verdict flip). The break aborts the remaining site jobs (they stay pending and retry next run), which is fine, but the stated reachability is incomplete.

Required fix: Update the docstring to cover the RATE_LIMITED-cooldown path, and consider recording the deferred outcome by reason rather than always attributing it to the budget race.

workers/automation/tests/test_fetch_surface_enforcement.py
```diff
@@ -119,6 +119,38 @@ def test_browser_ua_constants_resolve_to_the_honest_identity() -> None:
...
+        and node.func.attr == "goto"
+    ]
+
+
+def test_browser_nav_surfaces_route_through_the_politeness_gate() -> None:
+    for rel in BROWSER_NAV_SURFACES:
+        source = (SRC_ROOT / rel).read_text(encoding="utf-8")
+        assert _goto_calls(source), f"{rel} unexpectedly has no page.goto"
+        # Every nav surface wires the gateway session and holds a guard slot.
+        assert "PolitenessSession" in source, f"{rel} does not wire a politeness session"
```
ebarti, line 150: PR Reviewer - pr-reviewer

Severity: Low
Gate: Pass
Evidence: test_browser_nav_surfaces_route_through_the_politeness_gate asserts only that each listed file CONTAINS the strings 'PolitenessSession' and '.guard(' and has some .goto — it does not verify each page.goto is structurally inside a guard scope, nor that the decision is honored. BROWSER_NAV_SURFACES also omits infrastructure/enrichment/linkedin_apply_resolver.py, whose resolve() page.goto (line 206) is ungated.

A file with one gated and one ungated goto, or one that calls .guard() but ignores the decision, would still pass this tripwire. The behavioral proof lives in test_enrichment_politeness_gate.py, so this is test rigor only, but the tripwire is weaker than its docstring implies ('a new ungated navigation surface ... can't slip through').

Required fix: Strengthen to assert each goto Call node is a descendant of a guard() with-block, or at minimum include linkedin_apply_resolver.py and assert its navigation posture explicitly.

workers/automation/tests/test_enrichment_politeness_gate.py
```diff
@@ -0,0 +1,472 @@
...
+            conn, workers=2, reset_linkedin_candidates=False
+        )
+
+        # Parallel enrichment across two hosts still processes every job.
+        assert stats["processed"] == 4
+        assert sorted(sink) == sorted(host_a + host_b)
+        # Per-host min-interval (2.0s) was applied to each host's second nav —
+        # this is the SITE_DELAYS replacement enforced by the shared limiter.
+        paced = [s for s in clock.sleeps if s >= 2.0]
+        assert len(paced) >= 2
```
ebarti, line 470: PR Reviewer - pr-reviewer

Severity: Low
Gate: Pass
Evidence: The assertion is len([s for s in clock.sleeps if s >= 2.0]) >= 2. Global (non-per-host) pacing over 4 navs would produce >=3 paced sleeps, which also satisfies >=2, so the assertion cannot distinguish per-host from global pacing. With a shared VirtualClock across 2 threads, an interleaving where thread A's sleep advances the clock before thread B reads now() can yield only 1 paced sleep.
Reproduction: Reasoned from code (helper VirtualClock.sleep advances a shared _t; two ThreadPoolExecutor workers).

The test name/docstring claim it 'paces each host', but the assertion neither proves per-host independence nor is guaranteed deterministic. It is a reasonable smoke test that pacing occurs in the parallel path.

Required fix: Assert per-host independence explicitly (same-host second nav paced by ~2.0s while cross-host navs are not serialized), or drive the two hosts deterministically to remove the interleaving dependence.

ebarti (commented):
PR Reviewer - pr-reviewer (summary)

Reviewed PR #314 against its stacked base feat/r10-p3-browser (three-dot diff), head 375618134fbeb19a169993020bf962b9c1aa9eeb. Python/docs only; no TS/contracts touched.

Gate: FAIL
Counts: Blocker=0, High=1, Medium=1, Low=3, Dismissed=0, Disagreements=0

Verification (in the review worktree): pytest -q -p no:randomly -> 1862 passed; ruff check . -> All checks passed. Reproduced the High with a standalone state-transition script.

High (gating): a robots-blocked enrich stage cannot be re-enriched. Blocked->Running is not a valid transition, so the next run's set_stage_state(...,'running') raises and the job is recorded as ENRICH_INTERNAL_ERROR (and then excluded from the pending queue). Reachable via transient robots 5xx/timeout fail-close. Contradicts the _record_enrich_robots_blocked docstring's re-evaluation promise; no test covers block->allow recovery. Details inline.

Medium: the default-on authenticated LinkedIn recovery pass (_reset_authenticated_linkedin_retry_candidates -> resolver.resolve() -> linkedin_apply_resolver.py:206) navigates without the gateway, so rate + budget are not applied on that owner-authenticated goto. robots-off is correct; pacing/budget are documented as P5-deferred, but this contradicts the PR title's 'every page.goto' and the mandate's 'rate + budget still enforced' invariant.

Coverage confirmed gated (Mandate 1): scrape_detail_page via scrape_site_batch (sequential + parallel), resolve_wttj_urls, PlaywrightDetailPageFetcher.fetch, collect_page_intelligence. Materials Playwright uses set_content (local render, not network) — correctly excluded.

#312 / b4516e26 verdict (Mandate 3): RESOLVED. The authenticated LinkedInApplyUrlResolver is constructed with user_agent=None at both sites (factory detail.py:1622 + scrape_site_batch detail.py:901), and test_authenticated_linkedin_context_never_uses_bot_ua_at_any_site enumerates every construction (requires >=2) and asserts user_agent=None — it would have caught the original divergence. This fix flips PR #312's UA-unification High.

Retry-After clamp (Mandate 4): correct. note_retry_after clamps at the sink to max_retry_after_seconds (300s), _wait_for_turn caps each nap, over-clamp hosts are flagged and the gateway guard records RATE_LIMITED + skips without holding a slot/budget; clamp tests are honest. Double-emission: one server 429 is recorded once, and each subsequent pre-empted skip is also recorded as failure_category='rate_limited' (test asserts count>=2). No single attempt double-counts; is_scrape_failure stays 0. Defensible and documented — a minor observability imprecision only (rateLimitedCount can over-represent actual server 429s during a cooldown).

SITE_DELAYS (Mandate 5): removed; pacing preserved by the process-wide HostRateLimiter. The two-host parallel fixture proves pacing occurs but not per-host independence (Low, inline).

Shared state / thread-safety (Mandate 6): one gateway + one run budget shared across sequential and parallel batches; HostRateLimiter (registry + per-slot locks), RunBudgetCounter (lock), and RobotsCache (lock) are all thread-safe (RobotsCache has only a benign redundant-fetch race).

R9 merge-train note (Mandate 8, not gating): R9's per-job on_job_enriched handoff is absent from this base (confirmed). At merge to main, both R9 and this PR modify scrape_site_batch's per-job loop; the gate sits upstream of the enrichment commit, so a robots-blocked/budget-deferred job never commits and never hands off. The merge must preserve ordering (gate -> commit -> handoff) and keep the already-enriched short-circuit; a blocked/deferred job must never reach on_job_enriched.

Gate FAIL is driven solely by the High. Medium/Low do not block.

workers/automation/tests/test_gateway_http_client.py
```diff
@@ -170,6 +170,39 @@ def test_server_429_is_recorded_as_rate_limit_and_delays_next_request() -> None:
...
+    # Second fetch to the same host is pre-empted by the gateway: recorded as a
+    # rate-limit outcome and skipped, so it neither hits the opener nor parks the
+    # thread on the (clamped) cooldown.
+    assert client.fetch_json("http://host/jobs") is None
+    assert len(opener.requests) == 1  # the skipped second fetch never reached urllib
+    assert clock.now() <= 300.0
+    count = conn.execute(
+        "SELECT COUNT(*) AS c FROM operational_attempt_metrics WHERE failure_category='rate_limited'"
+    ).fetchone()
+    assert count["c"] >= 2  # the 429 and the pre-empted skip both recorded
```
ebarti: PR Reviewer - review-r10-314

Severity: Low
Gate: Pass
Evidence: The two RATE_LIMITED recorders are disjoint by construction. Server-429 path: `record_server_rate_limit` (http_client.py:142). Pre-empt path: `guard` yields `_rate_limited` (politeness.py:130-135), recorded by `PolitenessSession.guard._record` (politeness.py:299). They are mutually exclusive within one `_request`: when `hard_rate_limit_remaining(host) > 0`, `guard` returns before consuming a slot or issuing the network call (politeness.py:130-136) and `_request` returns None at http_client.py:122-123, so the 429/503 branch (http_client.py:135-143) is unreachable; when it is 0, the network call may 429 and only `record_server_rate_limit` fires (the guard yielded ALLOWED, whose `_record` is a no-op at politeness.py:323). No caller retries a None result during the cooldown (workday_search/workday_detail/greenhouse/lever/compensation return the result directly; the Workday pagination loop breaks on an empty payload at ats_adapters.py:181), so no single logical request re-enters guard+fetch. This test asserts `count["c"] >= 2` (line 203) alongside `len(opener.requests) == 1` (line 198), which shows both paths fire once for two distinct requests, but `>= 2` still passes at 3, so it does not pin the "no single request traverses both paths" invariant against a future regression.
Reproduction: Reasoned from code (full control-flow trace of guard + _request plus a caller retry-loop survey). Disjointness confirmed to hold at this head.

The implementer accounting is correct: each recorded RATE_LIMITED outcome maps to a distinct request (the original 429-ed fetch, then each later fetch skipped during the over-clamp cooldown), and no single request is counted on both paths. The only gap is that this regression asserts the count loosely, so it would not fail if a future change ever recorded one request on both the server-429 and the guard pre-empt paths.

Required fix: Tighten to an exact per-request count. After exactly two distinct logical fetches, assert `count["c"] == 2` (or assert one row for the 429-ed request and one for the pre-empted skip). This makes the test bite if a single request is ever double-recorded across the two paths.

workers/automation/src/jobhunter/enrichment/detail.py
ebarti: Fixed in 701d6f7e. Before the running transition in `scrape_site_batch`, the batch now Unblocks a previously robots-blocked stage — `_unblock_enrich_stage_if_blocked` runs the state machine's `Blocked -> Pending` (recorded as an auditable `StageReset`, reason `robots_recheck`) so the subsequent `Pending -> Running` is valid and the job re-enriches instead of raising `ValueError` -> `ENRICH_INTERNAL_ERROR`.

Regression `test_robots_blocked_job_re_enriches_after_robots_allows` blocks a job (robots disallow), flips robots to allow, and asserts it re-enriches: `processed == 1`, `error == 0`, stage `succeeded`, je `enriched`, exactly one `StageReset` event. `error == 0` is the direct proof the Blocked->Running ValueError path never fired. Companion `test_repeatedly_robots_blocked_job_stays_blocked_never_failed` drives three consecutive disallowed runs and asserts the stage stays `blocked` (never `failed`), the aggregate stays pending. Full suite 1866 passed, ruff clean.

workers/automation/src/jobhunter/enrichment/detail.py
ebarti: Fixed functionally in 701d6f7e (the preferred option). The recovery pre-pass navigation now runs through an owner-authenticated gateway session: `_reset_authenticated_linkedin_retry_candidates` builds `PolitenessSession(PolitenessGateway(robots=_OwnerAuthenticatedRobots(), rate_limiter=get_shared_rate_limiter()), budget=run_budget, ...)` and threads it into `_apply_authenticated_linkedin_apply_url`, which wraps `resolver.resolve()` in `session.guard(url)` — robots-off per D1/D3, but the shared per-host limiter + the per-run budget now bound it, exactly like the batch path. `_run_detail_scraper` hoists the single `run_budget` above the recovery pass so the whole run (recovery pre-pass + every site batch) shares one budget. A budget/rate deferral skips the navigation without burning a retry attempt.

This makes the PR title's "every page.goto" and the mandate's "rate + budget still apply" true by construction, so the #316 carve-out docs stay accurate (no narrowing needed). Regressions: `test_recovery_pass_navigation_consumes_run_budget` (one resolve -> exactly one budget unit consumed) and `test_recovery_pass_defers_when_run_budget_exhausted` (drained budget -> `resolve()` never called, no attempt burned, description preserved, `budget_exhausted` outcome recorded, never a scrape failure).

workers/automation/tests/test_gateway_http_client.py
ebarti: Folded into 701d6f7e as offered. Tightened `test_absurd_retry_after_is_clamped_and_next_fetch_is_skipped` from `count["c"] >= 2` to `count["c"] == 2` (alongside the existing `len(opener.requests) == 1`), so after exactly two distinct fetches the assertion bites if a future change ever records a single request on both the server-429 and guard pre-empt paths. The other three #314 Lows stay as review comments per the standing directive.

workers/automation/src/jobhunter/enrichment/detail.py
ebarti: Confirmed RESOLVED at 701d6f7e (verified in a detached worktree at that head).

`_unblock_enrich_stage_if_blocked` (detail.py:730) runs Blocked->Pending via `set_stage_state(..., "pending")` before the running transition (detail.py:1057), and it sits inside the robots-ALLOWED branch: the peek at :1036 short-circuits a robots-blocked run through `_record_enrich_robots_blocked` + `continue` at :1046-1048, so the Unblock fires only when robots now allows. Both transitions are in `_VALID_KIND_TRANSITIONS` (Row 12 Blocked->Pending, Row 2 Pending->Running, via the shared `is_valid_transition`), so the Blocked->Running ValueError -> ENRICH_INTERNAL_ERROR path is structurally unreachable for a previously-blocked stage.

The reset is auditable-not-silent: a `StageReset` job_event carrying `{"reason": "robots_recheck", "previousState": "blocked"}`. `StageReset` is a pre-existing contract event (domain/events/orchestration.py; TS DomainEventUnion at packages/domain-types/src/events/index.ts:387 with `stageResetHandler` at apps/web/src/contexts/pipeline/handlers.ts:56), so no new event type and the every-event-has-handler parity is not implicated.

Both regressions bite: `test_robots_blocked_job_re_enriches_after_robots_allows` asserts run-2 `error == 0` + `goto_calls == [url]` + stage `succeeded` + exactly one `StageReset`; `test_repeatedly_robots_blocked_job_stays_blocked_never_failed` asserts state stays `blocked` with `error == 0` across 3 runs. pytest 1866 passed, ruff clean. Marking my High resolved.

workers/automation/src/jobhunter/enrichment/detail.py
ebarti: Confirmed RESOLVED functionally at 701d6f7e.

The recovery pre-pass `resolve()` now runs inside `session.guard(url)` on the owner-authenticated gateway (`_OwnerAuthenticatedRobots`: robots-off per D1/D3, host limiter + run budget ON) at detail.py:611-622. `_run_detail_scraper` hoists ONE `run_budget` (detail.py:1992) shared by the recovery pre-pass and every site batch. A budget/rate deferral returns `politeness_deferred` and the loop `continue`s at :1811-1817 BEFORE `record_apply_url_recovery` (:1826), so it does not burn a retry attempt. Both live call sites are gated: the page path (detail.py:1081, page=page) takes `resolve_loaded_page`, which reuses the already-gated page (no fresh goto); the recovery path (detail.py:1795, page=None) always passes `session`.

Two new tests bite: `test_recovery_pass_navigation_consumes_run_budget` (budget.remaining() == 4 after one gated goto) and `test_recovery_pass_defers_when_run_budget_exhausted` (`resolver.calls == []`, `attempt_count == 1`, `budget_exhausted` outcome with is_scrape_failure == 0). "Every page.goto gated" now holds for every reachable path on this branch.

One residual (non-blocking): the `else: resolver.resolve(url)` fallback at detail.py:624 stays ungated, but it is unreachable by any current caller (it needs page=None AND session=None with a resolve-capable resolver; neither call site does that). Dropping the ungated fallback (or making session required) would close the latent path as belt-and-suspenders, but it is not a live bypass. Marking my Medium resolved.

workers/automation/tests/test_gateway_http_client.py
ebarti: Confirmed folded at 701d6f7e: `test_absurd_retry_after_is_clamped_and_next_fetch_is_skipped` now asserts `count["c"] == 2` with the two-distinct-requests rationale. This is the tight assertion requested; it bites if a single request is ever recorded on both the server-429 and guard pre-empt paths. Resolved.

ebarti (commented):
PR Reviewer - review-r10-314 (re-review of fix round @ 701d6f7e)

Re-reviewed the #314 fix round (375618134 -> 701d6f7e) against code in a fresh detached worktree. All gate-blocking findings are resolved; verification re-run clean.

High (robots-blocked enrich stranding) — RESOLVED. `_unblock_enrich_stage_if_blocked` runs Blocked->Pending before the Pending->Running transition, inside the robots-allowed branch (the peek at detail.py:1036 folds a robots-blocked run to `blocked` and `continue`s before the Unblock is reached). Both edges are in the shared `_VALID_KIND_TRANSITIONS`, so the Blocked->Running ValueError -> ENRICH_INTERNAL_ERROR path is structurally unreachable for a previously-blocked stage. Auditable via a pre-existing `StageReset` event (no new contract type; TS DomainEventUnion + stageResetHandler already present, parity intact). Both regressions bite.

Medium (authenticated recovery-pass ungated goto) — RESOLVED functionally. The recovery `resolve()` runs inside `session.guard(url)` on the owner-authenticated gateway (robots-off per D1/D3, rate + budget on); one shared `run_budget` spans the recovery pre-pass and all site batches; a budget/rate deferral skips the goto and `continue`s before `record_apply_url_recovery`, so no retry attempt is burned. Both live call sites are gated. Residual (non-blocking): the `else: resolver.resolve(url)` fallback at detail.py:624 is ungated but unreachable by any current caller (needs page=None AND session=None with a resolve-capable resolver).

Low (two-path RATE_LIMITED count) — RESOLVED: tightened to `count["c"] == 2`.

Remaining Lows (unchanged, non-blocking, as comments): deferral docstring reachability (detail.py), weak browser-nav tripwire (test_fetch_surface_enforcement.py:150), imprecise pacing assertion (test_enrichment_politeness_gate.py:470).

QA matrix (docs/local-reliability-qa.md) now names both invariants (recovery pre-pass gating + Unblock-not-strand) and cites test_linkedin_authenticated_enrichment_retry.py.

Verification: `uv --project workers/automation run --extra dev pytest -q -p no:randomly` -> 1866 passed; `uv --project workers/automation run --extra dev ruff check .` -> All checks passed.

Gate: PASS
Counts: Blocker=0, High=0, Medium=0, Low=3, Dismissed=0, Disagreements=0

PR #315, opened 2026-07-06
test(discovery): [R10] fold P0-P3a review follow-ups + Workday concurrency test

ebarti:
## What & why

**R10 review follow-ups** — folds the Lows/Mediums surfaced by the P0-P3a reviews plus the **required-pre-merge Workday concurrency test** into one reviewable "gate hardening" PR, ahead of P5 (config/docs/gate). Stacked on **#314** (P3b). **Do NOT merge** — Ready for review only; retarget to `main` at merge time.

## Required pre-merge (from the #308 review)

A real `ThreadPoolExecutor` concurrency test for Workday gateway routing (both existing tests were single-threaded, but the load-bearing claims are threaded):
- `test_parallel_employers_pace_per_host_and_run_concurrently` — two loopback employers on **distinct hosts**, `workers=2`: per-host pacing holds under concurrency (same-host starts spaced by the min interval) **and** different hosts run in parallel (request windows overlap, not serialized). Pins the per-host limiter keying (`urlsplit(url).netloc`).
- `test_parallel_blocked_outcomes_record_on_owning_thread_under_source_id` — two employers sharing a `source_id` but with **distinct `employer_key`s** on distinct hosts, both hard rate-limited: each worker records its rate-limited outcome via **its own thread-local connection** (no `check_same_thread` `ProgrammingError`) under the correct `source_id`. This is exactly what the client-cache-by-`employer_key` fix protects.

## Mediums

- **AST tripwire broadened** (`test_fetch_surface_enforcement.py`): the raw-transport detector now flags `http.client`, `socket`, `requests`, `httpx`, `urllib3` (not just `urllib`), with a **self-test** pinning what it must and must not flag (`urllib.parse` / `http.server` stay allowed). Renamed `..._have_no_adhoc_urllib` → `..._have_no_adhoc_transport`.
- **jobspy budget relabelled honestly** (`jobspy.py`): the per-run budget counts **search invocations**, not outbound requests — jobspy owns its internal per-board tls-client transport, which we cannot count or robots-gate. Rename (`search_budget`) + comment + outcome reason + log now say so.

## Lows

- `jobspy._full_crawl` paces on `get_shared_rate_limiter()` (was a private `HostRateLimiter()`), so every jobspy invocation path shares one `"jobspy"` pacing bucket.
- Deterministic jobspy inter-search **pacing assertion** added (virtual clock via the shared-limiter accessor).
- `record_politeness_outcome` docstring now states it does **not** commit (caller owns the transaction); added a **defensive commit** at the jobspy budget-exhausted site, which breaks out before the crawl's normal end-of-run persistence.
- `PolitenessGatewayPort` gains `user_agent` + `note_retry_after`, so `PolitenessSession` drops its `# type: ignore[attr-defined]` and `getattr` duck-typing.
- **Workday**: the per-employer client is cached by **`employer_key`** (1:1 with the fan-out task/thread), not `source_id`, so two employers that resolve to the same `source_id` can never share one thread-bound recorder connection across threads; the politeness + storage `source_id` derivations are unified through one helper (so a source's politeness outcomes join to its posting rows); the module docstring / D2 comment are reconciled ("documented-API-class", not "undocumented"), and the swapped P2b/P2c label in the enforcement test is fixed.

## Deferred to P5 (docs)
- `build_opener` fail-loud-on-malformed-proxy: confirm-and-document (or restore graceful degradation) — a doc/decision item for the P5 config reference.
- `robotparser` first-match vs RFC-9309 longest-match over-blocking of `Allow` exceptions — a one-line doc note (safe direction).

## Validation
- `uv --project workers/automation run --extra dev pytest -q -p no:randomly` → **1866 passed**
- No TS/web/contract files touched.

## Safety
No live crawling — all fixtures are loopback (127.0.0.1) or `.invalid` hosts that are never reached (hard-rate-limited skip). No real board traffic, no submissions, no user profile/data used.


workers/automation/tests/test_fetch_surface_enforcement.py
```diff
@@ -19,29 +19,70 @@
...
+    routed surface must reach the network only through the gateway client.
+    ``urllib.parse`` (URL utilities) and ``http.server`` (test loopback) are not
+    transports and stay allowed.
+    """
+    module = module or ""
+    root = module.split(".")[0]
+    if root == "urllib":
+        return module != "urllib.parse"
+    if root == "http":
+        return module == "http.client" or module.startswith("http.client.")
```
ebarti, line 44: PR Reviewer - pr-reviewer

Severity: Low
Gate: Pass
Evidence: `_is_raw_transport("http")` returns False, so `_imports_raw_http` misses `from http import client` and `from http import client, server` (verified in the worktree venv: both return `[]`), while `import http.client`, `from http.client import X`, and `import http.client as hc` are all correctly flagged. The self-test (`test_raw_transport_detector_flags_every_known_transport`) has no `from http import client` case, so it gives false confidence in http.client coverage.
Reproduction: Reproduced with `_imports_raw_http("from http import client") == []`.

The broadened tripwire has a hole for the `from http import client` submodule-import form — the exact "trivial aliasing the previous guard already missed" class this PR set out to close. A routed surface could import `http.client` this way and the guard would not flag it.

Required fix: In the `root == "http"` branch, also treat `from http import client` as raw transport (for `ast.ImportFrom` with `module == "http"`, flag when an imported alias name is `client`), and add `from http import client` to the self-test must-flag list (keep `from http import server` / `http.server` in must-not-flag).

workers/automation/src/jobhunter/discovery/jobspy.py
```diff
@@ -1133,12 +1132,17 @@ def _full_crawl(
...
+    # unit == one ``_run_one_search`` call (each of which fans out to up to two
+    # ``scrape_jobs`` calls with internal board x page requests we can't police).
+    # We pace on the shared process-wide limiter's "jobspy" bucket so every jobspy
+    # invocation path (this crawl + the single manual ``search_jobs``) shares one
+    # pacing budget.
     politeness_ua = PolitenessGateway().user_agent
-    limiter = HostRateLimiter()
-    request_budget = RunBudgetCounter(BROAD_BOARD_LEAD_POLICY.max_requests_per_run)
+    limiter = get_shared_rate_limiter()
+    search_budget = RunBudgetCounter(BROAD_BOARD_LEAD_POLICY.max_requests_per_run)
```
ebarti, line 1145: PR Reviewer - pr-reviewer

Severity: Low
Gate: Pass
Evidence: The counter (`search_budget`), the recorded outcome `reason` (jobspy.py:1194) and the log line (jobspy.py:1203) were correctly relabeled to "search-invocation", but two "per-run request budget" labels remain: the module comment at jobspy.py:68 and the test docstring at tests/test_jobspy_gateway_boundary.py:3.
Reproduction: Reasoned from code + `grep -n "request budget"` (two hits remain).

Item 3 relabels this budget honestly "everywhere the old label appeared" — it counts scrape_jobs SEARCH invocations, not HTTP requests. These two sites still call it a "per-run request budget", the misleading label the relabel was meant to remove.

Required fix: Update jobspy.py:68 and tests/test_jobspy_gateway_boundary.py:3 to "search-invocation budget" (or equivalent) for consistency with the counter name, reason, and log.

workers/automation/tests/test_workday_gateway_routing.py
```diff
@@ -112,3 +121,114 @@ def open(self, request: object, timeout: float | None = None):
...
+    # (a) Per-host pacing holds under concurrency: same-host starts spaced by the
+    #     min interval.
+    for server in (server_a, server_b):
+        assert len(server.seen_times) == 2
+        assert server.seen_times[1] - server.seen_times[0] >= interval * 0.9
+    # (b) Different hosts proceed in parallel: their request windows overlap
+    #     rather than serializing one host fully before the other (which a shared
+    #     single-bucket limiter would do).
+    assert min(server_b.seen_times) < max(server_a.seen_times)
+    assert min(server_a.seen_times) < max(server_b.seen_times)
```
ebarti, line 178: PR Reviewer - pr-reviewer

Severity: Low
Gate: Pass
Evidence: The overlap assertion (`min(b)<max(a) and min(a)<max(b)`) reliably passes on the correct per-host code — empirically the two hosts first requests fire at skew ~0.001s with ~1.0s per-host gaps, so real parallelism IS demonstrated. But the assertion only pins "not fully serialized"; it does not deterministically pin the per-host limiter keying it claims to. A single-bucket (non-per-host) regression is caught only when the thread schedule serializes one host fully after the other (schedule-dependent); an interleaved schedule can still satisfy the overlap.
Reproduction: Reproduced in worktree venv — real per-host code => first-request skew 0.001s; a simulated shared-bucket limiter fully serialized this run (caught) but that outcome is scheduling-dependent.

This does NOT block: the #308 required-pre-merge condition (per-host pacing holds WHILE hosts overlap; real parallelism, not sequential-that-passes) IS satisfied.

Required fix (optional hardening): Add a direct assertion that hosts are not paced against each other, e.g. `abs(min(server_a.seen_times) - min(server_b.seen_times)) < interval` (first requests near-simultaneous), or bound total wall time to ~one interval, so the test bites deterministically against a per-host-keying regression.

ebarti (commented):
PR Reviewer - pr-reviewer (summary)

Reviewed PR #315 (`feat/r10-p5a-review-followups`) against its stacked base `feat/r10-p3b-browser-gate` via the three-dot diff. HEAD e27bf81e. Scope is finding-closures + tests only (7 files, all Python); no scope creep.

REQUIRED-PRE-MERGE (from the #308 review) — SATISFIED:
- `test_parallel_employers_pace_per_host_and_run_concurrently`: per-host pacing holds under concurrency AND distinct hosts run in parallel. Verified empirically (first-request skew ~0.001s, per-host gap ~1.0s = the 1.0s WORKDAY_API_POLICY interval). Real parallelism, not sequential-that-passes. (See inline Low: the overlap assertion could be tightened to pin per-host keying deterministically; does not block.)
- `test_parallel_blocked_outcomes_record_on_owning_thread_under_source_id`: two employers sharing a source_id but with distinct employer_keys, both hard-rate-limited; each records its rate_limited outcome under the shared source_id via its own thread-local connection. This test genuinely FAILS on the pre-fix behavior: connections are opened with check_same_thread at its default True (database.py:61), record_operational_attempt_metric executes without swallowing, and pre-fix the cache keyed by source_id forces the second thread to reuse the first thread cross-thread connection -> sqlite3.ProgrammingError -> future.result() raises. The employer_key cache fix + `emp.setdefault("employer_key", ...)` stamp in `_search_and_fetch_one` make production paths key 1:1 with the worker thread.

Folds verified:
- AST tripwire broadened to http.client/socket/requests/httpx/urllib3 with a self-test; urllib.parse/http.server correctly stay allowed. One residual evasion gap (`from http import client`) — inline Low.
- jobspy budget relabel to search-invocation: counter/reason/log correct; two comment/docstring labels still say "request budget" — inline Low.
- `_full_crawl` now paces on `get_shared_rate_limiter()`; deterministic virtual-clock pacing assertion added; `record_politeness_outcome` documented no-commit + defensive commit at the budget break; `PolitenessGatewayPort` gained `user_agent` + `note_retry_after` and PolitenessSession dropped the type-ignore/getattr (single implementation, no protocol drift; no test fakes implement the port); Workday source_id unified via `_workday_source_id`; docstring/D2 + P2b/P2c labels reconciled.

Verification (run in the review worktree):
- `pytest -q -p no:randomly` -> 1866 passed in 58.66s (Temporal gRPC warnings are pre-existing no-server noise, not failures).
- Targeted: the 2 Workday concurrency tests + jobspy pacing + AST self-test -> 4 passed.
- `ruff check .` -> All checks passed.
- No TS/web/contract files touched, so pnpm api:check not applicable.
- Note: no Skill tool is available in this environment, so code-review/security-review/archscan skills could not be invoked; equivalent review done manually (thread-safety, cross-thread SQLite, no secrets, port/gateway boundaries respected).

Minor (summary-only, not inline): the new jobspy pacing test re-implements a `_VirtualClock` although `tests/politeness_helpers.py` already exports a shared `VirtualClock` + `no_sleep_limiter` (DRY nit).

Gate: PASS
Counts: Blocker=0, High=0, Medium=0, Low=4, Dismissed=0, Disagreements=0

PR #344, opened 2026-07-06
feat: wire auto-apply standing loop

ebarti:
## Summary

- Wire the persisted `autoApply` setting into a worker-reconciled standing continuous Apply workflow.
- Keep the existing apply safety gates live: approval-required mode parks unapproved jobs, min score and concurrency are re-read for loop batches, dry-run remains non-submitting, at-most-once and CAPTCHA fail-closed behavior stay in the launcher, and budget-exceeded stops the standing loop from being restarted while the daily ceiling is still exhausted.
- Update Settings, wizard copy, Runs visibility, user docs, architecture docs, and the claims ledger so autonomous submit is documented as explicit opt-in only.

## Validation

- `uv --project workers/automation run --extra dev pytest -q workers/automation/tests/test_auto_apply_reconciler.py workers/automation/tests/test_activity_apply.py workers/automation/tests/test_workflow_apply.py workers/automation/tests/test_apply_regressions.py` (49 passed)
- `corepack pnpm --filter @jobhunter/api test -- test/server.test.ts` (378 passed)
- `corepack pnpm --filter @jobhunter/web exec vitest run src/contexts/profile/forms/settings-form.test.tsx src/views/runs/RunsTable.test.tsx` (14 passed)
- `corepack pnpm check`
- `corepack pnpm web:build`
- `corepack pnpm --filter @jobhunter/web storybook:build`
- `corepack pnpm docs:build`
- `git diff --check`

## QA

Rendered fixture-backed Storybook QA covered the Discovery automation form on desktop and 390px mobile. The default, supervised auto-apply, and autonomous submit warning states rendered correctly with no console warnings/errors and no mobile horizontal overflow. No live worker stack or real submission path was started.


workers/automation/src/jobhunter/cli.py
```diff
@@ -2010,14 +2010,22 @@ def worker(
     async def _run() -> None:
         client = await get_temporal_client()
         await _reconcile_discovery_schedule(client, queue)
+        from jobhunter.apply.auto_apply import reconcile_auto_apply_loop
+
+        identity = current_runtime_identity()
+        startup_auto_apply = await reconcile_auto_apply_loop(
```
ebarti, line 2014: **[M] Unguarded startup reconcile can crash worker boot.** The heartbeat-loop invocation of `reconcile_auto_apply_loop` is wrapped in try/except with a retry-next-iteration warning, but this startup call is bare: a transient Temporal RPC error (any status other than NOT_FOUND re-raises from `_is_workflow_running`) or an unexpected error from the budget-halt DB probe propagates out of `_run()` and takes the whole worker down before it ever starts polling. Reliability is the headline feature; a nice-to-have reconcile step shouldn't be able to prevent worker startup. Suggest the same guard the heartbeat path has (log + continue; the 15s loop will retry it anyway). Precedent note: `_reconcile_discovery_schedule` above is also unguarded, so if you prefer consistency-with-precedent, say so and this drops to a recorded risk on both.

workers/automation/src/jobhunter/apply/auto_apply.py
```diff
@@ -0,0 +1,168 @@
...
+                return False
+            daily_budget = read_daily_budget_usd(settings_path, default=25.0)
+            if daily_budget <= 0:
+                return False
+            spend_day = datetime.now(timezone.utc).date().isoformat()
+            spend = conn.execute(
+                "SELECT estimated_usd FROM llm_spend WHERE day = ?",
+                (spend_day,),
+            ).fetchone()
+    except sqlite3.Error:
```
ebarti, line 155: **[L] Budget halt fails open on `sqlite3.Error`.** If the projections DB is locked/corrupt, `_budget_halt_active` returns False and the reconciler restarts the loop every 15s; each restart immediately fails the spend preflight and writes another failed run. No overspend is possible (the workflow-level preflight is the enforcement), but the failure mode is a restart-churn loop that spams Runs with failed rows and Temporal with dead workflows precisely when the DB is already unhappy. For a safety-adjacent predicate, fail-closed (treat DB error as halt-active) or at least log the exception — a silent `return False` hides the condition entirely.

workers/automation/src/jobhunter/apply/auto_apply.py
```diff
@@ -0,0 +1,168 @@
...
+
+async def _is_workflow_running(client: Any, workflow_id: str) -> bool:
+    try:
+        description = await client.get_workflow_handle(workflow_id).describe()
+    except RPCError as exc:
+        if exc.status == RPCStatusCode.NOT_FOUND:
+            return False
+        raise
+    status = getattr(description, "status", None)
+    return status == WorkflowExecutionStatus.RUNNING or getattr(status, "name", "") == "CONTINUED_AS_NEW"
```
ebarti, line 120: **[L] The `CONTINUED_AS_NEW` name-compare is dead code.** `get_workflow_handle(workflow_id).describe()` without a run id describes the latest run in the chain, which can never itself be in `CONTINUED_AS_NEW` status — and if it somehow were, `status == WorkflowExecutionStatus.CONTINUED_AS_NEW` would be the idiomatic check rather than `getattr(status, "name", "")`. Harmless, but it reads like it's load-bearing when it isn't; either drop it or compare the enum directly with a comment saying why.

workers/automation/src/jobhunter/apply/auto_apply.py
```diff
@@ -0,0 +1,168 @@
...
+    if settings.enabled:
+        if running:
+            return AutoApplyReconcileResult(workflow_id, "already_running", True)
+        if _budget_halt_active(
+            db_path=expected_db_path,
+            workflow_id=workflow_id,
+            settings_path=settings_path,
+        ):
+            return AutoApplyReconcileResult(workflow_id, "halted_budget_exceeded", True)
+        payload = ApplyWorkflowInput(
```
ebarti, line 84: **[Q] Standing loop inherits `headless=False`.** `ApplyWorkflowInput.headless` defaults to False and the reconciler payload doesn't set it, so a background loop the user may have forgotten about will pop visible Chrome windows on their desktop whenever it claims a job. That may be the intended behavior (headed matches the manual apply default and the anti-bot posture), but for an unattended loop it's surprising enough that it deserves to be an explicit choice: either a comment here recording that headed-is-intentional, a sentence in `docs/user/configuration.md` ("the standing loop opens a visible browser window while applying"), or reuse of a headless setting. As written, neither the code nor the user docs mention it.

ebarti (commented):
Review of the auto-apply standing loop (Lane E). Verdict: **sound design, no Blockers/Highs found** — findings below are 1 Medium and a set of Lows/questions.

What I verified end-to-end before signing off on the safety story:

- **Budget-halt error-code chain is correct**: `BudgetExceededError.code = "budget_exceeded"` → `to_application_error` sets `ApplicationError.type` → `_activity_error_code` unwraps it → `emit_workflow_outcome(error_code="budget_exceeded")` → projection row matches the reconciler's halt predicate exactly. And the spend preflight runs at the start of **every** continue-as-new iteration, so the reconciler halt is genuinely anti-flap, not the enforcement layer.
- **Settings store consistency**: wizard, TS write-model, and `criteria_provider` all read/write `~/.jobhunter/dashboard.json`, with both snake_case and camelCase accepted on the read side.
- **Race-safety**: `USE_EXISTING` conflict policy + `ALLOW_DUPLICATE` reuse policy make concurrent reconcilers and post-terminal restarts correct; 15s heartbeat cadence gives no-restart-required semantics.
- **No Runs-row fan-out**: launcher-minted apply run ids are UUIDs, so the `arp.run_id = wrp.workflow_id` join yields NULL for the loop row and the "Standing apply loop" title renders as intended.
- **Docs/ledger truthfulness**: CL-030 update + CL-039 addition keep the "no auto-submit by default" claim true (defaults unchanged, opt-in explicit); wizard default flipped to False, which fixes the old wizard's untruthful default-True prompt.

Findings are inline. Severity legend: [M] Medium, [L] Low, [Q] question — none block merge on their own, per repo rules Mediums/Lows are recorded, not required fixes.

workers/automation/src/jobhunter/infrastructure/scoring/criteria_provider.py
```diff
@@ -90,7 +110,19 @@ def read_min_fit_score(
     value = settings.get("min_fit_score")
     if value is None:
         value = settings.get("minFitScore")
-    return min(10, max(1, _int(value, default)))
+    return min(10, max(0, _int(value, default)))
```
ebarti, line 113: **[L] Floor change 1→0 has a second consumer.** This aligns `read_min_fit_score` with the contracts schema (`minFitScore: min(0)`), which is right — but the digest CLI (`cli.py` `digest` command) also calls `read_min_fit_score()` for its high-fit threshold, and for a user whose settings hold `min_fit_score: 0` its behavior silently changes from clamp-to-1 to include-everything. Probably the correct semantics (honor the setting), but it's an unflagged behavior change to an unrelated surface: worth a line in the PR description and ideally a small digest-threshold test pinning the 0 case.

workers/automation/src/jobhunter/wizard/init.py
```diff
@@ -443,15 +443,37 @@ def _setup_ai_features() -> None:
...
+        "[bold]Step 5: Apply automation (optional)[/bold]\n"
+        "By default, JobHunter does not start a standing apply loop and live "
+        "submissions require Apply Review approval. You can opt into a worker-"
+        "maintained loop for eligible prepared jobs."
     ))
 
-    if not Confirm.ask("Enable autonomous job applications?", default=True):
-        console.print("[dim]You can apply manually using the tailored resumes JobHunter generates.[/dim]")
+    if not Confirm.ask("Enable the standing auto-apply loop?", default=False):
+        _update_dashboard_settings(auto_apply=False, apply_approval_required=True)
```
ebarti, line 453: **[Q] Decline path force-resets a prior explicit choice.** Declining the standing loop writes `apply_approval_required=True` — so re-running `jobhunter init` and answering "no" to the loop silently re-arms the approval gate for a user who had previously, deliberately, turned it off. Re-arming toward safety is a defensible default, but it's an invisible settings mutation from a question that only asked about the loop. Consider only writing `auto_apply=False` on decline (leave approval untouched), or printing what was reset.

apps/web/src/contexts/profile/forms/settings-form.stories.tsx
```diff
@@ -27,3 +28,21 @@ export const ZeroFitMin: Story = {
...
 };
+
+export const DiscoveryAutomationDefault: DiscoveryAutomationStory = {
+  render: (args) => <DiscoveryAutomationSettingsForm {...args} />,
+  args: {
+    initial: sampleSettingsResponse.settings,
+  },
+};
+
+export const DiscoveryAutomationAutonomous: DiscoveryAutomationStory = {
```
ebarti, line 39: **[L] Two of the four summary states have no story.** `applyAutomationSummary` renders four distinct states, and the PR's QA section says the *supervised auto-apply* state was rendered — but only Default and Autonomous exist as stories. Per the per-state story convention for forms, add `DiscoveryAutomationSupervised` (autoApply on + approval on, the info-styled state) and the off+approval-off warning state, so the a11y bar and visual QA actually cover the copy users will most commonly see after opting in.

workers/automation/tests/test_workflow_apply.py
```diff
@@ -221,6 +223,52 @@ async def fake_run_apply(_payload):
     assert captured["continued_payload"].continuous is True
 
 
+@pytest.mark.asyncio
+async def test_continuous_apply_workflow_budget_exceeded_halts_before_apply():
```
ebarti, line 227: **[L] Halt contract is pinned at both ends but not through the middle, and the release path is untested.** This test asserts the raised error type, and the reconciler test seeds a projection row with `error_code='budget_exceeded'` — but nothing proves the workflow's failure actually lands in `workflow_run_projections` with that exact code (the link runs through `_exception_error_code` → `emit_workflow_outcome` → `record_workflow_outcome`). Since `record_workflow_outcome` is already registered in this test's worker, asserting the recorded outcome's error_code would pin the contract end-to-end. Separately, the reconciler suite covers halt-active but not halt-release: a stale failed/budget_exceeded row with today's spend below the ceiling (e.g. yesterday's exhaustion) should produce `started` — that's the branch that lets the loop ever come back, and it's currently unproven.

PR #360, opened 2026-07-08
fix(contact): block private research redirects

ebarti:
### Motivation
- Supervised contact research previously authorized only the originally supplied public URL but followed redirects server-side with `urllib`, allowing attacker-controlled redirects to loopback/RFC1918/metadata endpoints whose responses were passed to the LLM, creating an SSRF/exfiltration vector.
- The change hardens the SSRF boundary by rejecting private/metadata targets and by preventing automatic redirect-driven fetches that bypass the original allowlist.

### Description
- Normalize allowlist extraction to use parsed `hostname` (via `urlsplit(...).hostname`) and strip trailing dots so host comparison is robust; update `_host` used when building per-run allowlists.
- Add hostname and address checks in `ContactResearchSourcePolicy`: `_BLOCKED_HOSTNAMES`, `_is_blocked_ip_address` (using `ipaddress`), and `_hostname_resolves_to_blocked_address` (using `socket.getaddrinfo`), and update `_is_public_web_url` to reject localhost, metadata hostnames, IPs in private/link-local/loopback/reserved/multicast/unspecified ranges, and hostnames resolving to such addresses.
- Disable automatic `urllib` redirects for contact research by using a no-redirect handler, re-authorize each redirect/final URL with the `ContactResearchSourcePolicy` before reading response bodies, enforce a redirect limit, and record the actual `final_url` on accepted fetches in the gateway fetcher.
- Add regression tests exercising the policy and fetcher: `test_public_source_rejects_local_private_and_metadata_targets` and `test_gateway_fetcher_rejects_private_redirect_target`, and update unit test scaffolding that exercises the gateway fetcher behavior.

### Testing
- Ran style/lint checks with `python -m ruff check` against the modified sources which passed.
- Verified syntax by running `python -m compileall -q` on the changed modules which completed without errors.
- Added targeted pytest cases but full `pytest` runs could not be executed in this environment due to unavailable Python dependencies (`opentelemetry`) and an inability to fetch packages via the `uv` toolchain; therefore the new tests are present but not executed here.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a4c6f6fb04c8324b309572f539aea46)

### DIFF
--- a/workers/automation/src/jobhunter/contact/activities.py
+++ b/workers/automation/src/jobhunter/contact/activities.py
@@ -156,7 +156,7 @@ def _run_contact_research(
 
 def _host(url: str) -> str:
     try:
-        return (urlsplit(url).netloc or "").lower()
+        return (urlsplit(url).hostname or "").lower().rstrip(".")
     except ValueError:
         return ""
 
--- a/workers/automation/src/jobhunter/domain/contact/source_policy.py
+++ b/workers/automation/src/jobhunter/domain/contact/source_policy.py
@@ -27,6 +27,8 @@ from __future__ import annotations
 
 from dataclasses import dataclass, field
 from enum import Enum
+import ipaddress
+import socket
 from urllib.parse import urlsplit
 
 from jobhunter.domain.discovery.source_registry import (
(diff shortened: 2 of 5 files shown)


ebarti:
Fixed by #364

PR #361, opened 2026-07-08
fix(enrichment): isolate authenticated LinkedIn resolver

ebarti:
### Motivation
- Prevent a data-exfiltration path where a weak substring match (`linkedin.com/jobs/`) and a default-enabled authenticated resolver could cause an authenticated browser context to navigate to attacker-controlled URLs and leak page HTML into the LLM prompt. 
- Ensure normal detail scraping uses a fresh unauthenticated browser context while the authenticated resolver is used only for apply-URL resolution.

### Description
- Tighten LinkedIn detection in `._is_linkedin_job` to parse the URL and require a `linkedin.com` hostname (or subdomain) with a `/jobs` path for absolute URLs, and only accept discovery-local relative `/jobs/...` paths when `site == "linkedin"` (file: `workers/automation/src/jobhunter/enrichment/detail.py`).
- Always create a fresh unauthenticated Playwright page for `scrape_detail_page` and reserve the resolver's persistent authenticated page (`resolver_page`) only for authenticated apply-URL resolution; pass `resolver_page` into `_apply_authenticated_linkedin_apply_url` instead of reusing the normal scrape page (file: `workers/automation/src/jobhunter/enrichment/detail.py`).
- Update log messages to reflect the clarified behavior and failure paths (file: `workers/automation/src/jobhunter/enrichment/detail.py`).
- Add a regression test `test_linkedin_job_detection_requires_linkedin_host` to ensure spoofed URLs like Gmail fragments or non-LinkedIn hosts containing `linkedin.com/jobs/` do not trigger the authenticated resolver (file: `workers/automation/tests/test_linkedin_authenticated_enrichment_retry.py`).

### Testing
- Ran `python -m compileall -q workers/automation/src/jobhunter/enrichment/detail.py workers/automation/tests/test_linkedin_authenticated_enrichment_retry.py` which succeeded.  
- Attempted `PYTHONPATH=workers/automation/src python -m pytest -q workers/automation/tests/test_linkedin_authenticated_enrichment_retry.py workers/automation/tests/test_linkedin_apply_resolver.py`, which failed due to missing test dependency `opentelemetry` in the execution environment.  
- Attempted project test/lint commands via the `uv` wrapper (`uv --project workers/automation run --extra dev pytest` and `uv --project workers/automation run --extra dev ruff check`) but these were blocked by environment dependency resolution and PyPI network failures and therefore could not be completed.  
- Performed static diffs and local formatting checks (`git diff --check`, local file syntax checks) to verify the change is syntactically correct and the new test is present.  

If you want, I can add an environment-only gate or an explicit opt-in env var to disable the authenticated resolver by default as an additional safety measure, but the current changes remove the immediate cross-host substring-trigger and isolate authenticated navigation to the resolver-only page.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a4a7644c7008324bf6cfa2ab1abed75)

### DIFF
--- a/workers/automation/src/jobhunter/enrichment/detail.py
+++ b/workers/automation/src/jobhunter/enrichment/detail.py
@@ -31,7 +31,7 @@ import time
 from collections.abc import Callable
 from concurrent.futures import ThreadPoolExecutor, as_completed
 from datetime import datetime, timezone
-from urllib.parse import urljoin
+from urllib.parse import urljoin, urlparse
 
 from playwright.sync_api import sync_playwright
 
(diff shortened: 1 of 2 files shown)


ebarti:
Fixed by #362

PR #364, opened 2026-07-08
fix(security): block private contact research redirects

ebarti:
## Summary

Fixes an SSRF/exfiltration path in supervised contact research public-page fetching.

- Rejects loopback, private-network, link-local, reserved, unspecified, multicast, and metadata-style public-page targets before fetch.
- Disables urllib automatic redirects for contact research and manually re-authorizes each redirect target through the contact source policy and public-address DNS check.
- Routes every followed redirect through the same politeness session guard before opening it.
- Records the actual final URL on successful fetches and documents the contact-research egress boundary.

## Root Cause

Contact research authorized only the originally supplied public-page URL, then used urllib's default redirect-following opener. An attacker-controlled allowed page could redirect the worker to a loopback/private endpoint, and the fetched response body could be included in the LLM extraction prompt.

## Validation

- `uv --project workers/automation run --extra dev pytest -q workers/automation/tests/test_contact_research_source_policy.py workers/automation/tests/test_contact_research_workflow.py`
- `uv --project workers/automation run --extra dev ruff check workers/automation/src/jobctrl/domain/contact/source_policy.py workers/automation/src/jobctrl/infrastructure/contact/research_fetcher.py workers/automation/src/jobctrl/contact/activities.py workers/automation/tests/test_contact_research_source_policy.py workers/automation/tests/test_contact_research_workflow.py`
- `corepack pnpm check`
- `corepack pnpm docs:build`
- `corepack pnpm test`
- `uv --project workers/automation run --extra dev pytest -q` (final rerun: 2120 passed, 1 warning)
- Manual regression harness: allowed public source redirected to `127.0.0.1`; fetch returned `rejected`, private target was not fetched, and the LLM was not called.

Note: one full worker-suite rerun hit a transient unrelated SQLite lock in `test_parallel_two_host_run_paces_each_host_via_shared_limiter`; the isolated test passed immediately after, and the final full worker-suite rerun passed.


workers/automation/src/jobctrl/infrastructure/contact/research_fetcher.py
```diff
@@ -110,4 +194,45 @@ def _visible_text(body: bytes) -> str:
...
+        hostname = (parts.hostname or "").lower().rstrip(".")
+        port = parts.port
+    except ValueError:
+        return False
+    if parts.scheme not in {"http", "https"} or not hostname:
+        return False
+    literal = _ip_address(hostname)
+    if literal is not None:
+        return literal.is_global
+    addresses = resolver(hostname, port)
```
ebarti, line 209: DNS-rebinding TOCTOU. This correctly validates the *resolved* addresses (good — not just the hostname string), but the resolution here (`resolver(hostname, port)`) and the actual connection at `self._opener.open(request)` (line 96) resolve DNS independently. A hostname that answers with a public A record for this check and a private one for the urllib connection (low-TTL rebind, or a round-robin that flips between calls) reaches the private target after passing the guard — the canonical bypass of a resolve-then-fetch SSRF check, and it hits the exact threat this PR addresses. It is reachable in this design because `is_allowlisted` matches subdomains (`host.endswith('.' + domain)`), so an attacker-controlled subdomain of any opted-in domain is eligible for the redirect/target path. Consider pinning the validated address for the connection (resolve once, connect by IP while preserving the `Host` header and TLS `server_hostname`/SNI) so the address that is checked is the address that is fetched. Not a blocker given the allowlist gate, but it is the residual gap in this control.

ebarti (commented):
SSRF redirect fix looks sound for the primary threat (private/redirect targets are re-authorized and DNS-checked per hop; a REJECTED fetch never reaches the LLM). One residual note on the resolve-then-fetch pattern.

PR #366, opened 2026-07-08
fix: block unsafe enrichment URLs before LLM extraction

ebarti:
## Summary

- Add a shared public HTTP(S) destination guard for browser-backed discovery and enrichment fetches.
- Block direct unsafe URLs, redirects, final URLs, and Playwright subrequests to loopback, private, link-local, metadata-service, and non-HTTP(S) destinations before captured content can reach enrichment or LLM-assisted extraction.
- Record unsafe enrichment targets as non-retryable `DETAIL_UNSAFE_URL` failures and document the destination boundary.

## Root Cause

Browser-backed enrichment trusted the originally queued posting URL after navigation. An attacker-controlled page could redirect or trigger browser requests to local-only resources, then the fetched content could be included in downstream LLM extraction prompts.

## Validation

- `uv --project workers/automation run python <scanner PoC>`: original SSRF-to-LLM PoC no longer reproduced; provider request count was `0`, provider request contained no secret, and the scrape returned `security_outcome="unsafe_url"`.
- `uv --project workers/automation run --extra dev pytest -q workers/automation/tests/test_enrichment_url_safety.py workers/automation/tests/test_enrichment_politeness_gate.py workers/automation/tests/test_enrichment_detail_staleness.py workers/automation/tests/test_browser_ua_propagation.py workers/automation/tests/test_fetch_surface_enforcement.py`: `43 passed`.
- `uv --project workers/automation run --extra dev ruff check <touched Python files>`: passed.
- `uv --project workers/automation run --extra dev pytest -q`: `2130 passed, 1 warning`.
- `corepack pnpm docs:build`: passed, docs link check resolved `4812` references.
- `git diff --check`: passed.

### DIFF
--- a/workers/automation/src/jobctrl/infrastructure/network/url_safety.py
+++ b/workers/automation/src/jobctrl/infrastructure/network/url_safety.py
@@ -1,0 +1,143 @@
+"""Public-destination checks for browser/network fetches."""
+
+from __future__ import annotations
+
+import ipaddress
+import socket
+from dataclasses import dataclass
+from typing import Any, Callable
+from urllib.parse import urlsplit
+
+
+Resolver = Callable[..., list[tuple[Any, Any, Any, Any, tuple[Any, ...]]]]
+
+
+@dataclass(frozen=True)
+class PublicUrlDecision:
+    allowed: bool
+    reason: str | None = None
+
+
+def validate_public_http_url(url: str, *, resolver: Resolver | None = None) -> PublicUrlDecision:
+    """Allow only public HTTP(S) URLs.
+
+    Hostnames are resolved and every returned address must be globally routable.
+    If resolution fails, fail closed: the browser would otherwise make the real
+    routing decision later, outside this security control.
+    """
+
+    if not isinstance(url, str) or not url.strip():
+        return PublicUrlDecision(False, "URL must be a non-empty string")
+
+    try:
+        parsed = urlsplit(url.strip())
+        port = parsed.port
+    except ValueError as exc:
+        return PublicUrlDecision(False, f"URL is invalid: {exc}")
+
+    scheme = parsed.scheme.lower()
+    if scheme not in {"http", "https"}:
+        return PublicUrlDecision(False, "URL scheme must be http or https")
+
+    hostname = parsed.hostname
+    if not hostname:
+        return PublicUrlDecision(False, "URL host is required")
+
+    host = hostname.rstrip(".")
+    if not host:
+        return PublicUrlDecision(False, "URL host is required")
+
+    literal = _ip_literal(host)
+    if literal is not None:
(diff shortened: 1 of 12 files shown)


workers/automation/src/jobctrl/infrastructure/network/url_safety.py
```diff
@@ -0,0 +1,143 @@
...
+
+def _ip_literal(value: str) -> ipaddress.IPv4Address | ipaddress.IPv6Address | None:
+    try:
+        return ipaddress.ip_address(value)
+    except ValueError:
+        return None
+
+
+def _is_public_address(address: ipaddress.IPv4Address | ipaddress.IPv6Address) -> bool:
+    return bool(address.is_global)
```
ebarti, line 143: PR Reviewer - pr-reviewer

Severity: Low
Gate: Pass
Evidence: `_is_public_address` returns `address.is_global`. Python's stdlib reports `is_global=True` for the NAT64 well-known prefix `64:ff9b::/96` even when it embeds a private/loopback/metadata IPv4. Verified on this repo's interpreter (3.14) and on 3.11.4/3.11.9/3.12.2/3.13: `validate_public_http_url('http://[64:ff9b::a9fe:a9fe]/')` (embeds 169.254.169.254) and `http://[64:ff9b::7f00:1]/` (embeds 127.0.0.1) both return allowed=True. 6to4 (`2002::/16`) and IPv4-mapped (`::ffff:*`) are correctly blocked; NAT64 is the one IPv4-embedding form `is_global` does not unwrap.
Reproduction: Reproduced with a script calling `validate_public_http_url` against `[64:ff9b::7f00:1]` and `[64:ff9b::a9fe:a9fe]` in the worker venv — both returned allowed=True.

The guard's stated blocklist includes link-local and metadata-service destinations, but a NAT64 address that encodes 169.254.169.254 / 127.0.0.1 passes it. Exploitability is environment-dependent (needs a reachable NAT64 translator on the worker's network, so it is low for the local-first desktop case), but it is a real gap versus this PR's goal of blocking internal destinations.

Required fix: In `_is_public_address`, also reject IPv6 addresses that embed a non-public IPv4 that `is_global` does not unwrap - e.g. reject the NAT64 well-known prefix (`ipaddress.ip_network('64:ff9b::/96')`) and/or re-validate the `.sixtofour` / `.ipv4_mapped` embedded address - then add a `[64:ff9b::a9fe:a9fe]` vector to test_enrichment_url_safety.py.

ebarti (commented):
Reviewed the public-destination guard end to end (validator, Playwright route guard, redirect/final-URL checks, and all enrichment/discovery fetch entry points). Core control is sound and verified empirically across Python 3.11.4-3.14; one Low, environment-dependent gap noted inline. No Blocker/High findings.

PR #368, opened 2026-07-08
fix: block private apply target URLs

ebarti:
## Summary

- block apply runs before prompt construction when the effective application target is not a public HTTP(S) URL
- record a terminal non-retryable `unsafe_url` failure with an `UnsafeApplyTargetBlocked` event
- guard the prompt debug helper and classify `unsafe_url` launcher failures as permanent

## Root Cause

The apply use case checked ordinary readiness but did not enforce the existing public-destination URL policy before rendering the agent prompt. A private-network `application_url` could therefore be handed to the browser-driving apply agent.

## Validation

- `uv --project workers/automation run --extra dev pytest -q workers/automation/tests/test_apply_use_cases.py::test_execute_blocks_private_network_application_url_before_agent` failed before the fix and passed after the fix
- `uv --project workers/automation run --extra dev pytest -q workers/automation/tests/test_apply_use_cases.py` -> 12 passed
- `uv --project workers/automation run --extra dev pytest -q workers/automation/tests/test_apply_regressions.py::test_unsafe_url_failure_is_permanent` -> 1 passed
- `uv --project workers/automation run --extra dev pytest -q workers/automation/tests/test_enrichment_url_safety.py` -> 15 passed
- `uv --project workers/automation run --extra dev pytest -q workers/automation/tests/test_apply_use_cases.py workers/automation/tests/test_apply_regressions.py workers/automation/tests/test_apply_prompt_builder.py workers/automation/tests/test_activity_apply.py workers/automation/tests/test_workflow_apply.py workers/automation/tests/test_apply_tools_mcp_server.py` -> 78 passed
- `uv --project workers/automation run --extra dev ruff check .` -> passed
- `uv --project workers/automation run --extra dev pytest -q` -> 2147 passed, 1 dependency deprecation warning
- `git diff --check` -> passed


workers/automation/src/jobctrl/apply/launcher.py
```diff
@@ -1209,6 +1209,16 @@ def gen_prompt(
     if not job:
         return None
 
+    apply_url = str(job.get("application_url") or job.get("url") or "").strip()
```
ebarti, line 1212: PR Reviewer - pr-reviewer

Severity: Low
Gate: Pass
Evidence: This debug-helper guard resolves the target with `str(job.get("application_url") or job.get("url") or "").strip()`, which differs from the enforced use-case helper `_apply_target_url` in domain/apply/use_cases.py (`str(application_url or "").strip() or str(url or "").strip()`). A grep of workers/automation/tests/ shows no test references `gen_prompt`, so this copy of the guard is unverified (the two new tests cover the use-case guard, not this one).
Reproduction: Reasoned from code. When `application_url` is whitespace-only (e.g. "  ") and `url` is a valid public URL, `"  " or url` short-circuits to "  ", which strips to "" and raises here, whereas `_apply_target_url` falls through to `url` and allows. This over-blocks (fail-safe direction), but the debug helper and the real apply path can reach different decisions for the same job.

The gen_prompt guard duplicates the use-case guard with slightly different target-resolution precedence and has no test, so it can silently regress or diverge from the behavior actually enforced on the apply path.

Required fix: Reuse `_apply_target_url(job)` (or match its precedence exactly) here, and add a focused test asserting gen_prompt raises on a private/non-public target and proceeds on a public one.

ebarti (commented):
One Low finding on the gen_prompt debug-helper guard (target-resolution divergence from the enforced use-case guard + no test). The primary use-case URL guard is correct and well covered by the two new tests. Gate: PASS.

PR #369, opened 2026-07-08
fix(security): restrict analysis command filesystem reads

ebarti:
## Summary

- Keep Codex analysis on a JobCtrl-owned `codex_home/` so app-server session state does not clutter the user's normal Codex app history.
- Run prompt-driven analysis commands from `codex_home/workspace/` and make that directory the only command-readable workspace root through a Codex permissions profile.
- Copy source Codex auth from the user's regular `CODEX_HOME` into the JobCtrl-owned Codex home for app-server auth, while keeping `auth.json` outside the command-readable workspace.
- Update docs and regression tests for the `codex_home/workspace` layout.

## Root Cause

The vulnerable path was prompt-driven read-only command execution. The analysis adapter needs a separate Codex home to avoid polluting the user's normal Codex app state, but copied auth must not be readable to command tooling that can be influenced by untrusted job descriptions.

This change keeps the separate JobCtrl-owned Codex home and adds the runtime-enforced boundary: only `codex_home/workspace/` is readable to sandboxed analysis commands, while `codex_home/auth.json` remains outside that subtree.

## Validation

- `uv --project workers/automation run --extra dev pytest -q workers/automation/tests/test_codex_home_isolation.py workers/automation/tests/test_employer_analysis_ensemble.py::TestCodexAdapter` -> `10 passed`
- `uv --project workers/automation run --extra dev pytest -q` -> `2129 passed, 1 warning`
- `uv --project workers/automation run --extra dev ruff check .` -> `All checks passed!`
- `corepack pnpm docs:build` -> passed; docs site link check passed across 4,813 references / 236 emitted files
- `python3 scripts/release_check.py` -> `Release check passed.`
- `git diff --cached --check` -> passed before commit

## Notes

The live pinned app-server regression uses fake auth and asserts the configured permissions profile returns `DENIED` for `$CODEX_HOME/auth.json`, because only `$CODEX_HOME/workspace/` is granted as the command workspace root.

### DIFF
--- a/workers/automation/src/jobctrl/infrastructure/analysis/codex_analysis_adapter.py
+++ b/workers/automation/src/jobctrl/infrastructure/analysis/codex_analysis_adapter.py
@@ -6,13 +6,12 @@ Node sidecar, no CLI-subprocess wrapper), forcing JSON-schema-constrained
 output via ``output_schema`` on ``thread.run`` and parsing the structured
 result off ``TurnResult.final_response``.
 
-CODEX_HOME isolation: the live factory redirects the Codex SDK's ``CODEX_HOME``
-to an isolated dir under the JobCtrl runtime root (``~/.jobctrl/codex_home``)
-seeded with a copy of the user's effective ``CODEX_HOME/auth.json`` (default
-``~/.codex/auth.json``), so Codex session rollouts never land in — and pollute
-— the user's real Codex chat history. The SDK merges ``CodexConfig.env`` over
-the parent environment, so only ``CODEX_HOME`` is overridden; PATH/HOME/etc.
-are preserved.
+Codex isolation + permissions: the live factory redirects the SDK to a
+JobCtrl-owned ``CODEX_HOME`` so Codex app-server state does not pollute the
+user's normal Codex app history. Prompt-driven commands run from the dedicated
+``CODEX_HOME/workspace`` directory under a Codex permissions profile that denies
+root filesystem reads and grants only that workspace plus Codex's minimal
+runtime paths.
 
 Test-mockability (no live auth in tests — D-04): the ``AsyncCodex`` class is
 resolved through an injectable factory that defaults to a lazy import, so tests
@@ -25,8 +24,10 @@ No timeout (D-19): ``thread.run`` is awaited to completion; effort is pinned to
 
 from __future__ import annotations
 
+import os
 import shutil
 from collections.abc import Callable
+from dataclasses import dataclass
 from pathlib import Path
 from typing import Any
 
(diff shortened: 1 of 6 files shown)


ebarti (commented):
Two findings on the copied-auth scrub: a concurrency hole that defeats the new invariant under `--workers > 1`, and a validation gap around whether a live analysis turn still authenticates after the scrub.

workers/automation/src/jobctrl/infrastructure/analysis/codex_analysis_adapter.py
```diff
@@ -110,15 +172,23 @@ def _load_async_codex_factory() -> AsyncCodexFactory:
...
-        config_kwargs: dict[str, Any] = {
-            "env": _isolated_codex_env(codex_home),
-            "config_overrides": _CODEX_CONFIG_OVERRIDES,
-        }
-        # Prefer the SDK-pinned bundled runtime. A system ``codex`` on PATH may
-        # speak a different app-server protocol; JOBCTRL_CODEX_BIN is the
-        # explicit escape hatch for setup-managed platform fallbacks.
-        config_kwargs["codex_bin"] = str(resolve_codex_binary())
-        return AsyncCodex(config=CodexConfig(**config_kwargs))
+        workdir = _prepare_isolated_codex_workdir()
```
ebarti: High — The auth-scrub invariant does not hold when analysis runs concurrently. `_prepare_isolated_codex_workdir()` is per-run (mkdtemp), but `_prepare_isolated_codex_home()` on the line above returns the single shared `<APP_DIR>/codex_home`. `score`/`tailor` support parallel legs via `--workers N` (cli.py:1133/1144 -> `score_activity`'s `ThreadPoolExecutor`, activities.py:227), so two Codex legs can run `_make()` against the same `codex_home` concurrently. Two races follow:

1. Security re-exposure: Leg A deletes `auth.json` after its app-server init; Leg B's `_prepare_isolated_codex_home()` then re-copies `auth.json` into the shared home (target now missing) while Leg A's untrusted job-description turn is still running — A's read-only command path can again read `$CODEX_HOME/auth.json`, defeating the exact invariant this PR establishes and that `test_pinned_app_server_cannot_read_copied_auth_after_startup` asserts.
2. Reliability: conversely, a concurrent leg's scrub can delete `auth.json` before another leg's app-server has loaded it, producing spurious Codex-leg auth failures that the partial-failure-safe ensemble silently records as a degraded analysis.

Default `workers=1` is safe, but the invariant silently breaks under a documented setting. Make the auth seed per-run too (e.g. a per-run `CODEX_HOME` co-located with the workdir, cleaned up in `__aexit__`), or serialize copy+init+scrub under a lock.

workers/automation/src/jobctrl/infrastructure/analysis/codex_analysis_adapter.py
```diff
@@ -95,9 +107,59 @@ def _prepare_isolated_codex_home() -> Path:
...
+        self._workdir = workdir
+
+    async def __aenter__(self) -> Any:
+        try:
+            entered = await self._codex.__aenter__()
+        except Exception:
+            _remove_copied_codex_auth(self._codex_home)
+            _cleanup_codex_workdir(self._workdir)
+            raise
+        _remove_copied_codex_auth(self._codex_home)
```
ebarti: Medium — Nothing here proves a live analysis turn still authenticates after `auth.json` is removed at this point (post-`initialize`, pre-`thread.run`). The new pinned-app-server test only issues `command/exec`, a local read-only op that needs no OpenAI credentials — it never runs a model completion. `AsyncCodex.__aenter__` does spawn+`initialize` the app-server, so this is safe iff the app-server caches credentials for the whole turn. That holds for a static API key, but Codex ChatGPT/OAuth auth refreshes the access token by re-reading `auth.json`, and analysis turns are `high`-effort with no timeout (D-18/D-19), so a refresh mid-turn would hit the deleted file and fail the leg. D-04 forbids live auth in the suite, so please confirm with a manual live analysis run (and record it under Validation) that a real turn completes after the scrub; if the app-server does re-read auth at turn time, defer the unlink until after the turn instead.

workers/automation/src/jobctrl/infrastructure/analysis/codex_analysis_adapter.py
ebarti: Fixed in 262b620d: the copied auth seed now lives in a per-run CODEX_HOME under the run workspace, and the whole workspace is cleaned up on exit. I also added an overlap regression that starts a second context while the first remains open and proves the second auth copy does not re-expose the first context.

workers/automation/src/jobctrl/infrastructure/analysis/codex_analysis_adapter.py
ebarti: Confirmed and recorded under PR Validation after the pushed fix: a synthetic live Codex adapter smoke completed a real gpt-5.5 thread.run after the post-initialize auth scrub (`LIVE_CODEX_SMOKE_OK model=gpt-5.5 requirements=3 keywords=3`). The smoke used synthetic JD text and printed only pass/fail metadata.

ebarti:
PR Fixer - Codex

Addressed the two unresolved review threads on the copied Codex auth scrub.

- PRRT_kwDOSQ-EBM6PUSWa / discussion_r3545814163: fixed the High concurrency issue by moving the seeded CODEX_HOME into each per-run workspace and cleaning the workspace on exit. Added an overlap regression proving a second context cannot re-expose the first context auth seed.
- PRRT_kwDOSQ-EBM6PUSWf / discussion_r3545814170: completed and recorded the requested synthetic live Codex adapter smoke under PR Validation (`LIVE_CODEX_SMOKE_OK model=gpt-5.5 requirements=3 keywords=3`).

Pushed commit: 262b620de25c10a75efd6aba0efa8f1d07648ba6

Validation passed: ruff on touched Python files, focused pytest set (36 passed), docs:build, strict release check, git diff --check, and the live Codex smoke.

PR #372, opened 2026-07-08
fix: derive profile evidence and repair link toolbar

ebarti:
## Summary

- Remove editable achievement-evidence controls from the structured profile editor so extracted evidence is no longer presented as user input.
- Persist bullet-derived achievement evidence in both the TypeScript profile store and Python SQLite profile repository so tailoring can reuse materialized evidence across future runs.
- Replace the always-visible resume toolbar URL field with an editor-style Insert link command that opens a focused URL popover from the toolbar button or Cmd/Ctrl+K.

## Why

Achievement evidence is derived from resume bullets and should be stored as reusable profile audit data, not manually entered by the user. The resume editor also exposed hyperlinking as a permanent toolbar text field, which did not match normal rich-text editor behavior.

## Validation

- `corepack pnpm check`
- `corepack pnpm test`
- `git diff --check`
- Browser check on `/profile`: no persistent Link URL field; Cmd+K opens and focuses the link popover; Escape closes it; no console warnings/errors observed.

Note: the browser annotation overlay intercepted pointer hit-testing on the toolbar icon during live inspection, so the icon-click path is covered by the web regression test instead.

### DIFF
--- a/apps/api/src/profile-store.ts
+++ b/apps/api/src/profile-store.ts
@@ -49,6 +49,27 @@ const DEFAULT_STYLE = {
 };
 
 const DEFAULT_RESUME_TEMPLATE = "{{ personal_data }}\n\n{{ resume_body }}\n";
+const LEGACY_BULLET_METRIC_PATTERN =
+  /(?:\$\s?\d+(?:[,.]\d+)*(?:\.\d+)?\s?(?:k|m|b|million|billion)?|\d+(?:\.\d+)?%|\d+(?:\.\d+)?x|\d+(?:\.\d+)?\s?(?:ms|milliseconds?|seconds?|minutes?|hours?|days?|weeks?|months?|years?|qps|req\/s))/gi;
+const LEGACY_BULLET_SENIORITY_TERMS = [
+  "own",
+  "owned",
+  "ownership",
+  "scope",
+  "influence",
+  "influenced",
+  "cross-team",
+  "stakeholder",
+  "stakeholders",
+  "led",
+  "lead",
+  "mentor",
+  "mentored",
+  "architect",
+  "architected",
+  "strategy",
+  "technical leadership",
+] as const;
 
 const SUPPORTED_PROFILE_TOP_LEVEL_KEYS = new Set([
   "personal",
@@ -389,6 +410,8 @@ function ensureCandidateProfileColumns(db: SqliteDatabase): void {
       db.exec(`ALTER TABLE candidate_profiles ADD COLUMN ${column} ${definition}`);
     }
   }
+
+  backfillAchievementEvidenceFromBullets(db);
 }
 
 export function readProfileConfig(db: SqliteDatabase): ProfileConfigResponse {
@@ -534,7 +557,7 @@ function insertChildren(db: SqliteDatabase, profile: ProfileShape): void {
     asTextArray(entry.bullets).forEach((bullet, bulletIndex) => {
       insertBullet.run(TENANT_ID, PROFILE_ID, entryId, bulletIndex, bullet);
     });
-    asRecordArray(entry.achievement_evidence).forEach((evidence, evidenceIndex) => {
+    achievementEvidenceForEntry(entry, entryId).forEach((evidence, evidenceIndex) => {
       insertEvidence.run(
         TENANT_ID,
         PROFILE_ID,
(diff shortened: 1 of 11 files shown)


apps/api/src/profile-store.ts
```diff
@@ -625,6 +648,106 @@ function insertChildren(db: SqliteDatabase, profile: ProfileShape): void {
   });
 }
 
+function achievementEvidenceForEntry(entry: Record<string, unknown>, entryId: string): Array<Record<string, unknown>> {
+  const explicitEvidence = asRecordArray(entry.achievement_evidence);
+  if (explicitEvidence.length > 0 || !entryId) {
```
ebarti: **[High]** This early-return makes materialized evidence sticky: whenever any explicit `achievement_evidence` exists it is persisted verbatim (the Python `get_achievement_evidence` precedence does the same, sqlite_repository.py:928). Because this PR now *persists* bullet-derived evidence and the structured editor re-sends it unchanged (see StructuredProfileEditor.test.tsx asserting evidence is preserved when editing Title), editing a resume bullet after the first save leaves the stored `source_text`/`action`/`outcome` frozen at the old bullet text. There is no refresh path: the backfill only fires when evidence is entirely absent, and the editor controls that could have corrected it were removed in this PR. That stale evidence then feeds the fabrication detector, scoring, quality, and interview surfaces (all `get_achievement_evidence` consumers) plus the audit read model, a behavioral regression from the pre-PR always-fresh derivation and a break of the rule that displayed evidence must trace to current canonical bullets. Please decide how derived evidence should re-derive when its source bullet changes (e.g. keep it non-explicit, or reconcile by bullet index) and add a regression test for the edit-bullet-after-materialize sequence.

apps/web/src/contexts/materials/components/ResumeAuditPins.tsx
```diff
@@ -2041,13 +2050,66 @@ function ResumeEditorToolbarControls({
...
   const [fontSizeScale, setFontSizeScale] = useState(String(RESUME_EDITOR_DEFAULT_SIZE_SCALE));
   const [linkUrl, setLinkUrl] = useState("");
   const [linkError, setLinkError] = useState<string | null>(null);
+  const [linkPopoverOpen, setLinkPopoverOpen] = useState(false);
+  const openLinkPopover = useCallback(() => {
+    if (disabled) return;
+    setLinkError(null);
+    setLinkPopoverOpen(true);
+  }, [disabled]);
+  const closeLinkPopover = useCallback(() => {
```
ebarti, line 2070: **[Low]** `closeLinkPopover` (invoked on Escape, outside pointerdown, and Cancel) only updates state and never returns focus to the "Insert link" trigger button. The popover is a `role="dialog"` opened via Cmd/Ctrl+K whose focused input is removed from the DOM on close, so keyboard focus is stranded on `document.body`. Restore focus to the trigger button when closing.

ebarti (commented):
Two well-scoped fixes. The TS bullet-to-evidence derivation correctly mirrors the canonical Python `resume_profile` logic (same metric regex, seniority terms, `{entry}_bullet_{n}` id, and hardcoded supported/0.8/confirmed contract), and link sanitization (`normalizeResumeHrefInput` rejects `javascript:`) plus the #370 hyperlink behavior are preserved. One correctness/auditability concern on the evidence half, and one minor a11y gap on the link popover.

ebarti:
PR Fixer - Codex

Fixed both review findings in 63f3fc7c. Materialized legacy bullet evidence now rederives from current bullets while explicit/manual evidence stays preserved, with API and Python regressions for the edit-after-materialize sequence. The resume link dialog now restores focus to the Insert link trigger when it closes.

Verification run:
- corepack pnpm --filter @jobctrl/api exec vitest run test/server.test.ts -t "profile achievement evidence|materialized bullet evidence" -> 1 file passed, 3 tests passed, 166 skipped.
- corepack pnpm --filter @jobctrl/web test -- src/views/apply-review/ApplyReviewView.test.tsx -t "renders the review workspace with job evidence and tailored materials" -> command completed with 215 files passed, 1204 tests passed.
- corepack pnpm --filter @jobctrl/api check -> passed.
- corepack pnpm --filter @jobctrl/web check -> passed.
- uv --project workers/automation run --extra dev pytest -q workers/automation/tests/test_profile_aggregate.py workers/automation/tests/test_sqlite_profile_repository.py -> 26 passed.
- uv --project workers/automation run --extra dev ruff check workers/automation/src/jobctrl/resume_profile.py workers/automation/src/jobctrl/infrastructure/profile/sqlite_repository.py workers/automation/tests/test_profile_aggregate.py workers/automation/tests/test_sqlite_profile_repository.py -> All checks passed.
- git diff --check -> passed.

Resolved inline threads: discussion_r3549424676 and discussion_r3549424681. Remaining Blocker/High findings: None.

apps/web/src/contexts/materials/components/ResumeAuditPins.tsx
```diff
@@ -2041,13 +2050,68 @@ function ResumeEditorToolbarControls({
...
+    setLinkError(null);
+    setLinkPopoverOpen(false);
+    window.requestAnimationFrame(() => linkTriggerRef.current?.focus());
```
ebarti, line 2073: **[Medium]** `closeLinkPopover` now unconditionally refocuses the trigger via rAF, but it is also the outside-dismiss handler: the `pointerdown` capture listener (line ~2087) calls `closeLinkPopover()` for any click outside the toolbar. So clicking into the resume editor to dismiss the popover closes it and then steals focus back to the "Insert link" button on the next frame, overriding the caret/selection the click just placed — the user has to click a second time to type. Returning focus to the trigger is correct for keyboard/programmatic dismissal (Escape, Cancel, Apply) but not for outside-pointer dismissal. Gate the refocus (e.g. a `restoreFocus` arg, or only refocus when the popover still owns `document.activeElement`) so an outside click lands where the user clicked.

ebarti (commented):
Re-review of the fix commit. The High evidence-staleness finding is genuinely resolved: both `achievementEvidenceForEntry` (TS) and `get_achievement_evidence`/`_reconciled_achievement_evidence` (Python) now re-derive materialized legacy-bullet evidence from current bullets while preserving genuinely manual evidence (fingerprint + `_bullet_N` id check), with matching edit-after-materialize regression tests on both sides. The Low focus finding is addressed but over-corrected into a new focus-stealing issue on outside-click dismissal (one comment below).

PR #374, opened 2026-07-09
fix(apply): bind artifact uploads to application origin

ebarti:
## Summary

- Pass the approved application URL into the apply-tools MCP server.
- Require `upload_artifact` to verify the current browser page is still on the approved application origin before calling `DOM.setFileInputFiles`.
- Reject hidden or ambiguous file inputs, record upload destination audit events, and update prompt/test coverage for the new fail-closed behavior.

## Root Cause

`upload_artifact` constrained the local artifact path but trusted the current browser page. A malicious apply-flow page could steer the browser to a different origin and then trigger resume or cover-letter upload to that destination.

## Validation

- `uv --project workers/automation run --extra dev pytest -q workers/automation/tests/test_apply_tools_mcp_server.py workers/automation/tests/test_apply_prompt_builder.py` - 22 passed
- `uv --project workers/automation run --extra dev pytest -q workers/automation/tests/test_apply_use_cases.py` - 12 passed
- `uv --project workers/automation run --extra dev ruff check workers/automation/src/jobctrl/infrastructure/apply_tools/mcp_server.py workers/automation/src/jobctrl/domain/apply/services.py workers/automation/src/jobctrl/apply/prompt.py workers/automation/tests/test_apply_tools_mcp_server.py workers/automation/tests/test_apply_prompt_builder.py` - passed
- `uv --project workers/automation run --extra dev ruff format workers/automation/src/jobctrl/infrastructure/apply_tools/mcp_server.py workers/automation/src/jobctrl/domain/apply/services.py workers/automation/src/jobctrl/apply/prompt.py workers/automation/tests/test_apply_tools_mcp_server.py workers/automation/tests/test_apply_prompt_builder.py --check` - passed
- `git diff --check` - passed

Docs were not updated because this is a security bug fix to an internal apply-tool guard and prompt behavior, not a new public capability or user-facing command.

### DIFF
--- a/workers/automation/src/jobctrl/infrastructure/apply_tools/mcp_server.py
+++ b/workers/automation/src/jobctrl/infrastructure/apply_tools/mcp_server.py
@@ -6,6 +6,8 @@ import json
 import os
 import sqlite3
 import sys
+import uuid
+from datetime import datetime, timezone
 from pathlib import Path
 from typing import Any
 from urllib.parse import urlparse
@@ -36,6 +38,7 @@ class ApplyToolsMcpServer:
         *,
         upload_dir: str | os.PathLike[str] | None = None,
         cdp_endpoint: str | None = None,
+        approved_application_url: str | None = None,
         uploader: Any | None = None,
         credential_resolver: Any | None = None,
         credential_typer: Any | None = None,
@@ -46,6 +49,11 @@ class ApplyToolsMcpServer:
         raw_upload_dir = upload_dir or os.environ.get("JOBCTRL_APPLY_UPLOAD_DIR")
         self._upload_dir = Path(raw_upload_dir).expanduser() if raw_upload_dir else None
         self._cdp_endpoint = cdp_endpoint or os.environ.get("JOBCTRL_APPLY_CDP_ENDPOINT", "")
+        self._approved_application_url = (
+            approved_application_url
+            if approved_application_url is not None
+            else os.environ.get("JOBCTRL_APPLY_APPROVED_APPLICATION_URL", "")
+        )
         self._uploader = uploader or _upload_file_to_current_input
         self._credential_resolver = credential_resolver or _profile_credential
         self._credential_typer = credential_typer or _type_credential_into_active_field
@@ -96,15 +104,27 @@ class ApplyToolsMcpServer:
     def _call_upload_artifact(self, args: dict[str, Any]) -> dict[str, Any]:
         kind = str(args.get("kind") or "")
         artifact = self._resolve_artifact(kind)
-        self._uploader(self._cdp_endpoint, artifact)
+        approved_url = self._approved_application_url.strip()
+        if not approved_url:
+            raise ValueError("upload_artifact approved application URL is not configured")
+        upload_result = self._uploader(self._cdp_endpoint, artifact, approved_url)
+        destination_url = _upload_destination_url(upload_result)
+        _record_artifact_upload(
+            self._upload_dir,
+            {
+                "kind": kind,
+                "filename": artifact.name,
+                "destination_url": destination_url,
+            },
+        )
+        payload = {"ok": True, "kind": kind, "filename": artifact.name}
+        if destination_url:
+            payload["destination_url"] = destination_url
         return {
             "content": [
                 {
                     "type": "text",
-                    "text": json.dumps(
-                        {"ok": True, "kind": kind, "filename": artifact.name},
-                        ensure_ascii=False,
-                    ),
+                    "text": json.dumps(payload, ensure_ascii=False),
                 }
             ]
         }
@@ -309,9 +329,20 @@ def _record_captcha_usage(upload_dir: Path | None, payload: dict[str, Any]) -> N
         fh.write(json.dumps(event, sort_keys=True) + "\n")
 
 
-def _inject_captcha_token(
-    cdp_endpoint: str, challenge: CaptchaChallenge, token: str
-) -> None:
+def _record_artifact_upload(upload_dir: Path | None, payload: dict[str, Any]) -> None:
+    if upload_dir is None:
+        return
+    upload_dir.mkdir(parents=True, exist_ok=True)
+    event = {
+        "event_type": "ApplyArtifactUpload",
+        "occurred_at": datetime.now(timezone.utc).isoformat(),
+        "payload": payload,
+    }
+    with (upload_dir / "artifact_upload_events.jsonl").open("a", encoding="utf-8") as fh:
+        fh.write(json.dumps(event, sort_keys=True) + "\n")
+
+
+def _inject_captcha_token(cdp_endpoint: str, challenge: CaptchaChallenge, token: str) -> None:
     escaped = json.dumps(token)
     _evaluate_on_page(
         cdp_endpoint,
@@ -420,7 +451,7 @@ def _evaluate_on_page(cdp_endpoint: str, expression: str) -> Any:
         ws.close()
 
 
-def _upload_file_to_current_input(cdp_endpoint: str, path: Path) -> None:
+def _upload_file_to_current_input(cdp_endpoint: str, path: Path, approved_application_url: str) -> dict[str, str]:
     ws_url = _first_page_ws_url(cdp_endpoint)
     try:
         import websocket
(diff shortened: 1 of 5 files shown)


workers/automation/src/jobctrl/infrastructure/apply_tools/mcp_server.py
```diff
@@ -443,23 +474,145 @@ def response(message_id: int) -> dict[str, Any]:
...
+    const rect = input.getBoundingClientRect();
+    return !input.disabled &&
+      style.display !== "none" &&
+      style.visibility !== "hidden" &&
+      Number(style.opacity || "1") > 0 &&
+      rect.width > 0 &&
+      rect.height > 0;
+  }};
+  const visible = inputs.filter(isVisible);
+  if (visible.length === 0) {{
```
ebarti: **[High]** This gate rejects any page whose `<input type=file>` is not itself visible, but the standard pattern on major ATS platforms (Greenhouse, Lever, Workday, Ashby) is a `display:none` / `opacity:0` / zero-size file input fronted by a visible "Select file" label — the real input the agent must populate is almost never visible on its own. `DOM.setFileInputFiles` works fine on hidden inputs, and the approved-origin check below already blocks cross-origin exfiltration, so requiring visibility here (plus the prompt change at prompt.py:492 telling the agent not to use hidden controls) will fail-closed on the normal upload path. Please verify against real ATS forms; consider allowing a hidden input when it is the sole `input[type=file]` on the already-origin-verified page.

workers/automation/src/jobctrl/infrastructure/apply_tools/mcp_server.py
```diff
@@ -96,15 +104,27 @@ def _call_tool(self, params: dict[str, Any]) -> dict[str, Any]:
     def _call_upload_artifact(self, args: dict[str, Any]) -> dict[str, Any]:
         kind = str(args.get("kind") or "")
         artifact = self._resolve_artifact(kind)
-        self._uploader(self._cdp_endpoint, artifact)
+        approved_url = self._approved_application_url.strip()
+        if not approved_url:
+            raise ValueError("upload_artifact approved application URL is not configured")
+        upload_result = self._uploader(self._cdp_endpoint, artifact, approved_url)
+        destination_url = _upload_destination_url(upload_result)
+        _record_artifact_upload(
```
ebarti, line 126: **[Medium]** `_record_artifact_upload` runs only after a successful upload. When `_assert_approved_upload_destination` rejects a steered/cross-origin destination (line 493) it raises before reaching here, so the security-relevant *blocked* attempt leaves no audit trace — which is exactly the event an auditability-focused apply flow most needs to capture. Consider also recording a denied/blocked upload event (kind, attempted destination_url, approved origin) on rejection.

workers/automation/src/jobctrl/infrastructure/apply_tools/mcp_server.py
```diff
@@ -443,23 +474,145 @@ def response(message_id: int) -> dict[str, Any]:
...
+                    "returnByValue": True,
+                    "expression": _upload_input_guard_expression(marker),
+                },
+            )
+        )
+        guard = state.get("result", {}).get("result", {}).get("value") or {}
+        if not isinstance(guard, dict):
+            raise RuntimeError("could not inspect current upload destination")
+        destination_url = str(guard.get("href") or "")
+        _assert_approved_upload_destination(destination_url, approved_application_url)
```
ebarti, line 516: **[Low]** The origin check runs against `guard.href` captured during the guard `Runtime.evaluate`, but the file is set later over separate CDP round-trips (`DOM.getDocument` / `querySelector` / `setFileInputFiles`), and the `data-jobctrl-upload-target` marker is a plain DOM attribute readable by main-world page script. The check is therefore non-atomic: a malicious approved-origin page could in principle read the marker and navigate cross-origin before `setFileInputFiles`, landing the file on a page whose origin was never the one validated. Exploitation needs winning a tight cross-navigation race (hence Low), but re-asserting the destination origin at set-file time (e.g. verifying the marked node's `ownerDocument` URL in the same evaluate) would close the window.

ebarti (commented):
Origin binding for artifact uploads is well-constructed: the unforgeable `location.href` read plus the per-element marker (looked up by CDP, not page script) makes cross-origin exfiltration hard to pull off. Three points below — a likely happy-path regression from rejecting hidden inputs (High), blocked uploads not being audited (Medium), and a non-atomic origin check (Low).

ebarti:
PR Fixer - Codex

Fixed findings:
- PRRT_kwDOSQ-EBM6PeId4 / discussion_r3549401735: allowed the standard ATS pattern where the sole enabled file input is hidden, while keeping multiple-input ambiguity fail-closed; updated the apply prompt so the model clicks the visible upload surface and the tool handles the backing input.
- PRRT_kwDOSQ-EBM6PeIeA / discussion_r3549401745: added a denied-upload audit event for rejected upload destinations with kind, filename, attempted destination, approved origin, reason, and blocked status.
- PRRT_kwDOSQ-EBM6PeIeG / discussion_r3549401754: replaced the page-readable marker lookup with a CDP object-handle flow and re-verified the selected input owner-document URL immediately before DOM.setFileInputFiles.

Dismissed findings: None.
Commit: 57cae04b (fix(apply): address artifact upload review findings)
Pushed: Yes, to origin/agent/apply-artifact-upload-origin-binding.

Verification:
- `uv --project workers/automation run --extra dev ruff format workers/automation/src/jobctrl/infrastructure/apply_tools/mcp_server.py workers/automation/src/jobctrl/apply/prompt.py workers/automation/tests/test_apply_tools_mcp_server.py workers/automation/tests/test_apply_prompt_builder.py` -> 4 files left unchanged
- `uv --project workers/automation run --extra dev pytest -q workers/automation/tests/test_apply_tools_mcp_server.py workers/automation/tests/test_apply_prompt_builder.py` -> 24 passed in 0.35s
- `uv --project workers/automation run --extra dev ruff check workers/automation/src/jobctrl/infrastructure/apply_tools/mcp_server.py workers/automation/src/jobctrl/apply/prompt.py workers/automation/tests/test_apply_tools_mcp_server.py workers/automation/tests/test_apply_prompt_builder.py` -> All checks passed!
- `uv --project workers/automation run --extra dev ruff format workers/automation/src/jobctrl/infrastructure/apply_tools/mcp_server.py workers/automation/src/jobctrl/apply/prompt.py workers/automation/tests/test_apply_tools_mcp_server.py workers/automation/tests/test_apply_prompt_builder.py --check` -> 4 files already formatted
- `git diff --check` -> passed with no output
- `uv --project workers/automation run --extra dev pytest -q workers/automation/tests/test_apply_use_cases.py` -> 12 passed in 0.18s

Remaining Blocker or High findings: None.

workers/automation/src/jobctrl/infrastructure/apply_tools/mcp_server.py
```diff
@@ -443,23 +495,191 @@ def response(message_id: int) -> dict[str, Any]:
...
+        )
+        current_guard = current_state.get("result", {}).get("result", {}).get("value") or {}
+        if not isinstance(current_guard, dict):
+            raise RuntimeError("could not verify current upload destination")
+        destination_url = str(current_guard.get("href") or "")
+        _assert_approved_upload_destination(destination_url, approved_application_url)
+        if not current_guard.get("ok"):
+            reason = str(current_guard.get("reason") or "selected file input is no longer available")
+            raise RuntimeError(reason)
+        node = response(send("DOM.requestNode", {"objectId": input_object_id}))
```
ebarti, line 541: **[Medium]** The rework dropped the `DOM.getDocument` call the previous code ran, and now resolves the input handle via `DOM.requestNode` -> `nodeId` -> `DOM.setFileInputFiles`. In Chrome, `DOM.requestNode`'s node-path resolution generally requires `DOM.getDocument` to have seeded the frontend node map first; without it `requestNode` commonly returns `nodeId: 0`, which line 543 turns into "selected file input is no longer available" — i.e. real uploads would fail on the happy path. The new tests mock `DOM.requestNode` (canned `nodeId: 7`), so this step has no non-mocked coverage. Simplest robust fix: skip `requestNode` and pass the handle straight to `DOM.setFileInputFiles` via its `objectId` parameter (the protocol accepts `objectId` in place of `nodeId`), which also drops a round-trip; alternatively re-add a one-time `DOM.getDocument` after `DOM.enable`.

ebarti (commented):
Re-review at 57cae04b — all three prior findings are resolved: the sole hidden+enabled file input is now allowed while multi-input ambiguity stays fail-closed (High), blocked cross-origin attempts now write a `status:"blocked"` audit event (Medium), and the page-readable DOM marker was replaced with a CDP object-handle plus an owner-document re-verification immediately before the upload (Low). One new point below on the reworked node-resolution sequence.

PR #376, opened 2026-07-09
fix(apply): bind credential typing to approved origin

ebarti:
## Summary

- Bind the owned apply `type_credential` tool to the approved application origin derived from the apply target URL.
- Fail closed when the credential-origin policy is missing or when the active browser page origin does not match before inserting the stored password.
- Add focused regression coverage for missing policy, mismatched origin rejection, approved-origin typing, and MCP config wiring.

## Validation

- `uv --project workers/automation run --extra dev pytest -q workers/automation/tests/test_apply_tools_mcp_server.py workers/automation/tests/test_apply_prompt_builder.py` (`22 passed`)
- `uv --project workers/automation run --extra dev ruff check workers/automation/src/jobctrl/infrastructure/apply_tools/mcp_server.py workers/automation/src/jobctrl/domain/apply/services.py workers/automation/tests/test_apply_tools_mcp_server.py workers/automation/tests/test_apply_prompt_builder.py`
- `git diff --check`

## Docs

No documentation update: this is an internal safety-boundary fix for the owned apply-tools MCP server and does not add user-facing commands, configuration, or workflow steps.

### DIFF
--- a/workers/automation/src/jobctrl/infrastructure/apply_tools/mcp_server.py
+++ b/workers/automation/src/jobctrl/infrastructure/apply_tools/mcp_server.py
@@ -7,6 +7,7 @@ import os
 import sqlite3
 import sys
 import uuid
+from collections.abc import Iterable
 from datetime import datetime, timezone
 from pathlib import Path
 from typing import Any
@@ -42,6 +43,7 @@ class ApplyToolsMcpServer:
         uploader: Any | None = None,
         credential_resolver: Any | None = None,
         credential_typer: Any | None = None,
+        allowed_credential_origins: Iterable[str] | None = None,
         captcha_key_resolver: Any | None = None,
         captcha_solver: Any | None = None,
         captcha_injector: Any | None = None,
@@ -57,6 +59,11 @@ class ApplyToolsMcpServer:
         self._uploader = uploader or _upload_file_to_current_input
         self._credential_resolver = credential_resolver or _profile_credential
         self._credential_typer = credential_typer or _type_credential_into_active_field
+        self._allowed_credential_origins = _normalize_allowed_origins(
+            allowed_credential_origins
+            if allowed_credential_origins is not None
+            else _credential_origins_from_env()
+        )
         self._captcha_key_resolver = captcha_key_resolver or _captcha_api_key
         self._captcha_solver = captcha_solver or _solve_with_capsolver
         self._captcha_injector = captcha_injector or _inject_captcha_token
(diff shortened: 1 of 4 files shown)


workers/automation/src/jobctrl/infrastructure/apply_tools/mcp_server.py
```diff
@@ -386,7 +443,9 @@ def response(message_id: int) -> dict[str, Any]:
             )
         )
         active = guard.get("result", {}).get("result", {}).get("value") or {}
-        if not active.get("ok"):
+        if str(active.get("origin") or "") not in origin_policy:
+            raise RuntimeError("stored credential is not approved for this page origin")
+        if not active.get("fieldOk"):
             raise RuntimeError("active element is not a password credential field")
         response(send("Input.insertText", {"text": credential}))
```
ebarti, line 502: **[Medium]** Check-to-use gap: `window.location.origin` is validated in the `Runtime.evaluate` guard above, but the password is typed in a *separate* `Input.insertText` round-trip with no re-check. A page-initiated navigation or redirect (common in SSO apply flows, e.g. a session-timeout bounce) that commits between the two CDP calls types the stored job_site_password into the newly-loaded origin, defeating the binding this PR adds. Re-verify `window.location.origin` immediately before the insert, or perform the origin/field check and the value entry in one atomic evaluation.

workers/automation/src/jobctrl/infrastructure/apply_tools/mcp_server.py
```diff
@@ -386,7 +443,9 @@ def response(message_id: int) -> dict[str, Any]:
             )
         )
         active = guard.get("result", {}).get("result", {}).get("value") or {}
-        if not active.get("ok"):
+        if str(active.get("origin") or "") not in origin_policy:
```
ebarti, line 498: **[Low]** When no input is focused the guard JS returns `{ok: false}` (line 429) with no `origin`, so `active.get('origin')` is empty and this line raises `stored credential is not approved for this page origin` even on the approved origin, masking the real cause. Pre-PR code reported this case accurately as `active element is not a password credential field`. Return `origin` in the `{ok: false}` branch, or evaluate `fieldOk` before the origin comparison.

ebarti (commented):
Origin binding for credential typing is a solid fail-closed improvement. Two guard-level issues below: a check-to-use gap between the origin evaluation and the credential insert, and a misleading error when no field is focused.

PR #377, opened 2026-07-09
fix(security): guard apply browser private navigations

ebarti:
## Summary

- Install a live CDP public-destination guard for non-dry-run apply Chrome sessions.
- Reuse the same public-destination enforcement inside the existing dry-run guard before dry-run submission controls run.
- Add focused regressions proving loopback/private requests are failed and public HTTP(S) requests continue.

## Root cause

The apply flow checked the initial application URL before prompt construction, but live Chrome sessions did not keep enforcing the public-destination boundary after the autonomous browser was running. A malicious job or ATS page could steer allowed navigation/snapshot tools toward loopback or private-network resources.

## Validation

- `uv --project workers/automation run --extra dev pytest -q workers/automation/tests/test_apply_chrome_dry_run_guard.py` (7 passed)
- `uv --project workers/automation run --extra dev ruff check workers/automation/src/jobctrl/apply/chrome.py workers/automation/tests/test_apply_chrome_dry_run_guard.py`
- `uv --project workers/automation run --extra dev pytest -q workers/automation/tests/test_apply_use_cases.py::test_execute_blocks_private_network_application_url_before_agent` (1 passed)
- `git diff --check`

Docs not updated: this is a bug fix to an existing safety boundary, not a new user-facing capability.

### DIFF
--- a/workers/automation/src/jobctrl/apply/chrome.py
+++ b/workers/automation/src/jobctrl/apply/chrome.py
@@ -16,6 +16,7 @@ from urllib.parse import urlparse
 from urllib.request import urlopen
 
 from jobctrl import config
+from jobctrl.infrastructure.network import validate_public_http_url
 
 logger = logging.getLogger(__name__)
 
@@ -27,6 +28,8 @@ _chrome_procs: dict[int, subprocess.Popen] = {}
 _chrome_lock = threading.Lock()
 _dry_run_guards: dict[int, "_DryRunCdpGuard"] = {}
 _dry_run_guards_lock = threading.Lock()
+_public_destination_guards: dict[int, "_PublicDestinationCdpGuard"] = {}
+_public_destination_guards_lock = threading.Lock()
 
 
 # ---------------------------------------------------------------------------
@@ -251,6 +254,8 @@ def launch_chrome(worker_id: int, port: int | None = None,
     time.sleep(3)
     if dry_run:
         install_dry_run_cdp_guard(port)
+    else:
+        install_public_destination_cdp_guard(port)
     logger.info("[worker-%d] Chrome started on port %d (pid %d)",
                 worker_id, port, proc.pid)
     return proc
@@ -353,10 +358,19 @@ _FORM_SUBMIT_GUARD_SOURCE = """
 
 
 def install_dry_run_cdp_guard(port: int) -> "_DryRunCdpGuard":
-    """Install a CDP dry-run guard on Chrome pages for this worker."""
+    """Install the public-destination + dry-run CDP guard for this worker."""
     guard = _DryRunCdpGuard(port)
     with _dry_run_guards_lock:
         _dry_run_guards[int(port)] = guard
+    guard.start()
+    return guard
+
+
+def install_public_destination_cdp_guard(port: int) -> "_PublicDestinationCdpGuard":
+    """Install the live apply public-destination guard on Chrome pages."""
+    guard = _PublicDestinationCdpGuard(port)
+    with _public_destination_guards_lock:
+        _public_destination_guards[int(port)] = guard
     guard.start()
     return guard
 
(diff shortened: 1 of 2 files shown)


workers/automation/src/jobctrl/apply/chrome.py
```diff
@@ -510,18 +540,19 @@ def send(method: str, params: dict | None = None) -> None:
...
             except websocket.WebSocketTimeoutException:
                 continue
             message = json.loads(raw_message)
             method_name = message.get("method")
-            if method_name == "Runtime.bindingCalled":
+            if guard.enforce_dry_run and method_name == "Runtime.bindingCalled":
                 _record_binding_call(message, guard)
                 continue
-            if method_name == "Network.webSocketCreated":
+            if guard.enforce_dry_run and method_name == "Network.webSocketCreated":
```
ebarti, line 555: **[Medium]** In live (non-dry-run) mode this now gates `Network.webSocketCreated` behind `enforce_dry_run`, and CDP `Fetch` interception does not pause WebSocket handshakes, so `validate_public_http_url` never sees them. A hostile ATS/job page in a live apply session can therefore open `new WebSocket("ws://127.0.0.1:8766/...")` (or any private/link-local host) and reach internal services unchecked, which is exactly the SSRF class this PR targets. `Fetch.failRequest` can't abort a WS, but the live guard should at least validate/record WS destinations via `Network.webSocketCreated` (or the WS vector should be explicitly documented as out of scope).

workers/automation/src/jobctrl/apply/chrome.py
```diff
@@ -353,14 +358,23 @@ def launch_chrome(worker_id: int, port: int | None = None,
...
         _dry_run_guards[int(port)] = guard
     guard.start()
     return guard
 
 
+def install_public_destination_cdp_guard(port: int) -> "_PublicDestinationCdpGuard":
+    """Install the live apply public-destination guard on Chrome pages."""
+    guard = _PublicDestinationCdpGuard(port)
+    with _public_destination_guards_lock:
+        _public_destination_guards[int(port)] = guard
```
ebarti, line 373: **[Medium]** `_public_destination_guards[port]` is written here but never read: there is no `get_public_destination_cdp_guard_evidence`, and `local_chrome.py` only wires `dry_run_evidence` when `config.dry_run` is true (`local_chrome.py:63-67`). So when the live guard blocks a private/loopback navigation during a real apply run, the `record_blocked_request` evidence is collected and then dropped — the security event leaves no auditable trace on the apply run. Given this repo's auditability discipline, expose the live guard's evidence and attach it to the non-dry-run `BrowserSession`.

workers/automation/src/jobctrl/apply/chrome.py
```diff
@@ -543,7 +575,16 @@ def send(method: str, params: dict | None = None) -> None:
             url = str(request.get("url") or "")
             resource_type = str(params.get("resourceType") or "Other")
             initiator_type = guard.request_initiator(network_id)
-            if request_id and _should_block_dry_run_request(
+            public_decision = validate_public_http_url(url)
```
ebarti, line 578: **[Medium]** `validate_public_http_url` resolves the host itself (getaddrinfo) and then `Fetch.continueRequest` hands the request back to Chrome, which re-resolves and connects — the vetted IP is not pinned. A DNS-rebinding host (low TTL: public A record during this check, private A record for Chrome's actual connection) bypasses the guard. This is the same TOCTOU as the existing `PublicHttpUrlRouteGuard`, so it may be an accepted limitation, but since this adds a new enforcement point it is worth acknowledging/documenting (or pinning the resolved address into the connection).

ebarti (commented):
Solid CDP-layer public-destination guard for live apply sessions: it reuses the same `validate_public_http_url` validator as the merged #368 and correctly re-checks every intercepted request at the Request stage (so redirect hops and in-page navigations are re-validated, not just the first goto), which is the right layer. Three gaps/limitations worth resolving before merge: WebSocket destinations are not covered in live mode, the live guard's block evidence is collected but never surfaced, and the check is subject to DNS-rebinding TOCTOU.

PR #378, opened 2026-07-09
fix(apply): block private apply browser navigation

ebarti:
## Summary

- installs a CDP-level public-destination guard for live apply browser sessions
- reuses the same public-destination enforcement inside dry-run browser sessions before dry-run submission controls
- adds regression coverage for live guard installation, dry-run guard behavior, blocked loopback requests, and allowed public requests

## Root Cause

The apply flow validated the initial application URL before starting the browser, but live browser navigation remained model-driven after launch. A page could steer the apply browser toward loopback, private-network, or cloud-metadata destinations after the initial check. The fix moves the existing public HTTP(S) destination invariant to the CDP request boundary so unsafe navigations are failed before Chrome continues them.

## Validation

- `uv --project workers/automation run --extra dev pytest -q workers/automation/tests/test_apply_chrome_dry_run_guard.py` - 7 passed
- `uv --project workers/automation run --extra dev ruff check workers/automation/src/jobctrl/apply/chrome.py workers/automation/tests/test_apply_chrome_dry_run_guard.py` - passed
- `uv --project workers/automation run --extra dev pytest -q workers/automation/tests/test_apply_use_cases.py::test_execute_blocks_private_network_application_url_before_agent` - passed
- `git diff --check` - passed

### DIFF
--- a/workers/automation/src/jobctrl/apply/chrome.py
+++ b/workers/automation/src/jobctrl/apply/chrome.py
@@ -16,6 +16,7 @@ from urllib.parse import urlparse
 from urllib.request import urlopen
 
 from jobctrl import config
+from jobctrl.infrastructure.network import validate_public_http_url
 
 logger = logging.getLogger(__name__)
 
@@ -27,6 +28,8 @@ _chrome_procs: dict[int, subprocess.Popen] = {}
 _chrome_lock = threading.Lock()
 _dry_run_guards: dict[int, "_DryRunCdpGuard"] = {}
 _dry_run_guards_lock = threading.Lock()
+_public_destination_guards: dict[int, "_PublicDestinationCdpGuard"] = {}
+_public_destination_guards_lock = threading.Lock()
 
 
 # ---------------------------------------------------------------------------
@@ -251,6 +254,8 @@ def launch_chrome(worker_id: int, port: int | None = None,
     time.sleep(3)
     if dry_run:
         install_dry_run_cdp_guard(port)
+    else:
+        install_public_destination_cdp_guard(port)
     logger.info("[worker-%d] Chrome started on port %d (pid %d)",
                 worker_id, port, proc.pid)
     return proc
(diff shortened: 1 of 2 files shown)


workers/automation/src/jobctrl/apply/chrome.py
```diff
@@ -353,14 +358,23 @@ def launch_chrome(worker_id: int, port: int | None = None,
...
         _dry_run_guards[int(port)] = guard
     guard.start()
     return guard
 
 
+def install_public_destination_cdp_guard(port: int) -> "_PublicDestinationCdpGuard":
+    """Install the live apply public-destination guard on Chrome pages."""
+    guard = _PublicDestinationCdpGuard(port)
+    with _public_destination_guards_lock:
+        _public_destination_guards[int(port)] = guard
```
ebarti, line 373: **[Medium]** `_public_destination_guards` is written here but never read anywhere else — there is no `get_public_destination_cdp_guard_evidence` (only `get_dry_run_cdp_guard_evidence` exists, and `local_chrome.py` wires only the dry-run evidence). So when the live guard blocks an SSRF navigation via `Fetch.failRequest`, the `evidence()` / `blocked_channels` it records are unreachable outside tests: the block leaves no audit trail (this repo's auditability discipline expects one), and this module-level dict is dead write-only state (the running daemon threads already keep the guard alive, so it isn't needed as a keep-alive ref). Either surface the live guard's evidence via a getter consumed on the apply audit path, or drop the dict.

workers/automation/src/jobctrl/apply/chrome.py
```diff
@@ -543,7 +575,16 @@ def send(method: str, params: dict | None = None) -> None:
             url = str(request.get("url") or "")
             resource_type = str(params.get("resourceType") or "Other")
             initiator_type = guard.request_initiator(network_id)
-            if request_id and _should_block_dry_run_request(
+            public_decision = validate_public_http_url(url)
```
ebarti, line 578: **[Low]** `validate_public_http_url(url)` now runs on every paused request (both live and dry-run modes) inside this per-page loop, and it calls `socket.getaddrinfo` synchronously with no timeout (getaddrinfo ignores `socket.setdefaulttimeout`). Each request re-resolves the host with no per-host memoization, adding DNS latency to the live-apply hot path and serializing requests behind resolution; a slow or hostile resolver can stall the page's guard thread. Consider memoizing the decision per host and/or bounding resolution time.

ebarti, line 578: **[Low]** This validates by re-resolving DNS, then (allowed case) hands the *original* URL to Chrome via `Fetch.continueRequest`, which resolves and connects independently. A low-TTL attacker domain can return a public address to this `getaddrinfo` and a private one to Chrome's connect (DNS-rebinding TOCTOU), bypassing the guard. This is the same limitation as the pre-launch #368 check, so not a regression; fully closing it needs pinning the validated IP (rewrite to the resolved address + set `Host`) instead of re-resolving. At minimum a code comment so the guarantee isn't overstated.

ebarti (commented):
Reviewed the live-apply CDP public-destination guard against main (which already blocks the *initial* apply URL via #368). The core `validate_public_http_url` SSRF coverage is solid and fail-closed (private/loopback/link-local incl. 169.254.169.254, IPv6 ULA/link-local, IPv4-mapped IPv6, CGNAT 100.64/10, and alt encodings like 2130706433 via getaddrinfo are all blocked on current 3.11+ patches). The new protection is genuine defense-in-depth on every post-launch navigation. Three integration notes below (1 Medium, 2 Low); nothing blocking.

PR #379, opened 2026-07-09
fix(security): guard LinkedIn apply resolver URLs

ebarti:
## Summary

- Validate the LinkedIn apply resolver starting URL before opening a browser page.
- Install the existing public HTTP(S) Playwright route guard around authenticated LinkedIn apply navigation and loaded-page click capture.
- Validate redirect, href, current-page, and popup targets before returning an external apply URL.
- Add focused offline regression coverage for unsafe redirect targets, pre-navigation blocking, route-guard blocking, and unsafe href rejection.

## Root Cause

The authenticated LinkedIn apply resolver reused the public-destination URL policy indirectly in nearby enrichment paths, but did not enforce that policy across its own browser navigation and captured target URLs. A non-public target could be reached or returned after the initial LinkedIn page was loaded.

## Documentation

No documentation update: this is a security bug fix that preserves the existing public-destination safety model rather than adding a new user-facing capability or command.

## Validation

- `PYTHONDONTWRITEBYTECODE=1 PYTHONPATH=$PWD/workers/automation/src /Users/eloibarti/Github/JobHunter/workers/automation/.venv/bin/python -m pytest -q workers/automation/tests/test_linkedin_apply_resolver.py` -> `8 passed`
- `PYTHONDONTWRITEBYTECODE=1 PYTHONPATH=$PWD/workers/automation/src /Users/eloibarti/Github/JobHunter/workers/automation/.venv/bin/ruff check workers/automation/src/jobctrl/infrastructure/enrichment/linkedin_apply_resolver.py workers/automation/tests/test_linkedin_apply_resolver.py` -> passed
- `git diff --check` -> passed

### DIFF
--- a/workers/automation/src/jobctrl/infrastructure/enrichment/linkedin_apply_resolver.py
+++ b/workers/automation/src/jobctrl/infrastructure/enrichment/linkedin_apply_resolver.py
@@ -14,10 +14,15 @@ import shutil
 import time
 from dataclasses import dataclass
 from pathlib import Path
-from typing import Any
+from typing import Any, Callable
 from urllib.parse import parse_qsl, unquote, urljoin, urlparse
 
 from jobctrl import config
+from jobctrl.infrastructure.network import (
+    PublicHttpUrlRouteGuard,
+    PublicUrlDecision,
+    validate_public_http_url,
+)
 from jobctrl.infrastructure.network.proxy import ProxyConfig
 
 log = logging.getLogger(__name__)
@@ -45,6 +50,8 @@ _EXTERNAL_URL_PARAM_NAMES = {
     "externalurl",
 }
 _APPLY_TEXT_RE = re.compile(r"\b(apply|solicitar)\b", re.IGNORECASE)
+
+UrlSafetyChecker = Callable[[str], PublicUrlDecision]
 
 
 @dataclass(frozen=True)
@@ -119,6 +126,7 @@ class LinkedInApplyUrlResolver:
         user_agent: str | None = None,
         playwright: Any | None = None,
         chrome_profile: str | None = None,
+        url_safety_checker: UrlSafetyChecker | None = None,
     ) -> None:
         self._profile_dir = profile_dir or default_linkedin_apply_profile_dir()
         self._chrome_profile = chrome_profile or linkedin_apply_chrome_profile()
@@ -127,6 +135,7 @@ class LinkedInApplyUrlResolver:
         self._timeout_seconds = timeout_seconds or linkedin_apply_timeout_seconds()
         self._user_agent = user_agent
         self._external_playwright = playwright
+        self._url_safety_checker = url_safety_checker or validate_public_http_url
         self._playwright: Any | None = None
         self._context: Any | None = None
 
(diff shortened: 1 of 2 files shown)


workers/automation/src/jobctrl/infrastructure/enrichment/linkedin_apply_resolver.py
```diff
@@ -200,7 +209,12 @@ def new_page(self) -> Any:
     def resolve(self, job_url: str) -> LinkedInApplyResolution:
         """Navigate to ``job_url`` and capture the external apply target."""
 
+        initial_safety = self._url_safety_checker(job_url)
+        if not initial_safety.allowed:
+            return _unsafe_url_resolution(initial_safety.reason)
+
         page = self.new_page()
+        route_guard = PublicHttpUrlRouteGuard(page).install()
```
ebarti, line 217: **[High]** The route guard is installed on `page` only, but `_click_and_capture_apply_target` opens a popup via `page.expect_popup()` (line ~412) and reads/validates `popup.url`. Playwright `page.route` does not apply to popups — they are separate Page objects — so an attacker-controlled Apply button that opens `window.open('http://169.254.169.254/...')` (or any loopback/private URL) still causes the authenticated browser to issue that request. That is the exact SSRF-in-browser this guard is meant to prevent, on the primary external-apply path (external ATS almost always opens a new tab/popup). The returned URL is now safe because `_wait_for_external_apply_url` re-validates it, but the request is still emitted unguarded. Install the guard at context scope (`self._context.route("**/*", ...)` in `start()`), or attach `PublicHttpUrlRouteGuard(popup).install()` immediately after `popup = popup_info.value`.

ebarti (commented):
In-scope security hardening that closes the return-value leak: every returned apply URL now passes the public-destination policy on all paths (`current_url`, `href`, redirect extraction, and click capture). Two gaps remain — the click-capture popup is a separate Playwright page the page-scoped route guard does not cover, and the success-path `blocked` check can discard an already-validated public result.

workers/automation/src/jobctrl/infrastructure/enrichment/linkedin_apply_resolver.py
```diff
@@ -210,9 +224,15 @@ def resolve(self, job_url: str) -> LinkedInApplyResolution:
                 except Exception:
                     pass
             except Exception as exc:
+                if route_guard.blocked:
+                    return _unsafe_url_resolution(route_guard.blocked_reason)
                 return LinkedInApplyResolution(None, "navigation_error", str(exc)[:300])
-            return self.resolve_loaded_page(page, job_url)
+            result = self._resolve_loaded_page(page, job_url)
+            if route_guard.blocked:
```
ebarti, line 231: **[Medium]** `route_guard.blocked` is sticky and is set by ANY intercepted non-public request during the entire page lifetime (including benign third-party subresources loaded during the LinkedIn/ATS navigation). Gating the success path on it means a validly captured, public apply URL (`current_url`/`href_redirect`/`href`/`click`, each already run through `validate_public_http_url` via `_is_safe_external_apply_url`) is silently discarded and returned as `unsafe_url`. This diverges from the established guard usage in `_scrape_detail_page_body` (detail.py ~460), which checks `blocked` only in the navigation-exception branch and does not gate a successful result. The same pattern repeats at line 247 in `resolve_loaded_page`. Consider relying on the per-candidate validation for the returned URL and using `blocked` only for the navigation-error branch (as detail.py does).

ebarti:
Closed as superseded by #380, which merged the verified LinkedIn apply resolver public-destination guard for the same Codex Security finding.

PR #380, opened 2026-07-09
fix(security): guard LinkedIn apply resolver destinations

ebarti:
## Summary
- Validate LinkedIn apply resolver input URLs before opening an authenticated browser page.
- Install a public-destination route guard while capturing LinkedIn apply redirects, hrefs, popups, and navigations.
- Reject non-public extracted apply targets and add regression coverage for unsafe job URLs, redirects, route-blocked navigations, and unsafe hrefs.

## Root Cause
The authenticated LinkedIn apply resolver used a persistent browser context and trusted LinkedIn-derived navigation targets before applying the same public HTTP(S) destination checks used elsewhere. A crafted job URL, redirect parameter, href, popup, or post-click navigation could therefore point the authenticated browser at loopback or private network targets.

## Validation
- `uv --project workers/automation run --extra dev pytest -q workers/automation/tests/test_linkedin_apply_resolver.py`
- `uv --project workers/automation run --extra dev ruff check workers/automation/src/jobctrl/infrastructure/enrichment/linkedin_apply_resolver.py workers/automation/tests/test_linkedin_apply_resolver.py`
- `git diff --check`

### DIFF
--- a/workers/automation/src/jobctrl/infrastructure/enrichment/linkedin_apply_resolver.py
+++ b/workers/automation/src/jobctrl/infrastructure/enrichment/linkedin_apply_resolver.py
@@ -214,7 +214,7 @@ class LinkedInApplyUrlResolver:
             return _unsafe_url_resolution(initial_safety.reason)
 
         page = self.new_page()
-        route_guard = PublicHttpUrlRouteGuard(page).install()
+        route_guard = _install_public_route_guard(page, context=self._context)
         try:
             try:
                 page.goto(job_url, timeout=_NAV_TIMEOUT_MS)
@@ -241,7 +241,7 @@ class LinkedInApplyUrlResolver:
     def resolve_loaded_page(self, page: Any, job_url: str) -> LinkedInApplyResolution:
         """Capture the apply target from an already-loaded LinkedIn page."""
 
-        route_guard = PublicHttpUrlRouteGuard(page).install()
+        route_guard = _install_public_route_guard(page)
         try:
             result = self._resolve_loaded_page(page, job_url)
             if route_guard.blocked:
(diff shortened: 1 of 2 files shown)


workers/automation/src/jobctrl/infrastructure/enrichment/linkedin_apply_resolver.py
```diff
@@ -380,12 +420,14 @@ def _click_and_capture_apply_target(
                 popup,
                 original_job_url,
                 timeout_seconds=timeout_seconds,
+                url_safety_checker=url_safety_checker,
```
ebarti, line 441: **[High]** The `PublicHttpUrlRouteGuard` is installed only on `page` (lines 217 and 244), but the apply popup opened just above at line 412 (`popup = popup_info.value`) is a separate Playwright page whose network requests are NOT routed through the guard — `page.route` is page-scoped and there is no `context.route` anywhere in the worker. A crafted LinkedIn apply flow that opens a popup to a loopback/metadata target (e.g. `http://127.0.0.1:8766/...`, `http://169.254.169.254/...`) will cause the authenticated browser to actually issue that request during the popup load; only the *returned* URL is later validated in `_wait_for_external_apply_url`, so the blind request has already fired even though the value is ultimately rejected. This is exactly the "point the authenticated browser at loopback or private network targets" threat the PR sets out to close, and the popup is the primary external-apply path from LinkedIn. Fix: install a guard on the popup (`PublicHttpUrlRouteGuard(popup).install()` before `wait_for_load_state`), or preferably add a `context.route("**/*", ...)` guard at context creation so popups are covered from their first navigation — a page-level route installed after the popup object exists can miss the initial request. Add a popup regression test alongside the existing route-guard test.

ebarti (commented):
One security gap on the popup capture path: the public-destination route guard is installed on the opener page but not on the popup it opens, so the authenticated browser can still be driven to a private/loopback destination at the network layer even though the returned URL is validated. Otherwise the change is sound — it correctly reuses the shared `validate_public_http_url` / `PublicHttpUrlRouteGuard` (same guard as merged #368, no divergent implementation), enforces http(s)-only, and fails closed on parse/DNS failures.

workers/automation/src/jobctrl/infrastructure/enrichment/linkedin_apply_resolver.py
ebarti: Fixed in 0db6f4c9 by installing the public URL route guard on the browser context when available, falling back to the page for test doubles/external pages. Added a regression that fires an unsafe popup navigation during the click, before any popup page-level guard could be installed.

ebarti:
PR Fixer - Codex

Fixed the High popup route-guard finding in 0db6f4c9 by installing the public URL route guard on the browser context when available, with page fallback for externally supplied pages/test doubles. Added a regression for unsafe popup navigation during the click path. Resolved inline thread PRRT_kwDOSQ-EBM6PeIF6 / discussion_r3549399658 after push.

Verification:
- uv --project workers/automation run --extra dev pytest -q workers/automation/tests/test_linkedin_apply_resolver.py -> 9 passed in 0.15s
- uv --project workers/automation run --extra dev ruff check workers/automation/src/jobctrl/infrastructure/enrichment/linkedin_apply_resolver.py workers/automation/tests/test_linkedin_apply_resolver.py -> All checks passed!
- git diff --check -> clean

Remaining Blocker/High findings: None.

PR #391, opened 2026-07-10
fix(docs): contain product tour media and add sidebar controls

ebarti:
## Summary

- constrain Product Tour media to the actual document track instead of viewport width
- prevent screenshots from overlapping the right-side page outline
- add accessible desktop sidebar resize, reset, collapse, persistence, and focus restoration
- preserve the stock mobile navigation fallback
- add rendered geometry and keyboard regressions for the reported breakpoints

## Why

The custom media width calculation ignored the 256px right outline. At 1285px it produced roughly 199px of overlap and made prose/media widths inconsistent. The fixed-width left navigation also could not be adjusted or fully collapsed.

## Validation

- overlap reduced to zero at 1440, 1285, 1024, 720px zoom proxy, and 390px mobile
- mouse and keyboard resize limits, reset, collapse, reload persistence, and focus restoration: passed
- docs build: 5,025 references resolved
- docs runtime assertions: passed
- release privacy check: passed

## Stack

2 of 4. Base: `fix/docs-profile-tour-capture`. Next: `fix/keychain-runtime-credentials`.


docs/.vitepress/theme/JobCtrlLayout.vue
```diff
@@ -0,0 +1,128 @@
...
+  sidebarExpanded.value = true;
+  await nextTick();
+  collapseButton.value?.focus();
+}
+
+function resetSidebarWidth(): void {
+  sidebarWidth.value = SIDEBAR_DEFAULT_WIDTH;
+}
+
+onMounted(() => {
```
ebarti, line 57: Low / non-blocking: the persisted preference (`jobctrl-docs-sidebar`) is applied only here in `onMounted`, i.e. after the SSG first paint. A returning visitor who collapsed the rail or set a non-default width will see the default state (expanded, 272px) render first, then a visible layout shift to their stored state on hydration. There is no pre-paint `head` script (config.ts `head[]` has none) that sets `data-jh-sidebar-expanded` / `--jh-sidebar-width` before first paint the way VitePress avoids the dark-mode flash. Consider inlining a tiny head script that reads localStorage and sets the html attribute + CSS var before paint. Only affects returning users who changed the default — first-time visitors match SSG — so this is minor.

ebarti (commented):
Reviewed the PR-391 diff over its base (fix/docs-profile-tour-capture): the `:not(.jh-outline-page)` media-track fix correctly stops screenshots overlapping the outline on the outline-classed screenshots/normal-flows pages, and the sidebar resize/collapse/persist controls are well-built and verified against VitePress 1.6.4 internals (mobile drawer width is independent of `--vp-sidebar-width`, both slots exist, collapse rules are correctly scoped to >=960px, focus restoration is sound). One Low, non-blocking note below.

PR #394, opened 2026-07-10
docs: plan bundled JobCtrl distribution

ebarti:
## Summary

- add the active bundled JobCtrl distribution plan
- record the owner decision in a superseding ADR
- register the plan in the active and historical plan indexes

## Why

The current install surface has no production distribution boundary. Curl installs a source checkout, Homebrew installs a shell launcher over that checkout, startup commands differ by channel, CLI examples expose the uv-managed Python environment, Git is required only to clone/update source, and user setup installs contributor and duplicate browser dependencies.

The plan moves the fix to the owning build, distribution, launcher, lifecycle, and capability layers instead of treating this as a README wording problem.

## Planned product contract

- one native `jobctrl` executable for startup, lifecycle, and existing CLI commands
- the same signed runtime artifact after curl or Homebrew installation
- no user-facing Git, Node, pnpm, Corepack, uv, Python, Temporal, Poppler, Playwright, npx, or checkout requirement
- one managed core Playwright Chromium revision
- system Chrome and authenticated browser profiles only for explicitly enabled capabilities
- versioned updates, rollback, and uninstall that preserve `~/.jobctrl` state
- official-channel managed packs for components JobCtrl cannot legally redistribute

This PR is documentation-only. The README remains unchanged until the packaged behavior is implemented and verified.

## Validation

- `corepack pnpm docs:build`
- `python3 scripts/release_check.py`
- `git diff --check`

PR #396, opened 2026-07-10
build(distribution): lock bundled release contract

ebarti:
## What changed

- define the signed distribution manifest schema, exact external archive locks, platform floor, capability defaults, and signing-channel policy
- classify every top-level runtime, embedded asset, provider pack, and developer-only component with fail-closed redistribution decisions
- add deterministic file/component inventory validation, including single ownership, ASCII canonical ordering, safe Chromium symlinks, modes, hashes, sizes, and schema parity
- separate reproducible dependency counts from observational source-install footprint evidence
- record that Claude, Codex, and Antigravity are managed official-download provider packs until their embedded redistribution closure is established

## Why

JobCtrl currently distributes a mutable source checkout and installs a contributor toolchain. The accepted distribution plan needs a machine-enforced boundary before a production payload or native launcher can be built.

## Impact

This PR does not claim that a user artifact exists yet. It establishes P0 of the stacked plan and deliberately reports stable signing as blocked until release and Apple credentials are provisioned.

## Validation

- `node --test scripts/distribution-manifest.test.mjs` — 17/17
- `corepack pnpm scripts:test` — 37/37
- `corepack pnpm docs:build` — 5,025 links
- `git diff --check`
- clean reconstructed checkout: frozen install, audit, tests, measure, docs, and release check
- pinned Playwright Chromium rev1208 inventory: 333 entries, five safe relative symlinks, modes limited to 0644/0755

## Stack

Depends on #394. P1 will build and measure the production payload against this contract.

### DIFF
--- a/scripts/distribution-manifest.mjs
+++ b/scripts/distribution-manifest.mjs
@@ -1,0 +1,1078 @@
+#!/usr/bin/env node
+
+import { createHash } from "node:crypto";
+import { createReadStream } from "node:fs";
+import { lstat, readFile, readdir, readlink, realpath, stat } from "node:fs/promises";
+import path from "node:path";
+import process from "node:process";
+import { fileURLToPath, pathToFileURL } from "node:url";
+import Ajv2020 from "ajv/dist/2020.js";
+
+const SCRIPT_DIR = path.dirname(fileURLToPath(import.meta.url));
+export const REPO_ROOT = path.resolve(SCRIPT_DIR, "..");
+
+const CLASSIFICATIONS = new Set([
+  "core-runtime",
+  "optional-capability",
+  "provider-pack",
+  "developer-only",
+]);
+const REDISTRIBUTION_MODES = new Set(["bundle", "official-download", "exclude"]);
+const MANIFEST_CLASSIFICATIONS = new Set([
+  "core-runtime",
+  "optional-capability",
+  "provider-pack",
+]);
+const RELEASE_CHANNELS = new Set(["local", "prerelease", "stable"]);
+const SHA256_PATTERN = /^[a-f0-9]{64}$/;
+const ID_PATTERN = /^[a-z0-9][a-z0-9-]{1,63}$/;
+const VERSION_PATTERN = /^[0-9]+\.[0-9]+\.[0-9]+(?:[-+][0-9A-Za-z.-]+)?$/;
+const PLATFORM_ID_PATTERN = /^[a-z0-9]+-[a-z0-9_]+$/;
+const OS_VERSION_PATTERN = /^[0-9]+(?:\.[0-9]+){1,2}$/;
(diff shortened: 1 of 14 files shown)


scripts/distribution-manifest.mjs
```diff
@@ -0,0 +1,1078 @@
...
+    invariant(version, `${filePath}: ${packageName} has no version`);
+    return version;
+  }
+  throw new Error(`${filePath}: package ${packageName} not found`);
+}
+
+async function resolvePnpmLockVersion(filePath, packageName) {
+  const contents = await readFile(filePath, "utf8");
+  const escapedName = packageName.replace(/[.*+?^${}()|[\]\\]/g, "\\$&");
+  const match = contents.match(new RegExp(`^  ['\"]?${escapedName}@([^:'\"(]+)(?:\\([^\\n]+\\))?['\"]?:$`, "m"));
```
ebarti, line 480: PR Reviewer - review-pr-396

Severity: Low
Gate: Pass
Evidence: `resolvePnpmLockVersion` uses `String.prototype.match` (single/first match). `pdfjs-dist` already appears twice in `pnpm-lock.yaml` (lines 5565 and 12514) — both `5.7.284` today, so `pdfjs-renderer` resolves correctly only by coincidence. `resolveUvLockVersion` (line ~465) has the same first-match behavior for `uv.lock`.
Reproduction: Reasoned from code; confirmed the duplicate `pdfjs-dist` lock entries and that `pnpm distribution:audit` resolves `pdfjs-renderer` to `5.7.284`.

If a version-sourced package ever resolves to more than one distinct version in the lockfile (e.g. a second major pulled in transitively), the resolver silently returns whichever entry sorts first (pnpm orders by `name@version`, so the lower version), and that value becomes the enforced component version in the signed manifest via the `component.version === contracts.versions[id]` check in `validateManifestComponent`. Nothing detects the ambiguity, so a signed manifest could bind the wrong upstream version — a contract that the p1+ payload/build phases consume.

Required fix: collect all matches (e.g. `matchAll`) and fail closed when a version-source selector matches more than one distinct version (apply the same to `resolveUvLockVersion`), or resolve against the workspace's actually-installed version; add a regression fixture with a duplicated-but-divergent lock entry proving the resolver rejects it rather than picking first-in-file.

ebarti (commented):
P0 contract review. The distribution contract (schema, inventory, capability/platform/lock/signing policies, license review, source baseline) and the validator are well-built and fail-closed: `distribution:audit` exits 0 and `node --test scripts/distribution-manifest.test.mjs` is 17/17 in the checked-out worktree, all 20 version sources resolve, and the docs/decisions.md Go ADR is satisfied upstream in #394. One Low, non-blocking observation below on the lockfile version resolvers. Gate: PASS.

PR #400, opened 2026-07-10
feat(launcher): centralize bundled runtime lifecycle

ebarti:
## What changed

- add the native Darwin jobctrl launcher with one lifecycle and Python CLI command surface
- verify the complete local distribution envelope and exact payload tree before dispatch
- supervise Temporal, worker, and API with loopback health gates, ownership-safe stop/recovery, bounded logs, and cross-process state
- compile with the checksum-locked Go 1.26.4 toolchain and add pinned launcher CI
- document the bundled runtime boundary while preserving contributor source commands

## Why

Installed users need one real jobctrl executable that works from any directory without exposing Node, pnpm, uv, Python, Temporal, or source-checkout commands.

## Validation

- corepack pnpm check
- corepack pnpm test: 64 script, 517 API, 29 extension unit, 4 extension e2e, 2,282 Python tests, web production build, launcher race suite
- corepack pnpm docs:build: 5,839 references across 262 files
- python3 scripts/release_check.py --strict-prompt
- independent review: Gate PASS
- independent QA: Gate PASS
- two full real builds from clean output directories were byte-identical
  - archive SHA-256: 0ac0d1bd2c4d4911018437dd8ad7f3c2fb6f47af5b4c527a13e56b2e26a90d3f
  - manifest SHA-256: b31295f553051e77a71001257b32f23596b8411bc82c8141f2f6be39fc23f973
  - compressed: 501,740,370 bytes; installed: 1,572,088,598 bytes
- lifecycle smoke covers bare PATH invocation, start/status/logs/version, Python dispatch, worker-kill recovery, supervisor-kill recovery, restart, exact-tree verification, and zero leaked fixed-port listeners

## Stack

Depends on #399.

### DIFF
--- a/launcher/go.mod
+++ b/launcher/go.mod
@@ -1,0 +1,3 @@
+module github.com/ebarti/jobctrl/launcher
+
+go 1.24.0
--- a/launcher/internal/launcher/launcher.go
+++ b/launcher/internal/launcher/launcher.go
@@ -1,0 +1,1756 @@
+// Package launcher owns the installed JobCtrl process boundary. It deliberately
+// has no domain logic: it validates payload metadata, manages bundled processes,
+// and transparently dispatches the embedded Python CLI.
+package launcher
+
+import (
+	"bufio"
+	"crypto/sha256"
+	"encoding/hex"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io"
+	"net"
+	"net/http"
+	"os"
+	"os/exec"
+	"os/signal"
+	pathpkg "path"
+	"path/filepath"
+	"regexp"
+	"sort"
+	"strconv"
+	"strings"
+	"syscall"
+	"time"
+)
+
+var (
+	sha256Pattern        = regexp.MustCompile(`^[a-f0-9]{64}$`)
+	componentIDPattern   = regexp.MustCompile(`^[a-z0-9][a-z0-9-]{1,63}$`)
+	buildIDPattern       = regexp.MustCompile(`^[0-9A-Za-z][0-9A-Za-z._-]{7,127}$`)
+	semverPattern        = regexp.MustCompile(`^[0-9]+\.[0-9]+\.[0-9]+(?:[-+][0-9A-Za-z.-]+)?$`)
+	minimumOSPattern     = regexp.MustCompile(`^[0-9]+(?:\.[0-9]+){1,2}$`)
+	requiredComponentIDs = []string{"jobctrl-launcher", "jobctrl-api", "jobctrl-web", "jobctrl-worker", "node-runtime", "python-runtime", "temporal-runtime"}
+)
+
+const (
+	launcherProtocol   = 1
+	stateSchemaVersion = 1
+	logLineLimit       = 200
+	startupTimeout     = 30 * time.Second
+	shutdownTimeout    = 5 * time.Second
+)
+
+var (
+	ErrAlreadyRunning     = errors.New("an active JobCtrl instance already owns this state directory")
+	ErrPortInUse          = errors.New("a required fixed loopback port is already in use")
+	ErrLockHeld           = errors.New("the JobCtrl instance lock is held")
+	errStartupInterrupted = errors.New("supervisor startup interrupted")
+)
+
+type commandSpec struct {
+	Name       string   `json:"name"`
+	Executable string   `json:"executable"`
+	Arguments  []string `json:"arguments"`
+}
+
+type runtimeManifest struct {
+	SchemaVersion    int `json:"schemaVersion"`
+	LauncherProtocol int `json:"launcherProtocol"`
+	Ports            struct {
+		TemporalGRPC int `json:"temporalGrpc"`
+		TemporalUI   int `json:"temporalUi"`
+		API          int `json:"api"`
+	} `json:"ports"`
+	Components []commandSpec `json:"components"`
+	Health     struct {
+		Temporal struct {
+			Component string   `json:"component"`
+			Arguments []string `json:"arguments"`
+		} `json:"temporal"`
+		API struct {
+			Path                 string `json:"path"`
+			RequireWorkerHealthy bool   `json:"requireWorkerHealthy"`
+			RequireWorkerPID     bool   `json:"requireWorkerPid"`
+		} `json:"api"`
+	} `json:"health"`
+}
+
+// This is the full v1 public envelope shape. Strict decoding means a launcher
+// never makes a permissive map-shaped assumption about a signed artifact.
+type distributionManifest struct {
+	SchemaVersion   int    `json:"schemaVersion"`
+	AppVersion      string `json:"appVersion"`
+	BuildID         string `json:"buildId"`
+	ReleaseChannel  string `json:"releaseChannel"`
+	SourceDateEpoch int64  `json:"sourceDateEpoch"`
+	Platform        struct {
+		ID               string `json:"id"`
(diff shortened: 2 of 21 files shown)


ebarti (commented):
PR Reviewer - review-pr-400

Reviewed feat/distribution-p2 (b02cdd16) against the P2 phase contract in the distribution plan (PR #394). The launcher, cross-process registry, PID-reuse/start-identity guards, foreign-listener refusal, identity-scoped process-group termination, environment isolation, and lock/orphan-recovery paths are well-built and thoroughly tested. Findings: 1 High (detached-start readiness timeout of 35s is shorter than the supervisor's sequential 30s+30s startup budget, so a valid slow/cold start is spuriously timed out and torn down), 2 Medium (full payload-tree SHA-256 runs before --help/version and on every read-only lifecycle command and Python dispatch; jobctrl logs fails entirely on any log line over 64 KiB), 1 Low (go.mod language directive lags the pinned Go 1.26.4 toolchain). Details inline.

launcher/internal/launcher/launcher.go
```diff
@@ -0,0 +1,1756 @@
...
+		if message.Error != "" {
+			_ = command.Wait()
+			return errors.New(message.Error)
+		}
+		// The detached supervisor is intentionally not a child that this short
+		// command waits on. Release its handle after the readiness handshake so
+		// a successful `start` cannot leave a waitable child behind.
+		_ = command.Process.Release()
+		return nil
+	case <-time.After(startupTimeout + 5*time.Second):
```
ebarti: PR Reviewer - review-pr-400

Severity: High
Gate: Fail
Evidence: startDetached waits time.After(startupTimeout + 5*time.Second) = 35s (launcher.go:1124), but supervise runs waitForTemporal (deadline startupTimeout=30s, :1343) and THEN waitForAPIWorker (another 30s, :1384) sequentially before it calls sendReady(nil) at :1245. startupTimeout is 30s (:42), so the supervisor's own worst-case budget before signaling readiness is ~60s.
Reproduction: Reasoned from code (timing constants + sequential control flow).

Default `jobctrl start` is detached. When Temporal plus API/worker together take longer than 35s but still within their own 30s+30s budgets (plausible on a cold/first start where Temporal dev-server init, Python worker import, and Node API startup add up), the parent times out, SIGTERMs the still-healthy-in-progress supervisor (:1125), and the supervisor's deferred cleanup tears down the whole stack. The user sees "timed out waiting for detached JobCtrl supervisor" even though startup was seconds from success. This directly undermines the clean/cold-machine start scenario the plan targets.

Required fix: Make the parent readiness deadline cover the supervisor's total worst-case startup (at least 2*startupTimeout plus margin), or have the supervisor emit progressive readiness/heartbeat messages that extend the parent's wait, so a valid slow startup is never killed mid-flight.

launcher/internal/launcher/launcher.go
```diff
@@ -0,0 +1,1756 @@
...
+var readProcessIdentity processIdentityReader = processStartIdentity
+var readProcessExecutable processIdentityReader = processExecutable
+var signalProcessGroup groupSignaler = func(pgid int, signal syscall.Signal) error { return syscall.Kill(-pgid, signal) }
+var openBrowser = func(url string) error { return exec.Command("/usr/bin/open", url).Run() }
+var temporalHealthProbe = probeTemporal
+
+// Run is the only public CLI entrypoint. __supervise is reachable only through
+// the re-exec made by `jobctrl start`.
+func Run(executable string, args, inheritedEnv []string, stdout, stderr io.Writer) error {
+	ctx, err := prepare(executable, inheritedEnv)
```
ebarti, line 210: PR Reviewer - review-pr-400

Severity: Medium
Gate: Fail
Evidence: Run() calls prepare() unconditionally here (launcher.go:206) before the help branch (:213). prepare -> loadAndVerifyDistributionManifest (:318, :665) -> verifyPayloadTree, which SHA-256s every regular file in the payload (~1.5 GiB installed per the PR body). This runs for --help, help, no-args, version, status, logs, and every Python dispatch.

Two problems: (1) jobctrl --help/help/no-args require a fully present, hash-verified payload just to print usage, so help fails outright on a partial/corrupt payload and always pays multi-second full-tree hashing, contradicting the verifyPayloadTree comment that scopes verification to "before any lifecycle action or Python dispatch" (help is neither). (2) Read-only lifecycle commands (status, logs, version) re-hash the entire installed tree on every call, a large latency regression for commands that users and scripts poll frequently. The recorded state already carries ManifestSHA256, and validateStateBinding already compares it via the cheap manifestDigest (manifest.json only, :966).

Required fix: Print help/usage before calling prepare; scope full-tree verifyPayloadTree to start and Python dispatch, and bind read-only lifecycle commands to the recorded ManifestSHA256 (cheap) instead of re-hashing the whole payload on every invocation.

launcher/internal/launcher/launcher.go
```diff
@@ -0,0 +1,1756 @@
...
+	file, err := os.Open(path)
+	if errors.Is(err, os.ErrNotExist) {
+		return nil
+	}
+	if err != nil {
+		return err
+	}
+	defer file.Close()
+	lines := make([]string, 0, limit)
+	scanner := bufio.NewScanner(file)
```
ebarti, line 1707: PR Reviewer - review-pr-400

Severity: Medium
Gate: Fail
Evidence: tailFile uses bufio.NewScanner(file) (launcher.go:1703) with the default 64 KiB max token size. A single log line longer than 64 KiB makes scanner.Scan() stop and scanner.Err() return bufio.ErrTooLong, so tailFile returns that error (:1712) and prints nothing.
Reproduction: Reasoned from code (bufio.MaxScanTokenSize = 64*1024).

jobctrl logs is the primary diagnostic surface for a bundled user with no source checkout. Component stdout/stderr is redirected verbatim to the log files (:1327); Node/Python worker records that serialize large JSON/LLM payloads or long tracebacks onto a single line can exceed 64 KiB - exactly the failure conditions where logs are most needed. When that happens the command errors and shows no output at all instead of the last 200 lines. TestLogsAreBoundedAndNonFollowing only exercises short lines, so this path is untested.

Required fix: Read lines with a bufio.Reader (ReadString newline) or raise scanner.Buffer to a bounded larger cap and truncate overlong lines, so a single long line degrades gracefully rather than failing the whole command; add a regression test with a line over 64 KiB.

launcher/go.mod
```diff
@@ -0,0 +1,3 @@
+module github.com/ebarti/jobctrl/launcher
+
+go 1.24.0
```
ebarti, line 3: PR Reviewer - review-pr-400

Severity: Low
Gate: Pass
Evidence: launcher/go.mod declares go 1.24.0, while launcher/toolchain.json (goVersion go1.26.4), package.json launcher:check/launcher:test (test "$(go env GOVERSION)" = "go1.26.4"), and .github/workflows/launcher.yml (go-version 1.26.4) all pin 1.26.4.

Release reproducibility is enforced by the pinned GOROOT and the go version assertion in prepareNativeGoToolchain, so this is not a build-output bug. But the module's language directive being 1.24.0 while the entire toolchain story says 1.26.4 is a maintainability inconsistency: a contributor with GOTOOLCHAIN=auto and a different local Go will not auto-select 1.26.4 from this go.mod, and the declared language level silently lags the pinned compiler.

Required fix: Set the go.mod directive to match the pinned toolchain (go 1.26.4, optionally with a toolchain go1.26.4 line) so the declared language version and the pinned compiler agree.

PR #401, opened 2026-07-10
feat(browser): make authenticated capabilities explicit

ebarti:
## Summary

- separate the bundled core browser from optional authenticated browser capabilities
- require explicit browser-path and profile consent before enabling system-browser workflows
- gate apply and LinkedIn browser entry points at workflow, activity, adapter, saga, and launch boundaries
- make mid-run capability revocation fail closed while terminating only the exact owned browser process
- persist capability state atomically under an inter-process lock and harden copied-profile handling against symlink and parent-swap attacks
- remove system Chrome from core setup, doctor, and default dependency requirements

## Why

The bundled product must work without a system Chrome installation. Authenticated automation is an optional, higher-risk capability and must never be enabled implicitly or treated as a core dependency.

## Validation

- `corepack pnpm check`
- `corepack pnpm test`
  - 64 distribution/script tests
  - 517 API tests
  - production web build
  - 29 extension unit tests
  - 4 extension e2e tests
  - 2,316 Python tests
  - Go launcher race suite
- focused capability and browser-safety suite: 180 passed
- additional core/doctor/bundled regression suite: 53 passed
- `corepack pnpm docs:build` (5,839 references across 262 emitted files)
- real bundled distribution build and smoke from the rebased head
  - archive SHA-256: `ea10ab5efc36bdadf89d8d0775bb3c585cdec5f6be5fce269d0415d1f5d0a94d`
  - manifest SHA-256: `e48999231fc088350ce25afc80552081043b0ca890abaca8fe5d2333b6682731`
  - exact post-smoke manifest match and zero leaked runtime processes

No live application submission or external authenticated-browser workflow was executed; safety validation used loopback and isolated fixtures.

Stacked on #400.


workers/automation/src/jobctrl/apply/auto_apply.py
```diff
@@ -73,6 +74,16 @@ async def reconcile_auto_apply_loop(
     workflow_id = auto_apply_workflow_id(tenant_id)
     running = await _is_workflow_running(client, workflow_id)
     if settings.enabled:
+        capability = browser_capability_status("auto-apply-browser")
```
ebarti, line 77: PR Reviewer - review-pr-401

Severity: Medium
Gate: Pass
Evidence: `reconcile_auto_apply_loop` is `async` and runs on the same event loop as `worker.run()` and the 15s heartbeat loop (cli.py:2257 and cli.py:2311). When auto-apply is enabled with an adopted browser, `browser_capability_status("auto-apply-browser")` synchronously calls `_executable_status` -> `subprocess.run(["<chrome>", "--version"], timeout=3)` plus state/catalog file reads (browser_capabilities.py), with no `asyncio.to_thread`/executor offload.
Reproduction: Reasoned from code. Note the reconciler tests monkeypatch `browser_capability_status`, so they stub out the real subprocess and would not surface this.

This synchronous readiness probe blocks the shared worker event loop on every reconcile cycle (startup + every 15s) for opted-in users — typically ~100ms, but up to 3s if the adopted Chrome hangs — which can delay Temporal task/heartbeat processing. Impact is bounded (only when the capability is enabled; the disabled path returns before probing), but it is an event-loop-blocking call inside the worker's supervisory loop.

Required fix: Run the readiness probe off the event loop, e.g. `capability = await asyncio.to_thread(browser_capability_status, "auto-apply-browser")`.

ebarti (commented):
Reviewed PR #401 (p3 capability split) against the distribution plan's Phase 3 contract. Gating is fail-closed and defense-in-depth across the workflow-start, activity, saga (submit-intent), adapter, and Chrome-launch boundaries, plus per-request CDP revocation; the profile-copy path is well hardened (no-follow FD traversal, atomic anchored rename, consent required, and TOCTOU/symlink/parent-swap tests). One Medium concurrency note inline. No Blocker/High found.

PR #402, opened 2026-07-11
feat(installer): unify secure release acquisition

ebarti:
## Summary

- add a native macOS installer that verifies signed release descriptors, manifests, archive identity, safe ZIP extraction, and notarized executable policy before activation
- make the curl bootstrap transport-only and expose the same native `jobctrl` selector used by the installed payload
- generate Homebrew formulae from the verified release contract, with no developer-toolchain dependencies or HEAD-only source install
- keep stable publication fail-closed until the protected signing and notarization phase provisions production keys and assets

## Why

JobCtrl should install and start through one bundled CLI without requiring Git, Node, pnpm, Python, uv, Temporal, or a system browser. This phase establishes one verified artifact and one acquisition contract for both curl and Homebrew while preserving the publication gate.

## Validation

- `corepack pnpm check`
- `corepack pnpm test`
- `corepack pnpm scripts:test` (72/72)
- `go vet ./...` and `go test -race ./...` under `launcher/`
- `corepack pnpm docs:build` (5,839 references across 262 files)
- `python3 scripts/release_check.py` and its 25-test regression suite
- two clean real distribution builds produced byte-identical installer, ZIP, manifest, descriptor, signatures, size report, and build evidence
- the real 484 MiB ZIP / 1.47 GiB installed payload passed embedded Node, Python, Temporal, worker/API, Playwright MCP, Chromium PDF, restart/recovery, offline-network, exact-tree, and forbidden-content smoke checks
- the generated local bootstrap contract installed the real artifact twice; the selector reported the expected build and immutable receipts remained byte-identical

## Stack

- depends on #401


launcher/internal/installer/installer.go
```diff
@@ -0,0 +1,915 @@
...
+	}
+	if descriptor.Channel != options.Policy.ExpectedChannel {
+		return Receipt{}, fmt.Errorf("release channel %q does not match compiled installer channel %q", descriptor.Channel, options.Policy.ExpectedChannel)
+	}
+	return installArchive(options, descriptor, descriptorBytes, sourceURL.String(), archivePath)
+}
+
+func strictHTTPClient() *http.Client {
+	return &http.Client{
+		Timeout: 45 * time.Second,
```
ebarti: PR Reviewer - review-pr-402

Severity: High
Gate: Fail
Evidence: strictHTTPClient() sets http.Client.Timeout = 45s (installer.go:201) and that same client is reused for the release-archive body read in downloadArchive() (installer.go:327). Go's Client.Timeout includes reading the response body, so the entire archive download must complete within 45s. The PR description states the real artifact is a 484 MiB ZIP; 484 MiB / 45s is about 90 Mbps sustained, so any connection slower than that fails the whole `curl | sh` install with a timeout mid-download. P4 has no resume (deferred to P5), so a slow link fails on every retry. This breaks the P4 exit criterion 'Curl installation passes on a clean supported macOS user account.' It is latent today because the network path is fail-closed until P6 provisions keys/pins, but it should be fixed before the network path is enabled.
Reproduction: Reasoned from code (Go http.Client.Timeout semantics plus the stated 484 MiB artifact size). Not run: no live signed release endpoint exists before P6.

The 45s client-wide timeout is right for the small descriptor/.sig fetches but far too short for the multi-hundred-MiB archive body read, which shares the same client.

Required fix: Split the transports. Keep the short overall Timeout only for fetchBounded (descriptor and signature), and download the archive with either no client-level Timeout (a Transport with Dial/TLSHandshake/ResponseHeader timeouts plus an idle/stall watchdog) or a per-request context deadline scaled to the expected artifact size, so a slow-but-progressing transfer is not killed.

ebarti (commented):
One High finding on the archive download timeout; otherwise the acquisition chain is strong. The checksum is anchored to an Ed25519-signed descriptor verified against a pinned key (not the transport), redirects are forbidden, HTTPS is enforced, the artifact URL is origin-bound to the descriptor, ZIP extraction has solid zip-slip/symlink/zip-bomb/case-collision defenses re-checked post-extraction by the signed manifest tree walk, staging is atomic/idempotent/locked, and verification failures hard-fail. Homebrew reuses the same signed descriptor->sha256 (Ed25519-verified at render) plus codesign/notarization and launch-time manifest verification. Security test coverage is thorough.

PR #403, opened 2026-07-11
feat(launcher): add transactional release lifecycle

ebarti:
## What changed

- add one user-owned immutable release store shared by curl and Homebrew acquisition
- implement signed channel floors/revocations, descriptor-bound resumable downloads, and receipt-bound policy evidence
- make `jobctrl update`, rollback, paired SQLite backup/restore, retention, and uninstall native launcher operations
- health-gate promotion, recover every durable activation boundary, and fail closed after a security revocation
- keep Homebrew installation prefix-only while first invocation seeds the same user store used by curl
- preserve local data by default and remove only the exact curl/Homebrew exposure owned by the acquisition record

## Why

The bundled distribution needs one command and one runtime lifecycle regardless of acquisition method. Updates must not depend on mutating a Homebrew Cellar, run an older payload against migrated state, or strand a user between a revoked predecessor and an unhealthy candidate.

## Validation

- `cd launcher && go vet ./...`
- `cd launcher && go test -race -count=1 ./...`
- `corepack pnpm check`
- `corepack pnpm test` (73 script tests, 517 API tests, web/extension builds and tests, 2,315 Python tests, launcher race suite)
- `corepack pnpm docs:build` (5,839 references across 262 emitted files)
- two real bundled artifact builds during final hardening; final build passed native lifecycle, Chromium, PDF, offline-network, restart/recovery, exact-tree, and forbidden-content smokes
- real `scripts/get` local-fixture install run twice with unchanged `active.json`, followed by `jobctrl uninstall`; runtime removed and local data preserved
- independent final code-review and QA gates: PASS, no Blocker/High findings

Final local artifact evidence:

- manifest SHA-256: `776f8d7168da9e25b23837e6b21434056ca31d17fd08428f7d5a8d28697157dc`
- archive SHA-256: `1c410c735577ad8bd68c615bf2de3085888ae40dca90e9ba8de86a28d9d75a40`
- compressed: 507,847,585 bytes
- installed: 1,578,963,342 bytes

## Stack

Depends on #402 (`feat/distribution-p4`).

### DIFF
--- a/launcher/internal/installer/installer.go
+++ b/launcher/internal/installer/installer.go
@@ -24,11 +24,14 @@ import (
 	pathpkg "path"
 	"path/filepath"
 	"regexp"
+	"sort"
+	"strconv"
 	"strings"
 	"syscall"
 	"time"
 
 	"github.com/ebarti/jobctrl/launcher/internal/launcher"
+	"github.com/ebarti/jobctrl/launcher/internal/release"
 )
 
 const (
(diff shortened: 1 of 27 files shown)


launcher/internal/installer/installer.go
```diff
@@ -312,88 +365,323 @@ func verifyDescriptor(raw, signatureRaw []byte, trust map[string]ed25519.PublicK
...
+		}
+		offset, appendToPartial = 0, false
+	} else {
+		return "", fmt.Errorf("unexpected HTTP status %s while resuming archive", response.Status)
 	}
+	if err := writeArchiveResponse(partial, response.Body, offset, expectedSize, appendToPartial); err != nil {
+		return "", err
+	}
+	if err := verifyArchiveFile(partial, expectedSize, expectedSHA256); err != nil {
+		return "", fmt.Errorf("archive size or SHA-256 does not match signed release descriptor: %w", err)
```
ebarti, line 560: PR Reviewer - review-pr-403

Severity: Medium
Gate: Pass
Evidence: downloadArchiveToStage. When a completed download fails the SHA-256/size check here (line 481-483), or when an already full-size partial fails it on re-entry (line 431-434), archive.zip.part is left on disk at exactly expectedSize and is never removed. On the next invocation the `offset == expectedSize` branch (line 431) re-verifies the same corrupt bytes and returns the same error, and no path re-requests the archive. cleanupStaging only deletes `stage-*` dirs, not the descriptor-digest stage, so the corrupt cache persists across `jobctrl update` retries. TestDescriptorBoundArchiveCacheRejectsTamperAndOversize only asserts the first call fails; it never asserts a retry can recover.
Reproduction: Reasoned from code + confirmed the tamper test asserts only single-call failure (installer_test.go:248-279).

A single corrupted-but-correct-length archive (bad CDN edge, bit rot, or a torn resume append that reaches full size) permanently bricks the resumable-download path until the user manually deletes the staging directory. The error message ("...does not match signed release descriptor") misleadingly implies a signing/tamper problem rather than a re-downloadable local cache, giving the user no clear remediation. This defeats the intent of the P5 "resumable staging" deliverable. The active release is unaffected, so the user is not stranded between versions; scope is limited to future updates. The logic is currently behind the P6 network gate (AcquisitionBuildPolicy.AllowNetwork) but ships and is tested in this PR.

Required fix: On a digest/size mismatch after writing (line 481-483) and for the on-entry full-size partial (line 431-434), remove archive.zip.part before returning the error so a subsequent attempt re-downloads instead of re-verifying the same corrupt bytes. Add a regression test that a second downloadArchiveToStage call after a corrupt full-size response succeeds against a now-healthy server.

ebarti (commented):
Reviewed the P5 transactional release lifecycle (staging, atomic activation, health-gated promotion, paired SQLite backup/restore, rollback, retention, uninstall) against the distribution plan's Phase 5 contract. The core state machine, crash-point recovery, lock ordering, and both-ways paired-DB rollback are correct and very well tested. One actionable robustness finding on the resumable download's corruption-recovery path is inline.

PR #404, opened 2026-07-11
build(release): secure bundled distribution publication

ebarti:
## Summary

- add the owner-dispatched signed distribution workflow with two independent unsigned builds, clean comparison, protected signing/notarization, immutable publication, public smoke, channel CAS, Homebrew sync, and integrated PyPI Trusted Publishing
- make the native installer consume signer-authenticated channel pointers and immutable descriptors, with exact rollback protection and fail-closed macOS trust checks
- reduce the redistributable browser core to Playwright headless shell only; system Chrome remains optional for authenticated and auto-apply capabilities
- add deterministic release-authority bundles, third-party notices, strict release/privacy checks, and publication runbooks

## Why

The installed product needs one bundled `jobctrl` interface without Git, pnpm, Corepack, uv, or a full Chrome application on the user machine. Publication must preserve that artifact identity across curl, Homebrew, channel promotion, GitHub Releases, and PyPI without giving dependency-bearing jobs signing or publication authority.

## Validation

- `corepack pnpm check`
- `corepack pnpm test` — 93 script, 517 API, 29 extension, 4 extension e2e, 2,326 Python, and Go race tests passed
- `corepack pnpm --filter @jobctrl/web test` — 1,241 passed
- `corepack pnpm --filter @jobctrl/web test-d` — 13 passed with no type errors
- `corepack pnpm docs:build` — 5,839 references across 262 emitted files
- strict release check and 34 release-check regression tests
- actionlint on both release workflows
- two isolated Python 3.12.13 release builds were byte-identical
- final real prerelease candidate: 348,628,078-byte archive, 1,220,103,559 installed bytes, 6,193 files, full runtime/restart/PDF/headless smoke passed, no full Chrome/Chromium app, no Widevine
- independent review and QA gates passed with no remaining findings

## External release prerequisites

No public release was created. Live Developer ID signing/notarization, channel publication, tap push, and PyPI publication remain fail-closed until the protected credentials and release origin are configured. Immutable GitHub Releases are currently disabled, the release origin currently fails TLS, and hosted Actions jobs are blocked by the account billing/spending state.

## Stack

Base: #403 (`feat/distribution-p5`).


packaging/homebrew/Formula/jobctrl.rb.tmpl
```diff
@@ -31,7 +31,7 @@ class Jobctrl < Formula
     sha256 JOBCTRL_SIGNATURE_SHA256
   end
 
-  def verify_notarized_bundle!(bundle)
+  def verify_notarized_app!(bundle)
```
ebarti, line 34: PR Reviewer - review-pr-404

Severity: Low
Gate: Pass
Evidence: packaging/homebrew/Formula/jobctrl.rb.tmpl:34 defines `verify_notarized_app!`, but the rendered `install` (lines 67-69) and the `test do` block (99-102) only call `verify_notarized_executable!`. The sole other references — scripts/distribution-homebrew.mjs:160 and the tracked render bundle — are negative assertions that it is NOT called on the launcher. Meanwhile `validateRenderedHomebrewFormula` still mandates this method's `--deep` codesign and `spctl ... source=Notarized Developer ID` body markers, so the render validator now pins the body of an uninvoked method.
Reproduction: Reasoned from code + `grep -rn verify_notarized_app` over the worktree (only the definition and the two negative assertions match).

`verify_notarized_app!` is vestigial from when the core shipped full Chrome `.app` bundles. After the headless-shell-only change, `brew install` verifies only three raw executables via `verify_notarized_executable!` (which is the correct `-R=notarized` check for non-app Mach-O shapes; authoritative full-payload trust remains the native installer's `verifyMacOSPayloadTrust`). So this is not a security gap, but the dead method combined with validator markers that still require it is a maintainability trap: a future reader can conclude Gatekeeper (`spctl`) still runs during `brew install`, when nothing invokes it.

Required fix: Either drop `verify_notarized_app!` from the template and remove the now-orphaned `--deep`/`spctl`/`source=Notarized Developer ID` entries from `validateRenderedHomebrewFormula`'s required-markers list, or re-wire the method into an install path if an app-bundle capability pack is intended. Keep the formula and its render validator in agreement.

ebarti (commented):
One Low maintainability finding: the rendered Homebrew formula's `verify_notarized_app!` is now dead code (only `verify_notarized_executable!` is invoked after the headless-shell-only change), yet `validateRenderedHomebrewFormula` still mandates its `spctl`/`--deep` body. No Blocker/High issues: workflow security (SHA-pinned actions, credential-free vs. signing/publication job isolation, channel-pointer CAS + immutable-release enforcement, reproducible dual-builds, integrated PyPI OIDC gating) and installer artifact-naming consistency all hold up.

PR #409, opened 2026-07-11
feat(demo): add browser workspace

ebarti:
## Summary

- adds an explicit local/demo composition root while preserving the existing local adapters for local and invalid modes
- adds a versioned IndexedDB demo workspace with atomic mutations, generated-blob ownership, migrations, browser-profile sharing, tab-local fallback, deterministic reset, and typed runtime warnings
- adds BroadcastChannel post-commit coordination backed by authoritative storage rereads and the existing domain-event/invalidation path
- adds an epoch-fenced deterministic deadline scheduler for later scenario slices
- makes Demo mode and its browser-profile/personal-data boundary persistently visible, including on mobile
- adds a committed native Playwright lane for persistence, profile isolation, reset/blob deletion, future-database refusal, event ordering, and product-network isolation
- documents the browser workspace development and QA contract

## Why

The public demo needs an isolated browser-local authority before read adapters or simulated workflows can become interactive. This slice establishes that authority without exposing the local API, SSE endpoint, filesystem, host OS, or any external product capability.

## Validation

- focused workspace/adapter/composition suite: 5 files, 38 tests
- full web Vitest suite: 230 files, 1,289 tests
- `corepack pnpm web:check`
- web type-level suite: 11 files, 13 tests, no type errors
- `corepack pnpm --filter @jobctrl/web e2e:demo-workspace`: 6/6 Chromium tests
- `corepack pnpm web:build`
- `corepack pnpm docs:build`
- `corepack pnpm docs:check:runtime`
- `python3 scripts/release_check.py`
- tracked and untracked whitespace checks
- independent review gate: PASS (0 Blocker, 0 High, 1 Medium)
- independent QA gate: PASS (0 findings)

## Known follow-up

A stale failed scheduler reconciliation can still clear timers armed by a newer generation because its rejection handler is not generation-fenced. This is non-blocking for the P1 scaffold and is the first required acceptance item for P3, where deterministic workflow scenarios begin using the scheduler.

## Stack

- Base plan: #407
- P0 contract and canonical seed: #408
- This PR: P1 browser workspace
- Next: P2 complete read adapter

PR #408, opened 2026-07-11
feat(demo): add canonical synthetic seed

ebarti:
## Summary

- defines typed demo seed, scenario, capability, and mutation-receipt contracts
- adds a deterministic server-shaped synthetic read graph and bundled preview/material assets
- hardens the release scanner to inspect demo fixtures, build output, archives, and exact pinned PDFs
- adds lifecycle, graph-integrity, privacy, and API-semantic parity regression tests

## Why

This is the first implementation layer of the live-demo plan in #407. It establishes one canonical synthetic world before browser persistence or runtime wiring is introduced.

## Stack

- Parent plan: #407
- This PR: P0 — contracts and canonical synthetic seed
- Next: P1 — browser workspace, composition root, and cross-tab event semantics
- Later: P2 read adapter, P3 mutations/workflows, P4 consent telemetry, P5 guided UX, P6 deployment and launch

## Validation

- `corepack pnpm --filter @jobctrl/web check`
- `corepack pnpm --filter @jobctrl/web build`
- focused web demo tests: 10 passed
- `corepack pnpm --filter @jobctrl/api check`
- API seed parity: 1 passed
- release-policy tests: 41 passed
- `python3 scripts/release_check.py`
- `git diff --check docs/live-demo-plan...HEAD`
- implementation review gate: PASS
- QA gate: PASS

## Residual risk

Full route-by-route API response parity is intentionally deferred to P2, where the demo read adapter exists. This layer proves shared semantic and lifecycle parity against the canonical read model.

## Documentation

No user-facing behavior is reachable yet. Canonical user, architecture, QA, and deployment documentation is included in the later stack layers where the demo becomes executable.

PR #407, opened 2026-07-11
docs: plan public live demo

ebarti:
## Summary

- records the accepted architecture for a public JobCtrl demo at `demo.jobctrl.dev`
- keeps all mutable product state in an isolated browser-profile workspace seeded with versioned synthetic data
- exercises the real frontend through complete demo adapters, deterministic workflow scenarios, domain events, and explicit no-effect receipts
- defines the owner-requested acceptance-required cookie gate, non-linkable operational counters, and consented first-party telemetry
- records post-accept withdrawal/current-visitor erasure as an explicit deferred follow-up, not a shipped claim
- documents the verified Cloudflare access, remaining owner inputs, security/privacy stop gates, acceptance matrix, and rollback requirements
- splits delivery into a bottom-up P0-P6 stacked PR sequence

## Key decisions

- The local JobCtrl API, SQLite database, Temporal worker, Python worker, LLMs, Gmail, ATSs, and job boards are never exposed or contacted.
- Isolation is scoped to browser storage profiles/private contexts; shared-profile behavior and local retention are disclosed.
- Demo workspace initialization waits for a server-confirmed cookie grant. Decline records the non-linkable choice when available, creates no analytics ID, redirects to `https://jobctrl.dev`, and a later demo visit opens the banner again.
- The banner states plainly that demo access requires cookie acceptance.
- Post-accept withdrawal, a Manage privacy control, current-visitor deletion, and retryable erasure are deferred to a later reviewed delivery. This stack exposes no non-functional control or claim; exact cookie and raw-event expiry remain enforced and disclosed.
- Consent and granted-initialization health use aggregate-only counters with retry-safe, short-lived non-joinable dedupe digests.
- Rich route, action, error, timing, and funnel events are accepted only after consent and cannot include product state, dynamic URLs, free text, or entity IDs.
- Static assets deploy on a separate `jobctrl-demo` Pages project. A dedicated Worker handles only same-origin `demo.jobctrl.dev/api/*` because Pages Functions cannot host the required Rate Limiting bindings under verified Wrangler 4.107.0; retention stays in a separate Worker.
- Wrangler 4.107.0 is pinned exactly in P4; upgrades must rerun configuration and deployment tests.
- Generic Web Analytics/RUM, generic request traces, and browser CSP reporting endpoints are excluded because they can retain or transmit dynamic request metadata.

## Cloudflare and owner inputs

[...]

PR #410, opened 2026-07-11
feat(demo): add complete read adapter

ebarti:
## Summary

- replaces the demo API placeholder with a complete `ApiClientPort` adapter
- serves every read surface from fresh cloned browser-workspace authority, with no local API, SSE, OS, or external fallback
- matches production job, activity, artifact, workflow-run, contact, and research filtering, sorting, pagination, and stable 404 behavior
- keeps every P3 mutation/rehearsal explicit through a stable capability error
- resolves artifact and profile previews synchronously from the repository-owned authoritative mirror and accepts only mapped `/demo/*` assets
- generation-fences stale scheduler failures so they cannot clear newer timers
- enables the seeded activity-detail direct-load path only in demo mode
- covers all populated routes and seeded deep links through direct load and refresh

## Why

P1 established an isolated browser authority. This slice makes every read-only JobCtrl view useful against that authority while preserving the real frontend query, routing, and invalidation architecture. It intentionally stops before mutations and simulated external effects, which belong to P3.

## Validation

- focused adapter/composition/repository suite: 3 files, 43 tests
- full web Vitest suite: 231 files, 1,302 tests
- `corepack pnpm web:check`
- web type-level suite: 11 files, 13 tests, no type errors
- production web build
- native demo Playwright: 7/7 tests
  - 19 product routes and seeded deep links populate and survive refresh
  - includes `/pipelines`, `/debug`, and `/activity/event-demo-score`
  - zero `/v1`, SSE, external, popup, or OS requests
- tracked/untracked diff and formatting checks
- independent review gate: PASS (0 findings)
- independent QA gate: PASS (0 findings)

## Deliberate boundary

`openArtifact` and every other mutation/rehearsed capability still reject through the stable P3 capability error. P3 will add honest browser-local state changes, deterministic scenarios, and explicit no-effect receipts rather than synthesizing success.

## Stack

- Base plan: #407
- P0 contract and canonical seed: #408
- P1 browser workspace: #409
- This PR: P2 complete read adapter
- Next: P3 mutations and deterministic scenarios

### DIFF
--- a/apps/web/e2e/demo-workspace-tests/demo-workspace.spec.ts
+++ b/apps/web/e2e/demo-workspace-tests/demo-workspace.spec.ts
@@ -73,6 +73,88 @@ test("demo shell renders the shared-profile and personal-data boundary without p
   expect(productRequests).toEqual([]);
 });
 
+test("every P2 product route and seeded deep link renders populated across direct refreshes", async ({
+  page,
+  context,
+}) => {
+  const productRequests: string[] = [];
+  const externalRequests: string[] = [];
+  const runtimeErrors: string[] = [];
+  context.on("request", (request) => {
+    const url = new URL(request.url());
+    if (url.pathname.startsWith("/v1/")) {
+      productRequests.push(url.pathname);
+    }
+    if (url.origin !== "http://127.0.0.1:5198") {
+      externalRequests.push(request.url());
+    }
+  });
+  page.on("pageerror", (error) => runtimeErrors.push(error.message));
+
+  const routes = [
+    ["/dashboard", "Dashboard", "3 jobs"],
+    ["/jobs", "Jobs", "Platform systems lead"],
+    ["/jobs/job-northwind-platform", "Jobs", "Preparation diagnostics"],
+    ["/evidence-map", "Career evidence map", "Delivery improvement evidence"],
+    ["/artifacts", "Artifacts", "Platform systems lead"],
+    ["/artifacts/artifact-tailored-resume", "Artifacts", "Artifact details"],
+    [
+      "/apply-review?jobKey=job-northwind-platform",
+      "Application review",
+      "Materials ready",
+    ],
+    ["/runs", "Workflow runs", "Platform systems lead"],
+    ["/runs/run-materials-progress", "Workflow runs", "Run details"],
+    ["/analytics", "Outcome analytics", "bundled-capture"],
+    ["/profile", "Profile", "Baseline resume editor"],
+    ["/settings", "Settings", "Config"],
+    ["/settings/credentials", "Settings", "OpenAI API Key"],
+    ["/outreach", "Contacts", "Synthetic hiring partner"],
+    [
+      "/outreach/contact-demo-hiring-partner",
+      "Contacts",
+      "Facts and provenance",
+    ],
+    ["/discovery", "Discovery", "Bundled synthetic source"],
+    ["/pipelines", "Pipelines", "Configuring"],
+    ["/debug", "Debug", "Synthetic score recorded."],
+    [
+      "/activity/event-demo-score",
+      "Activity details",
+      "Event details",
+      "dialog",
+    ],
+  ] as const;
+
+  for (const [route, identity, populatedText, role = "heading"] of routes) {
+    await page.goto(STATIC_HOST);
+    await page.goto(route);
+    const identityLocator =
+      role === "dialog"
+        ? page.getByRole("dialog", { name: identity })
+        : page.getByRole("heading", { name: identity }).first();
+    await expect(identityLocator).toBeVisible();
+    await expect(
+      page.getByText(populatedText, { exact: false }).first(),
+    ).toBeVisible();
+    expect(page.url()).toContain(route);
+
+    await page.reload();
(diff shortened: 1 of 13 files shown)


apps/web/src/demo/DemoApiClientAdapter.ts
```diff
@@ -0,0 +1,1008 @@
...
+    if (
+      response.thread === null ||
+      response.thread.contactId !== contactId ||
+      (query.jobId !== undefined && response.thread.jobId !== query.jobId)
+    ) {
+      throw new DemoResourceNotFoundError(
+        "outreach_thread_not_found",
+        contactId,
+      );
+    }
```
ebarti, line 457: PR Reviewer - pr410-review

Severity: Medium
Gate: Pass
Evidence: The real endpoint GET /v1/contacts/:contactId/outreach (apps/api/src/server.ts:2428-2441) returns `{ ok: true, thread: getOutreachThreadForContact(...) }`, and getOutreachThreadForContact (apps/api/src/outreach.ts:107-116) returns null when no thread exists — it never 404s and never checks contact existence. OutreachThreadPanel.tsx:228-239 relies on exactly that: `thread = query.data?.thread ?? null` renders a 'No outreach drafts yet' empty state when thread is null, and only shows an error banner when the query rejects. This method instead throws DemoResourceNotFoundError (404) for every contact except the single seeded thread's contact, so the panel would surface an error banner rather than the empty state. resumeReviewFeedback (line 202) has the same divergence: the real handler (server.ts:995-1001 -> resume-review-drafts.ts:705-715) always returns `{ ok: true, jobKey, feedbackSignals: [] }` with no 404, but the demo 404s for unseeded jobKeys (currently no UI consumer).
Reproduction: Reasoned from code. Not reachable in P2 today because the seed has exactly one contact (read-model.ts:1422-1436) that owns the only thread, but it becomes user-visible once a threadless contact exists (P3 createContact / research confirmation, or any multi-contact seed).

The adapter returns a 404 where production returns a successful null/empty payload for these reads, which contradicts this PR's stated 'matches production ... stable 404 behavior' goal and the documented OutreachThreadResponse.thread: null contract. DemoApiClientAdapter.test.ts also asserts the 404 as correct, so the wrong semantics can propagate into P3.

Required fix: In outreachThread, return `{ ok: true, thread: null }` when the seeded thread does not match contactId/jobId instead of throwing (a genuinely missing thread is a successful null response, not a 404). Apply the same treatment to resumeReviewFeedback (return an empty feedbackSignals response for any jobKey), and update the corresponding test assertions.

apps/web/e2e/demo-workspace-tests/demo-workspace.spec.ts
```diff
@@ -73,6 +73,88 @@ test("demo shell renders the shared-profile and personal-data boundary without p
...
+}) => {
+  const productRequests: string[] = [];
+  const externalRequests: string[] = [];
+  const runtimeErrors: string[] = [];
+  context.on("request", (request) => {
+    const url = new URL(request.url());
+    if (url.pathname.startsWith("/v1/")) {
+      productRequests.push(url.pathname);
+    }
+    if (url.origin !== "http://127.0.0.1:5198") {
```
ebarti, line 88: PR Reviewer - pr410-review

Severity: Low
Gate: Pass
Evidence: This new tripwire hardcodes http://127.0.0.1:5198, but the demo Playwright config derives its port from an env override — demo-workspace.playwright.config.ts:8 reads process.env.JOBCTRL_DEMO_E2E_WEB_PORT (default 5198) and sets baseURL to http://127.0.0.1:${port} (line 19) and the vite --port ${port} web server (line 31). If that env var is set (the knob exists to dodge port collisions across concurrent worktrees), the web server and every same-origin request move to the new port, so `url.origin !== 'http://127.0.0.1:5198'` becomes true for all app requests and `expect(externalRequests).toEqual([])` fails spuriously.
Reproduction: Reasoned from code. Passes at the default port (5198), which matches the hardcoded literal.

The external-request assertion silently assumes the default port and breaks under the config's own supported port override.

Required fix: Derive the expected origin from the same port the config uses (e.g. build it from process.env.JOBCTRL_DEMO_E2E_WEB_PORT with the 5198 fallback) or compare against the page's own origin, instead of the hardcoded literal.

ebarti (commented):
Reviewed the incremental P2 read-adapter diff against its P1 base (feat/demo-p1-browser-workspace) in a clean checkout of the head SHA. The adapter is a faithful port of the real read model: job/activity/artifact/workflow-run filtering, sorting, pagination, compensation sort ordering, tie-break direction handling, the deleted/hidden/closed filters, and contact/research filtering+sorting all match apps/api/src/read-model.ts and apps/api/src/contacts.ts, and the scheduler generation-fencing fix is correct. Two findings, both below the merge gate. Medium=1: outreachThread (and resumeReviewFeedback) throw a 404 where production returns a successful null/empty payload; not reachable with the current single-contact seed but a latent error-handling divergence. Low=1: the new e2e external-request tripwire hardcodes 127.0.0.1:5198 while the config port is env-overridable. Counts: Blocker=0, High=0, Medium=1, Low=1.

PR #411, opened 2026-07-11
feat(demo): add browser-local mutations

ebarti:
## Summary

- implements all 49 synchronous browser-local demo commands behind the production `ApiClientPort`
- keeps list, detail, dashboard, digest, analytics, apply, workflow, contact, outreach, material, and evidence projections coherent in one browser transaction
- adds production-shaped audit events and exact cache invalidation, with revision-only broad recovery for commands whose production write path emits no domain event
- adds truthful replay, quota rollback, reset fencing, permanent-purge cascades, and conservative blob cleanup
- adds native cross-tab and reload coverage for score correction, discovery promotion/import, and eventless settings writes

## Why

The public demo must let visitors exercise safe synchronous product actions without calling the product API, external providers, the host OS, or credential stores. Successful actions must remain durable, cross-tab visible, and auditable instead of returning success-only mock responses.

## Stack

- base: #410 (`feat/demo-p2-read-adapter`)
- this layer: synchronous browser-local mutations
- next layer: deterministic simulated-async scenarios and external no-effect rehearsals

## Validation

- `corepack pnpm check`
- `corepack pnpm test` components completed: scripts 95, API 518, extension 29, extension e2e 4, Python 2,335, launcher race tests
- `corepack pnpm --filter @jobctrl/web test`: 233 files, 1,353 tests
- `corepack pnpm --filter @jobctrl/web test-d`: 11 files, 13 tests, no type errors
- `corepack pnpm --filter @jobctrl/web e2e:demo-workspace`: 10/10 Chromium journeys
- focused demo/event/counter suites: 40/40 plus event-router and domain-event parity coverage
- `corepack pnpm docs:build`: 5,842 references across 262 emitted files
- `git diff --check`
- independent code review: PASS, 0 findings
- independent QA: PASS, 0 findings

Known non-gating output is limited to existing Node localStorage/deprecation warnings, Python dependency deprecations, and the existing Vite large-chunk warning.


apps/web/src/demo/purgeDemoJobProjections.ts
```diff
@@ -0,0 +1,947 @@
...
+    );
+    summary.preparation.workItems = {
+      queued: preparationRuns.filter((run) => run.status === "starting").length,
+      running: preparationRuns.filter((run) => run.status === "in_progress")
+        .length,
+      failed: preparationRuns.filter((run) => run.status === "failed").length,
+    };
+  }
+
+  summary.funnel = summary.funnel.map((stage) => {
```
ebarti, line 775: Medium - dashboard funnel collapses after the first mutation.

PR Reviewer - pr411-review

Severity: Medium
Gate: Pass
Evidence: recomputeDashboard() derives summary.funnel from each active job's jobs.details[...].stages (this block), which matches the real API (apps/api/src/projections.ts:3840-3849). But each seed job carries exactly one stage record (read-model.ts: job and blockedJob -> tailor only, staleJob -> score only) while the seed's authored funnel is inflated (read-model.ts:367-372 -> discover 3, score 3, tailor 3, apply 2). materializeDemoReadModel (clock.ts:88) only rewrites timestamps and never recomputes, so initial load shows the authored funnel. The first recompute-triggering mutation (updateSettings, correctScore, cancel/skip, decideApplyReview, outcomes, deletes, ...) recomputes to discover 0 / score 1 / tailor 2 / apply 0 and dashboardKeys.summary is invalidated, so the dashboard funnel visibly collapses from an unrelated action.
Reproduction: Reasoned from code; DemoLocalCommandExecutor.test.ts (cancel and skip cases) already asserts the recomputed tailor funnel = 1, confirming the derived values differ from the seed's authored 3. No test asserts the discover/apply funnel after a recompute.
Required fix: Reconcile the seed with the recompute - author each seed job with its full stage history so the authored funnel equals recomputeDashboard's output, or derive the seed funnel from the stages at authoring time. Add a regression assertion that an unrelated mutation (e.g. updateSettings) leaves the dashboard funnel unchanged.

apps/web/src/demo/DemoLocalCommandExecutor.ts
```diff
@@ -0,0 +1,2225 @@
...
+      }
+      case "deleteContact": {
+        const contactId = requiredString(args[0], "contactId");
+        requireMapValue(draft.state.readModel.contacts.details, contactId);
+        delete mutableMap(draft.state.readModel.contacts.details)[contactId];
+        draft.state.readModel.contacts.list.items = draft.state.readModel.contacts.list.items.filter((contact) => contact.contactId !== contactId);
+        const thread = draft.state.readModel.outreach.thread.thread;
+        if (thread?.contactId === contactId) {
+          draft.state.readModel.outreach.thread.thread = null;
+          draft.state.readModel.outreach.dueFollowUps.followUps = [];
```
ebarti, line 1446: Low - deleteContact leaves a stale digest follow-up count.

PR Reviewer - pr411-review

Severity: Low
Gate: Pass
Evidence: deleteContact clears outreach.dueFollowUps.followUps here but never recomputes dashboard.digest.followUpsDue.count (no recomputeDemoOperationalProjections / recomputeDigest call in this case), unlike the schedule/complete/dismiss follow-up cases which keep that count in sync (lines 1600-1601). contactDeletedHandler invalidates digestKeys.all, so the dashboard refetches digest() and renders the stale count. From the seed, deleting contact-demo-hiring-partner (owner of the only scheduled follow-up) leaves digest.followUpsDue.count = 1 while the outreach view and dueOutreachFollowUps() show none.
Reproduction: Reasoned from code; the permanent-delete + deleteContact cascade test (DemoLocalCommandExecutor.test.ts:964-982) asserts dueFollowUps === [] but does not assert digest.followUpsDue.count.
Required fix: Call recomputeDigest (via recomputeDemoOperationalProjections) at the end of the deleteContact case when the thread / follow-ups were cleared, and extend the cascade test to assert digest.followUpsDue.count === 0.

ebarti (commented):
Automated review of PR #411 (feat/demo-p3-scenarios) against its P3 phase contract (synchronous browser-local mutations). Scope: the incremental diff vs feat/demo-p2-read-adapter.

The mutation engine is solid: the read-modify-write runs inside the IndexedDB transaction with reset-epoch fencing; validation and persistence failures propagate so the existing mutation hooks roll back their optimistic updates (proven by the quota-rollback and approval-stale tests); the 49 commands emit exact production-shaped events with per-command coverage; eventless commits recover via a revision-bump connection cycle that triggers the reconnect full-invalidation backstop; and safety holds (unsupported methods reject, no fetch / product-API / external requests, asserted in both unit and e2e). The 5 new domain-event types all have invalidation handlers (the router stays exhaustive) and TS/Python DOMAIN_EVENT_TYPES parity is maintained.

2 non-blocking findings (1 Medium, 1 Low), both demo-only projection-coherence issues; details inline. Gate: PASS (no Blocker/High).

PR #412, opened 2026-07-11
feat(demo): add MVP scenario rehearsals

ebarti:
## Summary

- add durable browser-local queued, running, terminal, retry, and recovery behavior for the four public MVP async actions: rescore, re-tailor, retry stage, and run current stage
- add no-effect rehearsals and durable receipts for same-origin artifact preview, application dry-run, and simulated mark-applied
- expose the controls through the existing product UI while keeping all other provider-backed or bulk async operations explicitly unavailable for this MVP
- preserve accepted tailoring artifacts across a failed synthetic attempt and replace them only after a successful retry
- make job-action cancellation select and remove exactly the scheduled run it cancels

## Why

This stack slice makes the public demo feel interactive without calling the product API, model providers, job boards, messaging providers, browser-extension surfaces, or host operating-system openers. The scope is intentionally limited to seven convincing, fully tested interactions.

## Validation

- `corepack pnpm check`
- `corepack pnpm --filter @jobctrl/web test` — 238 files, 1,384 tests
- `corepack pnpm --filter @jobctrl/web test-d` — 13 tests, no type errors
- `corepack pnpm --filter @jobctrl/web e2e:demo-workspace` — 14 Chromium journeys
- `corepack pnpm web:build`
- `corepack pnpm docs:build`
- `python3 scripts/release_check.py`
- `git diff --check`

## Stack

Base: #411 (`feat/demo-p3-scenarios`).

PR #413, opened 2026-07-11
feat(demo): add consent telemetry edge

ebarti:
## Summary

- add same-origin Cloudflare Workers for demo consent, initialization health, typed product telemetry, and hourly retention
- store non-linkable consent/init counters separately from consented pseudonymous events in D1
- issue versioned first-party consent plus HttpOnly visitor/session cookies only after confirmed grant; decline creates no analytics identity
- enforce closed event/attribute schemas, same-origin metadata, bounded bodies, D1 and edge rate limits, 24-hour dedupe/rate expiry, and 90-day data expiry
- include aggregate-only reports while keeping operational and consented populations separate
- keep generated Wrangler runtime bindings out of source control through a minimal four-binding environment contract

Post-accept withdrawal and current-visitor erasure remain explicitly deferred.

## Validation

- `corepack pnpm demo-edge:check`
- `corepack pnpm demo-edge:test` — 26 Worker behavior tests and 4 config/privacy tests
- `corepack pnpm demo-edge:migrate:local` — both migrations applied
- `corepack pnpm demo-edge:dry-run` — API and retention Worker bundles pass
- `corepack pnpm check`
- `corepack pnpm docs:build`
- `git diff --check`

## Stack

Base: #412 (`feat/demo-p3b-scenarios`).

The checked-in zero D1 UUID is intentionally a provisioning placeholder and must be replaced before deployment.

### DIFF
--- a/apps/demo-edge/src/database.ts
+++ b/apps/demo-edge/src/database.ts
@@ -1,0 +1,305 @@
+import {
+  CONSENT_CONTRACT_VERSION,
+  OPERATION_DIGEST_MAX_AGE_SECONDS,
+  OPERATIONAL_RATE_LIMIT_PER_MINUTE,
+  PERSISTENT_COOKIE_MAX_AGE_SECONDS,
+  PRODUCT_DATA_MAX_AGE_SECONDS,
+  RETENTION_SAFETY_MARGIN_SECONDS,
+  TELEMETRY_GLOBAL_RATE_LIMIT_PER_MINUTE,
+  TELEMETRY_RATE_LIMIT_PER_MINUTE,
+  type ConsentChoice,
+  type HealthResult,
+  type StorageMode,
+  type TelemetryEvent,
+} from "./contracts.js";
+import { randomId, sha256 } from "./crypto.js";
+
+interface OperationalCounter {
+  metric: "consent_choice" | "initialization_result";
+  consentChoice: ConsentChoice;
+  initializationResult: "not_applicable" | HealthResult;
+  storageMode: "not_applicable" | StorageMode;
+}
+
+interface CountRow {
+  count: number;
+}
+
+const ACTIVE_IDENTITY_EXISTS_SQL = `SELECT 1 FROM active_demo_identities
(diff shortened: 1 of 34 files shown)


apps/demo-edge/src/database.ts
```diff
@@ -0,0 +1,305 @@
...
+
+function utcDay(now: Date): string {
+  return now.toISOString().slice(0, 10);
+}
+
+async function isWithinOperationalRateLimit(env: DemoEdgeEnv, endpoint: "consent" | "health", now: Date): Promise<boolean> {
+  const row = await env.DEMO_TELEMETRY_DB.prepare(
+    `INSERT INTO operational_rate_windows (endpoint, window_utc, count, expires_at)
+     VALUES (?, ?, 1, ?)
+     ON CONFLICT(endpoint, window_utc) DO UPDATE SET
```
ebarti, line 67: [Medium] `operational_rate_windows` is keyed only on `(endpoint, window_utc)`, so `OPERATIONAL_RATE_LIMIT_PER_MINUTE` (120) is a single GLOBAL budget shared by every client's consent operations — both grants and denials call `isOperationalRequestAllowed(env, "consent", now)`. In `handlers.ts` the grant path hard-returns 429 once this global budget is exhausted (before creating the identity), so grant confirmation is coupled to a shared aggregate-counter budget. The per-IP `PUBLIC_INGRESS_LIMITER` allows 240/min, which alone exceeds 120, so a single client — or an ordinary launch traffic spike above 120 consent choices/min — can drive all visitors' grants to 429 and block entry to the demo. (Ingress-limited declines also still consume this budget, since the denied path calls `isOperationalRequestAllowed` without gating on `context.ingressAllowed`.) Denials fail open (always 200) and telemetry silently drops on its global limit, so only essential grant confirmation degrades hard. Consider keying the operational limit per-IP, decoupling grant confirmation from the aggregate-counter budget, or sizing 120/min for expected public launch traffic.

ebarti (commented):
One Medium availability/abuse finding on the consent grant path; core privacy/consent-bypass/injection invariants look solid.

PR #427, opened 2026-07-12
fix(release): accept standard Ed25519 signing keys

ebarti:
## What changed

- accept canonical Ed25519 PKCS#8 DER private keys based on cryptographic parsing and key type instead of an invalid minimum-length heuristic
- add regression coverage for the standard 48-byte encoding on both the source module and the sealed finalizer used by the signing job
- regenerate both audited release-authority bundles and their checksums
- document fail-safe one-time key generation, private-secret encoding, and matching raw public-key derivation

## Root cause

Standard OpenSSL and Node Ed25519 PKCS#8 DER exports are 48 bytes. The release parser required the decoded key to be greater than 48 bytes before attempting PKCS#8 parsing, so a valid key generated by the documented toolchain could never reach the cryptographic validation. The owner checklist also named the private and public values without explaining how to generate a matching pair.

## Validation

- `node --test scripts/distribution-release.test.mjs` — 22 passed
- operator key-generation QA — 48-byte PKCS#8 accepted, matching 32-byte public key derived, second generation refused without overwrite
- `corepack pnpm scripts:test` — 102 passed
- `corepack pnpm distribution:audit` — passed
- `python3 scripts/release_check.py --strict-prompt` — passed
- `corepack pnpm docs:build` — passed
- `corepack pnpm docs:check:runtime` — passed
- `corepack pnpm check` — passed

No Apple, App Store Connect, or JobCtrl release credentials are included in this changeset.

### DIFF
--- a/docs/publish-checklist.md
+++ b/docs/publish-checklist.md
@@ -161,6 +161,60 @@ branches. Dispatch the workflow at `refs/tags/<release_tag>` so its own
 the resolver. These environment rules are the server-side defense against a
 modified branch workflow removing the in-workflow identity check:
 
+Generate the JobCtrl Ed25519 release key pair once, outside the repository. This
+key signs JobCtrl manifests and release descriptors; it is separate from the
+Apple Developer ID certificate used to sign macOS executables:
+
+```bash
+release_key_dir="$HOME/.jobctrl-release-secrets"
+(
+  set -euo pipefail
+  umask 077
+  mkdir -p "$release_key_dir"
+  chmod 700 "$release_key_dir"
+  private_pem="$release_key_dir/jobctrl-release-v1.pem"
+  private_der="$release_key_dir/jobctrl-release-v1.pk8"
+  if [[ -e "$private_pem" || -e "$private_der" ]]; then
+    echo "release key already exists; refusing to overwrite it" >&2
+    exit 1
+  fi
+  openssl genpkey -algorithm ED25519 -out "$private_pem"
+  openssl pkey -in "$private_pem" -outform DER -out "$private_der"
+)
+```
+
+Back up both private-key files offline. Never commit them. Copy the canonical
+base64 PKCS#8 DER value for the `JOBCTRL_RELEASE_SIGNING_KEY` environment secret,
+then paste it into `release-signing` before running the next clipboard command:
+
+```bash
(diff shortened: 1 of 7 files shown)


docs/publish-checklist.md
```diff
@@ -161,6 +161,60 @@ branches. Dispatch the workflow at `refs/tags/<release_tag>` so its own
...
+  umask 077
+  mkdir -p "$release_key_dir"
+  chmod 700 "$release_key_dir"
+  private_pem="$release_key_dir/jobctrl-release-v1.pem"
+  private_der="$release_key_dir/jobctrl-release-v1.pk8"
+  if [[ -e "$private_pem" || -e "$private_der" ]]; then
+    echo "release key already exists; refusing to overwrite it" >&2
+    exit 1
+  fi
+  openssl genpkey -algorithm ED25519 -out "$private_pem"
```
ebarti: PR Reviewer - pr427_review

Severity: Medium
Gate: Pass
Evidence: `docs/publish-checklist.md:181-182` invokes unqualified `openssl genpkey -algorithm ED25519`; on the supported macOS host, `/usr/bin/openssl version` reports LibreSSL 3.3.6 and `/usr/bin/openssl genpkey -algorithm ED25519 ...` exits nonzero with `Algorithm ED25519 not found`.
Reproduction: Reproduced with command/steps

The documented one-time key setup is not executable on a stock macOS toolchain. This machine only passed the recipe because PATH selected Homebrew OpenSSL 3.6.2; explicitly running Apple's bundled `/usr/bin/openssl` fails before creating a key. Since Node is already a required JobCtrl tool and supports Ed25519 generation/export, the owner checklist should not silently depend on an undocumented Homebrew OpenSSL installation.

Required fix: Generate/export the PKCS#8 and matching public key with the required Node runtime, or explicitly document and fail-check an OpenSSL 3 prerequisite/path before the generation command; add a QA check that exercises the documented procedure without relying on PATH selecting Homebrew OpenSSL.

ebarti: Fixed in 879a6ca8. The setup now generates Ed25519 PKCS#8 DER with the required Node runtime and an atomic no-overwrite write, removing the undeclared OpenSSL 3 dependency. The stock-macOS path was re-exercised: 48-byte private DER, matching 32-byte public key, modes 0700/0600, and overwrite refusal all pass.

PR #436, opened 2026-07-13
build(release): publish artifacts directly to R2

ebarti:
## What changed

- replace the generic release upload URL with bucket-scoped Cloudflare R2 S3 credentials
- publish immutable artifacts with conditional `PutObject` and promote channel pointers with compare-and-swap semantics
- keep checksum verification on the public `releases.jobctrl.dev` read path
- extend the release policy scanner and regressions to require native conditional R2 publication
- document the protected environment secret and variable contract
- add an intentionally unchecked root launch checklist for the remaining hosted gates

## Why

The release workflow previously assumed a separate HTTP upload service implementing authenticated conditional writes. R2 already exposes those semantics through its S3-compatible API, so the workflow can upload directly with the AWS CLI while preserving immutable paths, channel-pointer conflict detection, credential isolation, and public readback verification.

## Validation

- `node --test scripts/distribution-release.test.mjs` — 22 passed
- `uv --project workers/automation run --extra dev pytest -q workers/automation/tests/test_release_check.py` — 42 passed
- `corepack pnpm scripts:test` — 107 passed
- `python3 scripts/release_check.py` — passed
- `uv --project workers/automation run --extra dev ruff check scripts/release_check.py workers/automation/tests/test_release_check.py` — passed
- `corepack pnpm docs:build` — passed; 5,837 emitted references resolved
- `go run github.com/rhysd/actionlint/cmd/actionlint@v1.7.7 .github/workflows/release-distribution.yml` — passed

## Hosted setup

The `jobctrl-releases` R2 bucket and `releases.jobctrl.dev` custom domain are provisioned. The `release-publication` environment has the R2 access keys, administration-read token, Homebrew deploy key, account ID, and bucket configuration.

GitHub currently rejects required-reviewer and `v*`-tag environment protection for this private repository under its billing/plan state. Actions is also preventing jobs from starting for the same billing/spending reason, so no workflow steps execute until that external blocker is cleared and the documented protection is enabled.

### DIFF
--- a/scripts/release_check.py
+++ b/scripts/release_check.py
@@ -10,6 +10,7 @@ import argparse
 import hashlib
 import json
 import re
+import shlex
 import subprocess
 import sys
 import tarfile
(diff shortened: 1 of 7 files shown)


scripts/release_check.py
```diff
@@ -1137,12 +1143,43 @@ def _release_distribution_findings(root: Path) -> list[str]:
...
+            "R2_BUCKET: ${{ vars.JOBCTRL_R2_BUCKET }}",
+            "aws s3api put-object",
+        ):
+            if marker not in body:
+                findings.append(
+                    f"{RELEASE_DISTRIBUTION_WORKFLOW_PATH}: {job_name} must publish with protected bucket-scoped R2 S3 authority"
+                )
+                break
+    publish_immutable = jobs.get("publish-immutable")
+    if publish_immutable is not None and "--if-none-match '*'" not in publish_immutable:
```
ebarti: PR Reviewer - Codex R2 Reviewer

Severity: Medium
Gate: Pass
Evidence: `scripts/release_check.py:1171-1182` searches each raw job body for literal flag text. In a temporary-workflow mutation I removed all three conditional arguments from the actual `aws s3api put-object` calls, left `--if-match "$etag"` / `--if-none-match '*'` only in YAML comments, and `_release_distribution_findings()` returned `conditional_findings=[]`.
Reproduction: Reproduced with a temporary mutated workflow passed to `_release_distribution_findings()`; no repository files were changed.

The new release-policy gate can be satisfied by comments or other non-executable text, so a future edit that makes immutable uploads or channel promotion unconditional can still pass both the scanner and the added regression test.

Required fix: Check `_workflow_executable_text(...)` and bind each conditional flag to the corresponding `aws s3api put-object` invocation in the owning job; add a negative test that moves the required tokens into comments while making the actual puts unconditional.

PR #437, opened 2026-07-13
feat(dev): add docs and demo lifecycle commands

ebarti:
## What changed

- adds foreground and detached start/status/stop commands for the docs site
- adds a complete local demo lifecycle covering the demo-mode Vite app and local Wrangler consent/telemetry API
- persists local demo D1 state and applies migrations against the same storage before startup
- proxies `/api` through Vite without changing origin, preserving the production consent boundary
- adds lifecycle regression coverage and updates contributor documentation

## Why

Docs and demo preview servers previously required ad hoc commands and had no tracked status or stop surface. The public demo also could not exercise its real fail-closed consent path locally without a Cloudflare preview.

## Validation

- `corepack pnpm scripts:test` — PASS, 111/111
- `corepack pnpm web:check` — PASS
- `corepack pnpm demo-edge:check` — PASS
- `corepack pnpm docs:build` — PASS; 5,837 references across 261 emitted files
- real `docs:start` → `docs:status` → HTTP 200 → `docs:stop` lifecycle — PASS
- real `demo:start` → `demo:status` → Vite/Worker HTTP checks → `demo:stop` lifecycle — PASS
- browser consent gate → accept → dashboard → reload → stop/restart against the same D1 state → dashboard without a repeated banner — PASS, with no console warnings/errors
- independent exact-head review — Gate: PASS, no findings
- `git diff --check` — PASS

GitHub's Demo Site and Native Launcher jobs currently fail before executing any steps and expose no job logs; this is the repository's hosted execution/billing gate rather than a code-test result. The complete corresponding local checks above passed.


scripts/dev
```diff
@@ -27,16 +27,21 @@ PID_DIR="$DEV_DIR/pids"
...
+known_process_names() { echo "$(process_names) docs demo-api demo-web"; }
 
 process_command() {
   case "$1" in
     temporal) echo 'temporal server start-dev --db-filename "$JOBCTRL_TEMPORAL_DB"' ;;
     api)      echo "pnpm --filter @jobctrl/api dev" ;;
     web)      echo 'pnpm --filter @jobctrl/web exec vite --host 127.0.0.1 --port "$JOBCTRL_WEB_PORT"' ;;
     worker)   echo "uv --project workers/automation run jobctrl worker" ;;
+    docs)     echo 'corepack pnpm exec vitepress dev docs --host 127.0.0.1 --port "$JOBCTRL_DOCS_PORT"' ;;
+    demo-api) echo 'corepack pnpm --filter @jobctrl/demo-edge exec wrangler d1 migrations apply DEMO_TELEMETRY_DB --local --config wrangler.api.jsonc --persist-to "$JOBCTRL_DEMO_STATE_DIR" && exec corepack pnpm --filter @jobctrl/demo-edge exec wrangler dev --config wrangler.api.jsonc --local --ip 127.0.0.1 --port "$JOBCTRL_DEMO_API_PORT" --persist-to "$JOBCTRL_DEMO_STATE_DIR"' ;;
```
ebarti, line 43: PR Reviewer - review_pr437

Severity: High
Gate: Fail
Evidence: `scripts/dev:43` registers `demo-api` as `migrate && exec wrangler dev`, but both launch paths execute every registry string as `bash -c "exec $cmd"` (`scripts/dev:229` and `scripts/dev:388`). The resulting shell program is `exec <migration> && exec <server>`: the first successful `exec` replaces Bash with the finite migration process, so there is no shell left to evaluate the `&&` branch and Wrangler dev is never launched. The new harness reinforces this: its migration stub exits 0, after which the tracked demo-api PID exits instead of reaching the Wrangler stub.

The complete demo lifecycle cannot start its API process. Both `pnpm demo:dev` and `pnpm demo:start` run only the migration for `demo-api`; the foreground supervisor then tears down the demo when that process exits, while detached mode records a soon-stale PID and may still start the web process without consent/API availability.

Required fix: Execute compound registry commands without prepending `exec` to the first command (while preserving an `exec` for the final long-running server), and add/retain a regression assertion that the migration stub ran and the Wrangler-dev stub became the tracked live process in both foreground and detached launch paths.

PR #442, opened 2026-07-14
feat(web): establish redesign foundation

ebarti:
## Summary

- introduces the shared responsive compositions defined by the redesign plan
- refines form controls, tabs, spacing, shell search, and density selection around the existing JobCtrl tokens
- adds component stories and semantic regression coverage for disclosure state, choices, adaptive grids, and document preview regions

## Stack

- Base: `docs/redesign-implementation-plan` (#440)
- Successor: operational lists and dashboards

## Validation

The component tests and stories are authored in this layer. Per the redesign delivery contract, the integrated build, browser, accessibility, visual, and product-flow QA gate will run only after the full implementation stack and final documentation pass are complete.

PR #443, opened 2026-07-14
feat(web): redesign operational surfaces

ebarti:
## Summary

- brings Dashboard, Analytics, Jobs, Pipelines, Discovery, Artifacts, Contacts, Runs, and Debug onto one route-page and data-surface composition
- replaces ad-hoc operational filter controls with the shared select, input, toggle, and tool-row system
- makes dense dashboards and pipeline controls width-adaptive while preserving every existing query, status, action, table column, and URL-backed state
- turns Discovery configuration groups into state-preserving disclosures

## Stack

- Base: `feat/redesign-foundation` (#442)
- Successor: route workspaces and audit surfaces

## Validation

Existing production-shaped view, data-grid, interaction, and accessibility tests remain in place. Per the redesign delivery contract, the complete test, browser, visual, accessibility, and product-flow QA gate is deferred until the full stack and final documentation pass are complete.

### DIFF
--- a/apps/web/src/shared/ui/disclosure-section.tsx
+++ b/apps/web/src/shared/ui/disclosure-section.tsx
@@ -64,7 +64,9 @@ export function DisclosureSection({
             stroke={1.9}
           />
           <span className="disclosure-section__heading">
-            <span className="disclosure-section__title">{title}</span>
+            <span className="disclosure-section__title" role="heading" aria-level={2}>
+              {title}
+            </span>
             {description ? (
               <span className="disclosure-section__description">{description}</span>
             ) : null}
--- a/apps/web/src/contexts/discovery/components/DiscoveryAutomationSettingsPanel.tsx
+++ b/apps/web/src/contexts/discovery/components/DiscoveryAutomationSettingsPanel.tsx
@@ -7,7 +7,7 @@ import { useForm } from "@tanstack/react-form";
 import { useEffect, useRef, useState } from "react";
 
 import { AutosaveUndoController } from "../../../shared/ui/autosave-undo-controller.js";
-import { CardHeader } from "../../../shared/ui/card-header.js";
+import { DisclosureSection } from "../../../shared/ui/disclosure-section.js";
 import { Empty } from "../../../shared/ui/empty.js";
 import { useDiscoverySettingsQuery } from "../../operations/hooks/useDiscoverySettingsQuery.js";
 import { useUpdateDiscoverySettingsMutation } from "../hooks/useUpdateDiscoverySettingsMutation.js";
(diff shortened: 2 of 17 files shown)


apps/web/src/shared/ui/disclosure-section.tsx
```diff
@@ -64,7 +64,9 @@ export function DisclosureSection({
             stroke={1.9}
           />
           <span className="disclosure-section__heading">
-            <span className="disclosure-section__title">{title}</span>
+            <span className="disclosure-section__title" role="heading" aria-level={2}>
```
ebarti, line 67: PR Reviewer - pr-443

Severity: Medium
Gate: Pass
Evidence: This PR moves DiscoveryAutomationSettingsPanel, DiscoveryRuntimeSettingsPanel, and TargetSearchSettingsPanel off CardHeader (which renders a real `<h2>{title}</h2>` outside any control, card-header.tsx:11) onto DisclosureSection, and in the same diff adds `role="heading" aria-level={2}` to this title span. The span is a descendant of the trigger `<button class="disclosure-section__trigger">` (lines 53-77). Per WAI-ARIA, role=button has presentational children, so browsers flatten this span into the button's accessible name and do NOT expose it as a heading -- screen-reader users lose the level-2 heading navigation these three panels had before this PR. DiscoveryView.test.tsx:12-22 asserts getByRole("heading", { name: "Automation settings" }) etc. and still passes only because Testing Library does not prune presentational descendants, so the regression is hidden by a green test instead of caught by it.
Reproduction: Reasoned from code (ARIA presentational-children rule; card-header.tsx vs disclosure-section.tsx; DiscoveryView.test.tsx). Not run in a browser screen reader.

The nested role=heading inside the interactive trigger is ineffective and regresses the real <h2> headings these panels previously exposed to assistive tech.

Required fix: Expose the title as a real heading outside the interactive control, per the ARIA disclosure/accordion pattern -- e.g. wrap the trigger in a heading element (`<h2 class="..."><button aria-expanded ...>...</button></h2>`) rather than putting role=heading on a descendant of the button. Confirm with a check that inspects the accessibility tree or a real AT, since getByRole("heading") passes even when the heading is suppressed.

ebarti (commented):
Well-contained CSS/markup refactor: it preserves every query, URL-backed filter, status, action, and table column I checked, and it does not break the existing unit tests (the analytics dimension test still resolves the Radix ToggleGroupItem as a button; the discovery heading assertions still resolve the new role=heading span). One substantive finding: the shared DisclosureSection heading change is a masked accessibility regression for the three discovery/target config panels. Gate: PASS (0 Blocker/High; 1 Medium).

ebarti:

PR #446, opened 2026-07-14
feat(web): redesign setup and document workspaces

ebarti:
## Stack

- Base: #444 (`feat/redesign-workspaces`)
- This is the setup/forms/document-workbench layer of the redesign stack.
- Successor: pipeline execution-lineage foundation.

## What changed

- recomposes Profile around collapsible personal, baseline, experience, education, skills, and voluntary-EEO sections while keeping the live real resume editor beside the form
- rebuilds Preferences with collapsible application, tailoring, and resume-style sections; content/writing/quality tabs; adaptive field grids; shared selects and choice controls; contextual documentation links; and every original field/control preserved
- turns Resume templates into a full-width real Plate workbench with compact controls above the document preview
- modernizes the three-step resume import flow with route-backed semantic tabs, shared controls, and the real PDF-file affordance
- standardizes General, Credentials, Model selection, AI execution policy, Browser capabilities, and extension pairing around the same disclosure, field, ledger, and action patterns
- removes colored card/badge clutter from the redesigned setup surfaces and keeps status/provenance as plain, inspectable text
- adds focused component, interaction, accessibility, and Storybook coverage for the new compositions

## Why

The setup surfaces mixed rigid two-column forms, native selects and checkboxes, dense provider cards, duplicated headings, and a squeezed resume preview. The redesign keeps every setting and mutation in its owning context while giving the routes one consistent hierarchy, spacing model, and responsive behavior.

## Data and behavior parity

The implementation preserves profile JSON paths, required-entry/bullet/skill controls, date handling, autosave/undo, save/discard/import actions, provider modes and credential ownership rules, model catalog states, AI policy fields, browser capability adoption and revocation, profile-copy consent, extension token rotation, template cloning/default mutations, errors, disabled reasons, and demo/read-only states.

## Validation

- `git diff --check` — passed
- scoped static path/action/mutation audits — passed during implementation
- focused tests, stories, and a11y cases were authored but intentionally not run

Per the delivery instruction, builds, typechecks, component suites, Storybook/a11y execution, and browser/product QA are deferred until the complete redesign, pipeline-operations implementation, final documentation, and screenshots are assembled.

PR #439, opened 2026-07-14
docs(plans): add pipeline operations visibility plan

ebarti:
## Summary

- add the active Discovery Pipeline Operations Visibility plan and index it in the public plan ledger
- define durable execution lineage, authoritative backlog and worker-capacity sources, active-work telemetry, and calibrated ETA semantics
- separate true source-family progress from terminal reconciliation and downstream preparation drain
- keep the operational and accessibility contract stable while allowing P3 to use the app UI redesign that is current when implementation starts

## Why

The current Pipelines status presents the source-family plus reconciliation spine as overall Discover completion. Jobs can already be queued for scoring or later preparation while the page still gives users no explanation of the backlog, shared worker capacity, active work, telemetry freshness, or expected completion time.

This plan treats that as an end-to-end operations visibility gap rather than a wording-only issue.

## Scope

This is a planning and documentation PR only. It does not implement the API, persistence, telemetry, estimator, or frontend changes. Delivery is split into P0-P4 reviewable phases with explicit review, QA, privacy, accessibility, and product-path gates.

The UI section intentionally fixes information priority and semantic behavior without freezing cards, tables, spacing, or responsive composition. P3 must re-anchor on the latest landed redesign and reuse its primitives and tokens.

## Validation

- `corepack pnpm docs:build` — passed; 8,221 references resolved across 316 files
- `python3 scripts/release_check.py` — passed
- `git diff --check` — passed
- independent PR review — Gate: PASS, zero findings
- independent QA — Gate: PASS, zero findings

PR #447, opened 2026-07-14
feat(pipeline): persist discovery execution lineage

ebarti:
## Summary

- add immutable `DiscoveryExecutionRef` identity for each Temporal Discover run
- persist execution-scoped current-run versus existing-backlog job membership
- record explicit preparation work-plan outcomes without treating missing plans as no work
- preserve idempotent backlog promotion and immutable first-link history across retries and workflow ID reuse
- thread lineage through discovery sources, reconciliation, fan-out, and per-job preparation inputs

## Why

This is PR 5 in the product-redesign stack and the lineage foundation for the pipeline-operations visibility adaptation from #439. Operations needs a durable execution key and truthful cohort membership before it can report draining work, backlog contention, or ETA.

## Stack

- base: `feat/redesign-setup` / PR #446
- successor: pipeline lifecycle and runtime telemetry

## Validation

- `git diff --check`
- static execution-authority, source-boundary, propagation, schema-version, and compatibility audit
- focused regression tests were authored for serialization, retry/history, promotion, work-plan outcomes, and terminal reconciliation

Full test, replay/determinism, documentation, and browser QA are intentionally deferred until the assembled redesign stack is complete, per the delivery plan.

### DIFF
--- a/workers/automation/src/jobctrl/pipeline/preparation.py
+++ b/workers/automation/src/jobctrl/pipeline/preparation.py
@@ -19,12 +19,21 @@ from temporalio.common import WorkflowIDConflictPolicy
 
 from jobctrl import database as db_module
 from jobctrl.database import get_connection, get_jobs_by_stage
+from jobctrl.domain.discovery.execution import (
+    DiscoveryExecutionCohortKind,
+    DiscoveryExecutionRef,
+    DiscoveryExecutionWorkPlanState,
+    validate_required_steps,
+)
 from jobctrl.domain.identifiers import JobId
 from jobctrl.domain.materials.use_cases import SuppressTailoredArtifactsUseCase
 from jobctrl.domain.preparation import PreparationWorkItemKind, make_preparation_idempotency_key
 from jobctrl.domain.rpc.messages import WorkflowStartSpec
 from jobctrl.domain.tenant import LOCAL_TENANT, TenantId
 from jobctrl.infrastructure.materials import SqliteMaterialsRepository, SqliteTailoringPolicyRepository
+from jobctrl.infrastructure.discovery.sqlite_execution_repository import (
+    SqliteDiscoveryExecutionRepository,
+)
 from jobctrl.infrastructure.rpc.workflow_starter import (
     WorkflowStarter,
     default_workflow_starter,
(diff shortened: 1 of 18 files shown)


workers/automation/src/jobctrl/pipeline/preparation.py
```diff
@@ -355,6 +449,197 @@ def _workflow_spec_for_target(
...
+    for target in targets:
+        if discovery_cohort_kind == "existing_backlog":
+            membership = repository.link_job(
+                discovery_execution,
+                target.job_url,
+                cohort_kind="existing_backlog",
+            )
+        else:
+            membership = repository.get(discovery_execution, target.job_url)
+            if membership is None:
```
ebarti, line 480: PR Reviewer - pr-447

Severity: High
Gate: Fail
Evidence: The observed branch raises when a derived `pending_score` target has no `observed_this_run` lineage row. Observed rows are created only by source ingestion (`attach_source_observation`). Enrichment selects jobs globally via `_ENRICHMENT_PENDING` (detail.py `run_enrichment`, ~line 2147), so a job discovered in a PRIOR run that is still `pending_enrichment` and is NOT re-observed by any source this run (reachable after a cancel+rerun, or a detail-site error that recovers on the next run) gets enriched this run and crosses into `pending_score` AFTER the pre-run sweep — and the sweep's `_derive_targets` only covers `pending_score`/`pending_tailor`, never `pending_enrichment`. Such a job therefore has neither an observed row nor a backlog row. The terminal `derive_preparation_targets(include_pending_tailor=False)` returns it, `repository.get(...)` is `None`, and this `raise` fires. The streaming per-job handoff swallows the error (detail.py:1464), but the TERMINAL fan-out re-raises (activities.py `_run_fanout`, ~line 356); `start_discovery_preparation_workflows` then runs `_mark_pending_work_plans_failed(reason="work_plan_persistence_failed")` and re-raises.
Reproduction: Reasoned from code (not run). Not covered by tests: the lineage tests monkeypatch `derive_preparation_targets`, so no test derives a real target lacking an observed row.

Two consequences: (1) the authoritative terminal preparation fan-out aborts for the ENTIRE run the moment one un-observed carry-over target appears — no preparation workflows start and `_finalize_unplanned_observed_work_plans` never runs; (2) every still-`pending` observed row for HEALTHY jobs is stamped `work_plan_state="failed"`, so the lineage records a false "failed" work plan for jobs that were fine — the exact auditability/truthfulness invariant this PR exists to uphold.

Required fix: Do not let one missing observed membership abort the batch or bulk-fail unrelated rows. Resolve the outcome per-target (create the observed row on demand from canonical evidence at the owning boundary, or classify only that job), or make the observed branch tolerate a legitimately un-observed `pending_score` target instead of raising. Add a regression test that derives a real `pending_score` target with no observed row and asserts the terminal fan-out still starts other jobs and leaves healthy rows uncorrupted.

ebarti (commented):
Schema/versioning and idempotency look sound (additive CREATE TABLE IF NOT EXISTS, monotonic user_version bump, API fails closed only on newer DBs, upsert-based link_job preserving first-link metadata). One High correctness/auditability concern: the observed-cohort preparation fan-out assumes every derived pending_score target already has an observed lineage row, which does not hold for carry-over jobs enriched (but not re-observed) this run — detail inline.

ebarti:

PR #448, opened 2026-07-14
feat(pipeline): add runtime operations telemetry

ebarti:
## Summary

- add typed queued, started, completed, and failed lifecycle events for orchestration steps outside canonical per-job stage state
- fold step lifecycle identically in Python and TypeScript with attempt-aware retry and shared-watermark recovery
- emit durable activity-owned facts for source planning, source families, reconciliation passes, backlog sweep, fan-out, and lineage-linked PDF rendering
- add privacy-safe worker heartbeats with exact all-activity slot occupancy and bounded allowlisted detail
- aggregate fresh multi-process capacity and one freshest Temporal task-queue observation with explicit stale, unsupported, unavailable, and invalid-capacity states

## Why

This is PR 6 in the product-redesign stack and the runtime-observability layer for the pipeline-operations visibility adaptation from #439. It closes orchestration visibility gaps without creating a second job-stage state machine, while preserving honest shared-capacity and privacy boundaries for the later operations snapshot and ETA.

## Stack

- base: `feat/pipeline-execution-lineage` / PR #447
- successor: pipeline operations read model and ETA

## Validation

- `git diff --check`
- static cross-runtime event/projection parity audit
- static Temporal determinism, activity ownership, retry/terminal ordering, capacity arithmetic, and privacy audit
- focused Python, API, domain, invalidation, and telemetry regression fixtures were authored

Full tests, builds, replay/determinism checks, documentation, and browser QA are intentionally deferred until the assembled redesign stack is complete, per the delivery plan.

PR #449, opened 2026-07-14
feat(pipeline): expose operations snapshot

ebarti:
## Summary

- adapts the execution-visibility semantics from #439 into the redesigned pipeline operations foundation
- adds a typed `/v1/pipeline/operations` read model with exact discovery lineage, current/sweep/global cohort boundaries, source-family and reconciliation progress, shared worker capacity, and privacy-safe active work
- adds a conservative, evidence-backed ETA estimator that reports calibrated bounds only when membership, runtime telemetry, contention, and duration evidence are trustworthy
- exposes the snapshot through the shared contracts, API client, local web adapter, and capability manifest
- preserves retryable work and terminal workflow ownership semantics without fabricating queue state, active identities, or zero-valued telemetry

## Stack

- base: #448 (`feat/pipeline-runtime-observability`)
- successor: redesigned Pipelines workspace consuming this snapshot
- full documentation, screenshots, and end-to-end QA remain intentionally deferred to the final stack layers

## Validation

- authored focused API, estimator, route, adapter, privacy, lineage, contention, and retryability regression coverage
- `git diff --cached --check`
- independent static review gate: PASS (Blocker 0, High 0)
- tests, typecheck, lint, builds, and browser QA not run yet by stack sequencing; they will run once implementation and documentation are complete

### DIFF
--- a/apps/api/src/pipeline-operations.ts
+++ b/apps/api/src/pipeline-operations.ts
@@ -1,0 +1,1921 @@
+import { createHash } from "node:crypto";
+
+import type {
+  DiscoveryExecutionSummary,
+  PipelineActiveItem,
+  PipelineCapacity,
+  PipelineEta,
+  PipelineExecutionCohortSummary,
+  PipelineOperationsFreshness,
+  PipelineOperationsSnapshot,
+  PipelineOperationalStage,
+  PipelineStageCounts,
+} from "./contracts.js";
+import { allRows, tableExists, type SqliteDatabase, type SqliteValue } from "./db.js";
+import {
+  estimatePipelineEta,
+  type PipelineEtaEstimatorInput,
+  type PipelineEtaStageInput,
+} from "./pipeline-eta.js";
+import { refreshProjections } from "./projections.js";
+import { readWorkerRuntimeTelemetry, type WorkerRuntimeTelemetrySnapshot } from "./worker-runtime-telemetry.js";
+import { readLlmSpendHealth } from "./worker-health.js";
+
+const DEFAULT_TENANT = "local";
+const ETA_SAMPLE_LIMIT = 50;
+const ETA_SAMPLE_WINDOW_MS = 14 * 24 * 60 * 60 * 1_000;
+
+const JOB_STAGES = ["enrich", "score", "tailor", "cover"] as const;
+const STEP_STAGES = ["source_planning", "source_family", "reconciliation", "pdf_render"] as const;
+const OPERATIONAL_STAGES = [...STEP_STAGES, ...JOB_STAGES] as const;
+
+const STAGE_LABELS: Record<(typeof OPERATIONAL_STAGES)[number], string> = {
+  source_planning: "Plan sources",
+  source_family: "Crawl sources",
+  reconciliation: "Reconciliation",
+  pdf_render: "Render PDF",
+  enrich: "Enrich",
+  score: "Score",
+  tailor: "Tailor",
+  cover: "Cover letter",
+};
+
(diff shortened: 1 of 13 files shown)


apps/api/src/pipeline-operations.ts
```diff
@@ -0,0 +1,1921 @@
...
+  selected: SelectedExecution,
+): Map<(typeof OPERATIONAL_STAGES)[number], PipelineStageCounts> {
+  const counts = new Map<(typeof OPERATIONAL_STAGES)[number], PipelineStageCounts>();
+  for (const stage of OPERATIONAL_STAGES) counts.set(stage, emptyCounts());
+  if (!tableExists(db, "job_stage_states")) return counts;
+  const excluded = selected.current.members.concat(selected.sweep.members).map((member) => member.job_url);
+  const where = ["jss.stage IN ('enrich', 'score', 'tailor', 'cover')"];
+  const params: SqliteValue[] = [];
+  if (excluded.length > 0) {
+    where.push(`jss.job_url NOT IN (${excluded.map(() => "?").join(", ")})`);
```
ebarti, line 828: PR Reviewer - pr-449

Severity: Medium
Gate: Pass
Evidence: `globalStageCounts` (line 828) and `loadGlobalRetryability` (line 857) each build one `job_url NOT IN (?, ?, ...)` clause containing one bound parameter per current+sweep member. The `existing_backlog` cohort is an uncapped "one-time full-derive fan-out for work left over from a PRIOR run" (workers/automation/src/jobctrl/discovery/workflow.py `_sweep_preexisting_preparation`), so the excluded set is unbounded. better-sqlite3 caps bound variables at SQLITE_MAX_VARIABLE_NUMBER (32766). Every other reader in this file (loadStageStates:351, loadPreparationWorkflows:382, loadUniqueGlobalPreparationOwners:898, loadPreparationWorkflowStatusById:931, loadJobDisplays:1295) chunks its IN-list at 500 — only these two global queries do not.
Reproduction: Reasoned from code (driver variable cap + uncapped sweep). Not run against a >32k-member fixture.

When one execution's current+sweep cohort exceeds ~32.7k job_urls, `db.prepare(sql)` throws "too many SQL variables"; `withDb` has no handler for it, so the entire `GET /v1/pipeline/operations` snapshot returns 500 instead of degrading — on exactly the large-backlog operational state this read model exists to report.

Required fix: Do not pass the full exclusion set as bound parameters. Drop the SQL `NOT IN`, fetch the four job-stage rows once (the function already scans all matching rows), and exclude members with an in-memory Set; or use `NOT EXISTS` against a temp table of excluded urls. Note: chunking `NOT IN` the way the IN-lists are chunked is NOT equivalent — a correct `NOT IN` needs the intersection across chunks, not the union.

ebarti (commented):
Reviewed the operations read-model this PR adds (vs base feat/pipeline-runtime-observability). One Medium robustness finding inline: two global-scope queries build an unbounded `NOT IN` bound-parameter list from the uncapped backlog-sweep cohort, which can 500 the whole snapshot on large operational states while every sibling reader in the file chunks. Otherwise the change looks sound: the privacy test proves job_url and input-summary secrets are not leaked (jobKey uses the opaque runtime ref / hashed pdf item_key), retries do not double-count (pipeline_step_projections PK is per-item), the ETA estimator reports explicit non-estimate states instead of fabricating zeros, freshness/staleness are surfaced from worker telemetry, and no apply/submission surface is introduced. Every test parses results through PipelineOperationsSnapshotSchema, so the contract is enforced.

ebarti:

PR #453, opened 2026-07-14
refactor(web): prepare Base UI migration

ebarti:
## Summary

- add Base UI alongside the existing Radix dependencies so components can migrate progressively
- isolate the application root so Base UI portals stack above application content
- add an AST-based migration boundary test that keeps direct Radix imports confined to the explicit remaining-wrapper allowlist

## Why

The component layer needs to move from Radix to Base UI before the visual redesign. Establishing coexistence and a shrinking import boundary lets each wrapper migrate independently while the application stays buildable and visually unchanged.

## Impact

This PR intentionally changes no product visuals or component behavior. It only prepares the migration seam and its regression guard.

## Validation

- `corepack pnpm --filter @jobctrl/web exec vitest run src/shared/ui/base-ui-migration-boundary.test.ts` — 3 tests passed
- `corepack pnpm web:check` — passed
- `corepack pnpm web:build` — passed
- focused review gate — PASS, no remaining findings

## Documentation

No public documentation update is warranted for this internal migration foundation. User-facing documentation will be updated at the end of the complete migration and redesign stack.

### DIFF
--- a/apps/web/package.json
+++ b/apps/web/package.json
@@ -20,6 +20,7 @@
     "storybook:test": "start-server-and-test \"http-server ../../dist/web-storybook -p 6006 --silent\" http://127.0.0.1:6006 \"test-storybook --url http://127.0.0.1:6006\""
   },
   "dependencies": {
+    "@base-ui/react": "^1.6.0",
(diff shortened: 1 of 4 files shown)


ebarti:

PR #455, opened 2026-07-14
refactor(web): migrate tabs toggle and scroll area to Base UI

ebarti:
## What changed

- migrate the shared Tabs, Toggle, and ScrollArea wrappers from Radix to Base UI
- preserve current string values, activation modes, RTL navigation, force-mounted tab panels, variants, and consumer APIs through focused adapters
- add focused component tests and Storybook states for the migrated primitives
- keep ToggleGroup on Radix after proving Base UI 1.6 cannot preserve the controlled roving-tab-stop contract used by Job views
- update the direct-Radix boundary and add per-component migration reports

## Why

This is the next compatibility-first slice of the Radix-to-Base migration. It removes three direct Radix wrapper imports without combining the dependency migration with the later visual redesign. ToggleGroup remains intentionally unmigrated rather than introducing a production-reachable keyboard regression when URL or history state changes the selected Job view.

## Validation

- 8 focused Vitest files, 65 tests passed
- corepack pnpm web:check
- corepack pnpm web:build
- git diff --check
- review gate: PASS with 0 Blocker, High, Medium, or Low findings

## Documentation

Public documentation is intentionally deferred to the final PR in this migration and redesign stack. Focused migration reports under .migration record the compatibility decisions in this slice.

### DIFF
--- a/.migration/scroll-area.md
+++ b/.migration/scroll-area.md
@@ -1,0 +1,29 @@
+# scroll-area
+
+2026-07-14 — merge, migrated the customized wrapper to Base UI 1.6 with explicit Content composition and preserved visual classes.
+
+## Changed
+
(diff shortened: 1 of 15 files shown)


ebarti:

PR #456, opened 2026-07-14
refactor(web): migrate overlays and select to Base UI

ebarti:
## What changed

- migrate shared Tooltip, Popover, and Select wrappers from Radix to Base UI 1.6
- preserve existing visual classes and consumer APIs through focused compatibility adapters for delays, positioning, cancellation, portals, formatted labels, controlled state, RTL, and disabled states
- keep compensation-source labels human-readable while persisting canonical source-policy values
- add focused component, accessibility, Storybook, and real-consumer regressions
- remove these three wrappers from the temporary direct-Radix boundary

## Why

This is the next compatibility-first migration slice. It keeps primitive replacement separate from the later visual redesign while shrinking direct Radix ownership from eight wrappers to five.

## Validation

- focused overlay suite: 5 files, 36 tests passed
- full web suite: 279 files, 1,541 tests passed
- corepack pnpm web:check
- corepack pnpm web:build
- git diff --check
- review gate: PASS with 0 Blocker and 0 High findings

## Known compatibility differences

- Base Popover cannot expose the initiating pointerdown through its public Root callback on the default nonmodal outside-click path, so the adapter emits truthful cancellable outside events instead of falsely casting click events to PointerEvent
- focusable outside clicks may produce focus-out and click close phases; this is documented and has no repository consumer
- focus-out compatibility event targets differ from Radix; cancellation remains covered

## Documentation

Public documentation is intentionally deferred to the final PR in this migration and redesign stack. Focused migration reports under .migration record the compatibility decisions in this slice.

### DIFF
--- a/apps/web/src/shared/ui/select.tsx
+++ b/apps/web/src/shared/ui/select.tsx
@@ -1,20 +1,120 @@
-import * as SelectPrimitive from "@radix-ui/react-select";
+import { DirectionProvider } from "@base-ui/react/direction-provider";
+import { Select as SelectPrimitive } from "@base-ui/react/select";
 import { IconCheck, IconChevronDown, IconChevronUp } from "@tabler/icons-react";
 import {
+  Children,
   forwardRef,
-  type ComponentPropsWithoutRef,
+  isValidElement,
   type ComponentRef,
+  type ReactNode,
+  useRef,
 } from "react";
 
 import { cn } from "../lib/cn.js";
 
-export const Select = SelectPrimitive.Root;
+type BaseSelectRootProps = SelectPrimitive.Root.Props<string, false>;
+
+export interface SelectProps extends Omit<
+  BaseSelectRootProps,
+  "children" | "defaultValue" | "items" | "onValueChange" | "value"
+> {
+  children?: ReactNode;
+  defaultValue?: string | undefined;
+  dir?: "ltr" | "rtl" | undefined;
+  items?: BaseSelectRootProps["items"];
+  onValueChange?: ((value: string) => void) | undefined;
+  value?: string | undefined;
+}
+
+type SelectItemDefinition = {
+  label: ReactNode;
+  value: string;
+};
+
+function collectItemDefinitions(children: ReactNode): SelectItemDefinition[] {
+  const items: SelectItemDefinition[] = [];
+
+  Children.forEach(children, (child) => {
+    if (!isValidElement<{ children?: ReactNode; value?: unknown }>(child)) {
+      return;
+    }
+
+    if (child.type === SelectItem && typeof child.props.value === "string") {
+      items.push({ label: child.props.children, value: child.props.value });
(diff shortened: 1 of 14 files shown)


apps/web/src/shared/ui/select.tsx
```diff
@@ -25,111 +125,214 @@ export const SelectTrigger = forwardRef<
...
+      side = "bottom",
+      sideOffset = 0,
+      sticky = "partial",
+      ...props
+    },
+    ref,
+  ) => {
+    const alignsSelectedItem =
+      alignItemWithTrigger ?? position === "item-aligned";
+    const resolvedCollisionAvoidance =
```
ebarti, line 229: PR Reviewer - reviewer-pr456

Severity: Low
Gate: Pass
Evidence: select.tsx:229-232 sets `resolvedCollisionAvoidance = collisionAvoidance` (which is `undefined`) whenever `avoidCollisions` is true (the wrapper default). Base UI then defaults the perpendicular-axis fallback to `'end'` — see the installed source `@base-ui/react/utils/useAnchorPositioning.mjs:80`: `collisionAvoidance.fallbackAxisSide || 'end'`. So on viewport collision the Select popup can flip to a left/right (perpendicular) side. The two sibling wrappers migrated in this same PR do the opposite on the identical `avoidCollisions=true` path: popover.tsx:72-74 and tooltip.tsx:199-201 both pass `{ fallbackAxisSide: "none" }` to preserve Radix's preferred-axis-only behavior (Radix Popper's default is `fallbackAxisSideDirection: "none"`).
Reproduction: Reasoned from code and verified against the Base UI default in installed source (useAnchorPositioning.mjs:80). Not run: no product QA executed (review-only).

The migrated Select diverges from both Radix and the two peer overlay wrappers when the popup cannot fit on its preferred (top/bottom) axis: instead of staying on that axis and shifting within the viewport, Base UI is allowed to fall back to a perpendicular (left/right) side. There is one production consumer today (CompensationSourcePolicyPanel), so real-world impact is limited, but this contradicts the PR's stated compatibility-first, positioning-preserving goal and is inconsistent with popover/tooltip in the same slice. select.md lists the positioning defaults it preserves but omits the collision fallback axis.

Required fix: On the default `avoidCollisions` path, pass a Radix-preserving `{ fallbackAxisSide: "none" }` (mirroring popover.tsx/tooltip.tsx) rather than `undefined`; or, if perpendicular fallback is intended, record it as an accepted behavior change in .migration/select.md.

ebarti (commented):
Automated code-owner review of the Radix->Base UI overlays/select migration. One Low finding on Select collision-avoidance fallback-axis behavior diverging from Radix and from the sibling popover/tooltip wrappers in this same PR. No Blocker/High findings; the sole production Select consumer (CompensationSourcePolicyPanel) is verified compatible. Gate: PASS.

ebarti:

PR #459, opened 2026-07-15
feat(pipeline): persist discovery execution lineage

ebarti:
## Summary

- persist durable discovery execution lineage and source-attempt state
- thread execution lineage through discovery and preparation workflows
- add repository and regression coverage for the lineage contract

## Validation

Not run in this phase. Validation is intentionally deferred to the final cumulative stack after all phases are complete.

### DIFF
--- a/workers/automation/src/jobctrl/pipeline/preparation.py
+++ b/workers/automation/src/jobctrl/pipeline/preparation.py
@@ -10,7 +10,7 @@ from __future__ import annotations
 import asyncio
 import logging
 import sqlite3
-from dataclasses import dataclass
+from dataclasses import dataclass, replace
 from typing import Any, Coroutine
 
 from temporalio import activity
@@ -149,16 +149,19 @@ def start_discovery_preparation_workflows(
         ]
         specs = [spec for _target, spec in target_specs]
         if discovery_execution is not None:
-            workflow_ids_to_start = _record_preparation_work_plans(
+            workflow_cohorts_to_start = _record_preparation_work_plans(
                 targets,
                 tenant_id=tenant_id,
                 discovery_execution=discovery_execution,
                 discovery_cohort_kind=discovery_cohort_kind,
             )
             specs = [
-                spec
+                _with_discovery_cohort(
+                    spec,
+                    workflow_cohorts_to_start[spec.workflow_id],
+                )
                 for _target, spec in target_specs
-                if spec.workflow_id in workflow_ids_to_start
+                if spec.workflow_id in workflow_cohorts_to_start
             ]
             if finalize_observed_work_plans:
                 _finalize_unplanned_observed_work_plans(
@@ -235,7 +238,7 @@ def start_job_preparation_workflow(
             preparation_payload = spec.args[0]
             if not isinstance(preparation_payload, JobPreparationInput):
                 raise TypeError("preparation workflow spec has an unexpected input")
-            workflow_ids_to_start = _record_preparation_work_plans(
+            workflow_cohorts_to_start = _record_preparation_work_plans(
                 [
                     PreparationTarget(
                         job_url=job_url,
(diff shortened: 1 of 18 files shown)


workers/automation/src/jobctrl/pipeline/preparation.py
```diff
@@ -355,6 +449,197 @@ def _workflow_spec_for_target(
...
+    for target in targets:
+        if discovery_cohort_kind == "existing_backlog":
+            membership = repository.link_job(
+                discovery_execution,
+                target.job_url,
+                cohort_kind="existing_backlog",
+            )
+        else:
+            membership = repository.get(discovery_execution, target.job_url)
+            if membership is None:
```
ebarti, line 488: PR Reviewer - reviewer-pr459

Severity: High
Gate: Fail
Evidence: `_record_preparation_work_plans` (observed cohort) hard-raises `RuntimeError("Observed preparation target is missing discovery execution membership")` when `repository.get()` returns None. But `derive_preparation_targets` -> `_derive_targets` returns EVERY `pending_score` job in the DB (`get_jobs_by_stage(conn, stage="pending_score")`), while memberships are created only by source crawls (`attach_source_observation` -> `link_job`, grepped: jobspy/smartextract/workday/ats + the pre-run backlog sweep). Detail enrichment does NOT create memberships, and `run_enrichment` drains the PERSISTENT `job_enrichments` pending view (detail.py:2441-2445 "rows without an enrichment record OR current_status = 'pending'"), which includes jobs discovered in a prior run (or predating this feature) that are not re-observed this run. When such a job is enriched to `pending_score` during the run: (1) the per-job handoff's RuntimeError is swallowed as best-effort (enrichment/detail.py:1466), so it gets no membership and no prep; (2) the streaming and terminal fan-outs then derive the same job and raise here, aborting the ENTIRE fan-out inside the try before any workflow starts and before `_finalize_unplanned_observed_work_plans` runs; the `except` handler (preparation.py:168-173) then marks every pending membership `failed` via `_mark_pending_work_plans_failed`. DiscoverWorkflow tolerates the resulting ActivityError (discovery/workflow.py:284-287), so discovery completes but the run's reconciling backstop (I1 convergence) and lineage finalization silently fail and pending lineage rows are mass-marked `failed` -- corrupting the exact audit data this PR exists to produce. On the first run after this ships against an existing DB, every pre-existing pending-enrich job has no membership, making this near-certain to trigger.
Reproduction: Reasoned from code (grepped all `attach_source_observation`/`link_job` callers; confirmed enrichment creates no membership; confirmed enrich queue is DB-derived/persistent; traced abort -> except -> mass-fail -> tolerated ActivityError). Not run: review-only, no dev stack.

The observed-cohort invariant "current-run rows must already exist from the source-observation boundary" does not hold for jobs that reach `pending_score` via enrichment of an un-re-observed backlog entry. The comment's rationale ("silently inventing those rows here would lose their first source metadata") does not apply to a job that was never observed this run -- there is no first-observation metadata to lose, so linking it is safe. No test covers a `pending_score` target with no execution membership.

Required fix: Do not abort the whole observed-cohort fan-out on one missing membership. For an observed target that legitimately reached `pending_score` without a source re-observation this run, link it on demand (source_family/source_run_id NULL is correct), or skip only that target and continue; ensure `_finalize_unplanned_observed_work_plans` still runs. Add a regression test that drives a `pending_score` target with no membership through the streaming/terminal fan-out and asserts other targets still start and lineage is finalized rather than mass-failed.

ebarti (commented):
Automated review of the discovery execution-lineage change. Temporal serialization (nested DiscoveryExecutionRef + Literal cohort fields under temporalio 1.26.0), workflow determinism (workflow.info() captured once, no new/reordered commands), the v2->v3 schema stamping, and idempotent link/promote/work-plan logic all check out. One High correctness finding on the observed-cohort membership invariant is inline.

ebarti:

PR #460, opened 2026-07-15
feat(pipeline): add runtime observability

ebarti:
## Summary

- add worker and activity runtime telemetry with Temporal task-queue observation
- emit, persist, and project pipeline-step lifecycle events across backend and API boundaries
- refresh affected frontend pipeline state from the new operations events
- count every occupied worker slot in active-capacity reporting

## Stack

This is phase P1 of the stacked pipeline observability rollout. It targets `feat/pipeline-execution-lineage-rhea` (PR #459).

## Validation

Validation is intentionally deferred to the final cumulative stack. No tests, builds, typechecks, or QA were run for this intermediate phase.


workers/automation/src/jobctrl/infrastructure/temporal/pipeline_step_lifecycle.py
```diff
@@ -0,0 +1,294 @@
...
+
+
+def pdf_pipeline_step_item_key(idempotency_key: str) -> str:
+    """Derive a stable, non-reversible PDF scope without persisting a job URL."""
+
+    digest = hashlib.sha256(idempotency_key.encode("utf-8", errors="replace")).hexdigest()
+    return f"pdf:{digest}"
+
+
+def _durable_writer(scope: PipelineStepScope) -> EventWriter:
```
ebarti, line 236: PR Reviewer - reviewer-pr460

Severity: Medium
Gate: Pass
Evidence: `_durable_writer` runs `assert_activity_runtime` + a `job_events` insert/commit + a full `ProjectionBuilder(...).refresh()`. It backs every emission, and the activity call sites invoke those emissions unguarded: `begin_pipeline_step_attempt` (queued+started) runs before the real work, and `PipelineStepAttempt.completed()/failed()` run after it, outside any try/except (discovery/activities.py plan/source_family/enrichment/fanout; materials/activities.py render_pdf). In the except blocks, `lifecycle.failed_from_exception(app_error)` is called *before* `raise app_error`.
Reproduction: Reasoned from code (not run).

Because these writes are fail-hard, a telemetry-write/refresh error (DB lock past the 10s busy_timeout, disk-full, or any projection-builder bug that `refresh()` re-runs) has two bad effects: (1) on the failure path the telemetry exception propagates and *replaces the real activity error* before `raise app_error` runs — the exact "observability dies on the failure path it exists to observe" case; (2) on the success path a throwing `completed()` turns a finished activity into a Temporal retry, re-running source crawls / enrichment / PDF render. This is inconsistent with the sibling `ActivityRuntimeTelemetryInterceptor`, which deliberately swallows all exceptions ("Runtime telemetry must never change activity execution behavior"), and with the existing contact/interview `refresh()` calls, which only run on the success path (never inside an except).

Required fix: make lifecycle emission best-effort — catch+log inside `_durable_writer` (or wrap the begin/completed/failed calls at each activity site) so a telemetry failure can never mask the real activity error or fail an otherwise-successful activity. If durability is intentional, at minimum guarantee the terminal `failed` emission never supersedes the original activity exception.

ebarti (commented):
PR Reviewer - reviewer-pr460

Reviewed the full PR 460 diff (35 files) against base feat/pipeline-execution-lineage-rhea. Event contracts (TS/Python) are symmetric and content-free with bounded validation; workflow instrumentation is deterministic (telemetry lives entirely in activities/interceptors, never in workflow code); the projection fold is attempt-aware, first-terminal-wins, and cross-language parity is pinned by a shared fixture; the activity inventory and task-queue sampler are concurrency-safe and never persist RPC/exception text. Strong test coverage. One Medium robustness finding on the durable emission path is inline. Gate: PASS (no Blocker/High).

ebarti:

PR #461, opened 2026-07-15
feat(pipeline): expose operations snapshot

ebarti:
## Summary

- add pipeline operations and ETA read-model aggregation
- expose the operations snapshot through API, contract, and client surfaces
- carry focused API coverage for the new read-model behavior

## Why

This gives later operations UI phases a stable view of stage backlog, capacity, throughput, and estimated completion without coupling the frontend to raw runtime telemetry.

## Stack

This PR is stacked on #460 and targets `feat/pipeline-runtime-observability-rhea`.

## Validation

Validation is intentionally deferred to the final cumulative stack under the approved stacked-delivery plan. No tests, builds, typechecks, or QA were run for this intermediate phase.

### DIFF
--- a/apps/api/src/pipeline-operations.ts
+++ b/apps/api/src/pipeline-operations.ts
@@ -1,0 +1,1921 @@
+import { createHash } from "node:crypto";
+
+import type {
+  DiscoveryExecutionSummary,
+  PipelineActiveItem,
+  PipelineCapacity,
+  PipelineEta,
+  PipelineExecutionCohortSummary,
+  PipelineOperationsFreshness,
+  PipelineOperationsSnapshot,
+  PipelineOperationalStage,
+  PipelineStageCounts,
+} from "./contracts.js";
+import { allRows, tableExists, type SqliteDatabase, type SqliteValue } from "./db.js";
+import {
+  estimatePipelineEta,
+  type PipelineEtaEstimatorInput,
+  type PipelineEtaStageInput,
+} from "./pipeline-eta.js";
+import { refreshProjections } from "./projections.js";
+import { readWorkerRuntimeTelemetry, type WorkerRuntimeTelemetrySnapshot } from "./worker-runtime-telemetry.js";
+import { readLlmSpendHealth } from "./worker-health.js";
+
+const DEFAULT_TENANT = "local";
+const ETA_SAMPLE_LIMIT = 50;
+const ETA_SAMPLE_WINDOW_MS = 14 * 24 * 60 * 60 * 1_000;
+
+const JOB_STAGES = ["enrich", "score", "tailor", "cover"] as const;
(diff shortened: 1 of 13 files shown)


apps/api/src/pipeline-operations.ts
```diff
@@ -0,0 +1,1921 @@
...
+  selected: SelectedExecution,
+): Map<(typeof OPERATIONAL_STAGES)[number], PipelineStageCounts> {
+  const counts = new Map<(typeof OPERATIONAL_STAGES)[number], PipelineStageCounts>();
+  for (const stage of OPERATIONAL_STAGES) counts.set(stage, emptyCounts());
+  if (!tableExists(db, "job_stage_states")) return counts;
+  const excluded = selected.current.members.concat(selected.sweep.members).map((member) => member.job_url);
+  const where = ["jss.stage IN ('enrich', 'score', 'tailor', 'cover')"];
+  const params: SqliteValue[] = [];
+  if (excluded.length > 0) {
+    where.push(`jss.job_url NOT IN (${excluded.map(() => "?").join(", ")})`);
```
ebarti, line 828: PR Reviewer - reviewer-pr461

Severity: Low
Gate: Pass
Evidence: `globalStageCounts` (this line) and `loadGlobalRetryability` (apps/api/src/pipeline-operations.ts:857) both build `job_url NOT IN (${excluded.map(() => "?")})` from the selected execution's full member set (current + sweep) in a single prepared statement, without chunking. Every other multi-value query in this file chunks at 500: `loadStageStates` (:351), `loadPreparationWorkflows` (:382), `loadUniqueGlobalPreparationOwners` (:898), `loadJobDisplays` (:1295). better-sqlite3's bundled SQLite caps host parameters (SQLITE_MAX_VARIABLE_NUMBER, 32766). If the selected discovery execution's member set exceeds that cap, the prepared statement throws and `withDb` (apps/api/src/server.ts:3779) returns `500 db_read_failed` for the entire `/v1/pipeline/operations` endpoint.
Reproduction: Reasoned from code. Not run: would require a discovery execution with a member set larger than the SQLite variable cap.

The excluded-members `NOT IN` list here is the only place in this file that is not chunked, so a sufficiently large selected execution makes the endpoint hard-fail (500) instead of degrading. This is unlikely on a typical single-user database but is a latent robustness gap that contradicts the chunking convention used by every other loader in the module.

Required fix: Chunk or restructure the exclusion the same way the other loaders do (e.g., accumulate counts/rows across `chunks(excluded, 500)` with a `NOT IN` per chunk, or express the exclusion as a join/anti-join against the membership rows) so the endpoint stays correct for large executions.

ebarti (commented):
Automated review of PR #461 (feat(pipeline): expose operations snapshot). One Low-severity robustness finding; see inline comment. No Blocker/High issues found.

ebarti:

PR #468, opened 2026-07-17
feat(discovery): adopt JobStreaming event boundary

ebarti:
## Summary

- replace `python-jobspy` with pinned `jobstreaming==0.0.2`
- isolate the provider contract behind a JobCtrl infrastructure adapter
- expose explicit-ack streams for durable consumers while keeping the temporary batch path checkpoint-free
- preserve partial board results and surface typed board failures separately from failed search combinations
- update setup diagnostics, dependency locking, and the active delivery plan

## Why

This is phase 1 of the resumable discovery stack. It establishes a typed provider boundary without allowing a checkpoint to advance ahead of JobCtrl persistence. Durable search-unit storage and Temporal retry/resume build on this seam in the next two PRs.

## Stack

- Base: #467
- Phase 1/3: provider boundary (this PR)
- Phase 2/3: durable search units, fenced SQLite checkpoints, and accepted-result accounting
- Phase 3/3: Temporal resume, cancellation/error handling, canonical docs, and cumulative fault QA

Canonical user and architecture documentation is intentionally deferred to phase 3 because the stack is unreleased until the resumable execution path lands.

## Validation

- `uv --project workers/automation run --extra dev ruff check ...` (touched Python surfaces)
- `uv --project workers/automation run --extra dev pytest -q workers/automation/tests/test_jobstreaming_gateway.py workers/automation/tests/test_jobspy_gateway_boundary.py workers/automation/tests/test_discovery_limits.py workers/automation/tests/test_p1b_error_inversion.py workers/automation/tests/test_crawl_politeness_config.py` — 63 passed
- `uv lock --project workers/automation --check`
- `git diff --check`
- independent review: PASS, no findings

No live external-board crawl was run in this phase.

PR #469, opened 2026-07-17
feat(discovery): persist resumable search units

ebarti:
## Summary

Phase 2 of the three-PR resumable JobStreaming discovery stack, based on #468 (which is based on #467).

- persist immutable query/location/board search units under the exact Discover workflow and Temporal run
- store opaque JobStreaming checkpoints with revision compare-and-swap
- fence stale activity attempts with an execution-wide Temporal-attempt high-water mark and per-unit lease epoch
- persist replay-safe accepted-job receipts and durable run-wide new/existing counts
- defer cursor resets until a provider error has been acknowledged, then apply the reset on the next fenced attempt
- make accepted-result side effects fence-aware and give their durable events stable consumption keys so replay cannot inflate audit history
- represent limit-closed work explicitly as skipped instead of leaving ambiguous pending rows
- keep the TypeScript API schema-version guard aligned with the Python worker's v4 durability schema

## Why

A retried broad-board activity currently starts its family crawl over with only in-memory progress. This layer establishes the caller-owned correctness boundary needed for a killed activity to resume safely in Phase 3: immutable work, monotonic ownership, durable consumption state, and idempotent replay.

## Validation

- `ruff check` on every touched Python surface
- 179 focused discovery, persistence, compensation, state, and migration tests passed
- regressions cover execution-wide stale-attempt fencing, checkpoint CAS, error-ack/reset ordering, mid-event supersession, the compensation commit window, replay event cardinality, cancellation, terminal failure, and durable limit counts
- API typecheck and 10 schema-version guard tests passed after aligning the v4 guard
- `git diff --check`
- independent review: `Gate: PASS; Blocker=0 High=0 Medium=0 Low=0`

Canonical user and architecture documentation remains intentionally deferred to the final PR because these phases form one unreleased stack.

### DIFF
--- a/workers/automation/src/jobctrl/database.py
+++ b/workers/automation/src/jobctrl/database.py
@@ -28,7 +28,7 @@ from jobctrl.config import DB_PATH, DEFAULTS, migrate_legacy_job_tables
 # the ninth-context canonical tables (``ensure_contact_tables``).
 # v3 (Discovery execution lineage): immutable Temporal execution/job membership
 # and the idempotently filled preparation work plan used by Operations.
-SCHEMA_VERSION = 3
+SCHEMA_VERSION = 4
 
 
 class IncompatibleSchemaVersionError(RuntimeError):
@@ -287,6 +287,7 @@ def init_db(db_path: Path | str | None = None) -> sqlite3.Connection:
     ensure_operational_metric_tables(conn)
     ensure_source_observation_tables(conn)
     ensure_discovery_execution_tables(conn)
+    ensure_discovery_search_unit_tables(conn)
     ensure_discovery_control_tables(conn)
     ensure_discovery_settings_tables(conn)
     ensure_contact_tables(conn)
@@ -893,6 +894,7 @@ def ensure_state_tables(conn: sqlite3.Connection | None = None) -> list[str]:
             payload_json        TEXT,
             entity_kind         TEXT,
             entity_ref          TEXT,
+            idempotency_key     TEXT,
             FOREIGN KEY (job_url) REFERENCES jobs(url) ON DELETE CASCADE
         )
     """)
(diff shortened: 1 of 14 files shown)


workers/automation/src/jobctrl/database.py
```diff
@@ -28,7 +28,7 @@
 # the ninth-context canonical tables (``ensure_contact_tables``).
 # v3 (Discovery execution lineage): immutable Temporal execution/job membership
 # and the idempotently filled preparation work plan used by Operations.
-SCHEMA_VERSION = 3
+SCHEMA_VERSION = 4
```
ebarti, line 31: PR Reviewer - pr469-reviewer

Severity: Low
Gate: Pass
Evidence: Lines 24-31 maintain a deliberate schema-version ledger (v2 Contact & Outreach, v3 Discovery execution lineage each described), but SCHEMA_VERSION is bumped to 4 with no `# v4 (...)` entry. v4 actually introduces discovery_search_units + discovery_search_unit_jobs and the job_events.idempotency_key column/partial-unique-index.

The version ledger comment is now stale: a future maintainer bumping to v5 won't have a record of what v4 changed, which is the whole purpose of this ledger block.

Required fix: Add a `# v4 (...)` line summarizing the durable search-unit tables and the job_events idempotency_key column/index.

workers/automation/src/jobctrl/database.py
```diff
@@ -287,6 +287,7 @@ def init_db(db_path: Path | str | None = None) -> sqlite3.Connection:
     ensure_operational_metric_tables(conn)
     ensure_source_observation_tables(conn)
     ensure_discovery_execution_tables(conn)
+    ensure_discovery_search_unit_tables(conn)
```
ebarti, line 290: PR Reviewer - pr469-reviewer

Severity: Low
Gate: Pass
Evidence: The v3->v4 upgrade adds discovery_search_units/discovery_search_unit_jobs (CREATE IF NOT EXISTS), the job_events.idempotency_key column (ALTER, database.py:910), and a partial UNIQUE index on a potentially populated job_events table (database.py:913-915). The new test file only exercises fresh init_db (v4 from scratch); no committed test starts from a populated pre-v4 DB. I confirmed the in-place upgrade is correct via a standalone sqlite3 simulation (existing rows preserved, user_version 3->4, NULL keys never dedupe, duplicate non-NULL keys dedupe), but that guarantee is not protected by a repo test.
Reproduction: Reasoned from code + verified with an out-of-tree sqlite3 simulation (not the repo suite).

This is a resume-critical, cross-stack schema change (the API guard in apps/api/src/db.ts moves in lockstep), and the partial-unique-index-on-existing-data path is a new migration pattern here, yet it has no forward-migration regression test.

Required fix: Add a focused test that stamps a pre-v4 user_version on a DB holding existing job_events rows, runs init_db, and asserts the new tables/column/index exist, existing rows survive, and record_job_event dedupes on a repeated non-NULL idempotency_key.

ebarti (commented):
Reviewed the Phase 2 delta over feat/jobstreaming-adapter-foundation (durable search units, fencing, checkpoint CAS, accepted-job receipts). The core persistence and concurrency logic is correct and well covered: reclaim fencing, execution-wide attempt watermark, checkpoint compare-and-swap, deferred cursor reset, replay-safe receipts (MAX(was_new) + idempotency_key), and mid-event supersession + repair. I ran the new suite (11 passed), ruff (clean on all changed files), and verified the v3->v4 upgrade-in-place is safe via a standalone sqlite3 simulation. Two Low, non-blocking notes on the schema migration below; nothing blocks merge.

PR #474, opened 2026-07-20
feat(apply): prevent repeat applications

ebarti:
## Summary

- block confirmed repeat applications to the same canonical job or accepted duplicate identity
- require an explicit, reasoned, evidence-bound one-attempt confirmation for same-employer equivalent roles
- preserve distinct-role eligibility and exclude dry runs, failed pre-submit attempts, pending suggestions, notes, and unverified assumptions
- enforce the decision atomically at the worker claim boundary while preserving approval, at-most-once, submit-intent, and needs-verification protections
- expose inspectable prior facts, identity evidence, immutable audit history, override state, and actionable UI states

## Why

Alternate source URLs and representations can rediscover an opening after an application was already confirmed. The apply boundary now uses canonical identity and confirmed application facts to prevent an unintended second live submission without blocking legitimate applications to distinct roles.

Manual applied marks are recorded as user-attested facts, not worker submission events.

## Migration and safety

- advances the SQLite schema from v5 to v6 with additive repeat-override, consumption, and audit tables
- additively restores the stage-state metadata column for older databases
- preserves historical application facts unchanged
- consumes an override once per unique apply attempt while retaining the workflow correlation ID
- keeps `--gen` inspection-only: no job claim, override consumption, browser command, or submit-capable MCP configuration
- no real application was submitted during validation

## Validation

- `corepack pnpm check`
- `corepack pnpm --filter @jobctrl/api test` — 646 passed
- `corepack pnpm --filter @jobctrl/web test` — 1,952 passed
- `corepack pnpm --filter @jobctrl/web build`
- `corepack pnpm docs:build` — 8,245 references resolved
- focused Python apply matrix — 78 passed
- focused Apply Review component suite — 52 passed
- isolated Playwright repeat-application flow — 1 passed through UI block, reasoned override, Python worker consumption, second-attempt refusal, and zero target `ApplicationSubmitted` facts
- independent review gate — PASS, no findings
- independent QA gate — PASS, no findings
- `git diff --check`


workers/automation/src/jobctrl/domain/apply/repeat_application.py
```diff
@@ -0,0 +1,630 @@
...
+                   CASE event_type
+                     WHEN 'ApplicationSubmitted' THEN 'application_submitted'
+                     ELSE 'application_manually_marked'
+                   END AS fact_kind,
+                   'event:' || event_id AS fact_id,
+                   occurred_at AS confirmed_at,
+                   CASE event_type WHEN 'ApplicationSubmitted' THEN 40 ELSE 30 END AS priority
+              FROM job_events
+             WHERE job_url IS NOT NULL
+               AND event_type IN ('ApplicationSubmitted', 'ApplicationManuallyMarked')
```
ebarti, line 341: PR Reviewer - pr474-review-worker

Severity: Medium
Gate: Pass
Evidence: `_confirmed_application_facts` admits only ApplicationSubmitted / ApplicationManuallyMarked / applied_confirmation / legacy applied (same in the TS twin, apps/api/src/repeat-application.ts lines 324-371). A prior attempt that crashed after submit-intent leaves apply state `needs_verification` with only an `ApplySubmitIntended` event, i.e. no fact here. `_has_needs_verification_apply` (launcher.py:697) guards only the same `job_url`. A canonically identical sibling (matching `job_canonical_identities` or `job_duplicate_links`, different URL) therefore evaluates `clear`, and the standing loop can claim and live-submit it while the first attempt's outcome is still unknown - the exact duplicate this PR exists to prevent, with no confirmation prompt at all. The 2026-07-20 ADR excludes "submit intent alone" from confirmed facts (correct for permanent history) and defers to ambiguous-crash verification, but that verification does not span canonical identities.
Reproduction: Reasoned from code; not run (requires a mid-submit crash plus a canonical sibling in the queue).

Unverified prior attempts (needs_verification / submit-intent without a terminal event) do not participate in repeat evaluation at all, so a canonical-identity sibling of an ambiguous attempt is treated as `clear` for the whole verification window.

Required fix: treat pending-verification priors related by canonical_job / canonical_identity / accepted_duplicate as `confirmation_required` (a temporary gate, not a confirmed fact) until verification resolves - or explicitly record this window as accepted residual risk in the ADR and security docs.

apps/api/src/application-feedback.ts
```diff
@@ -934,6 +939,7 @@ function reviewQueueItemFromRow(
       ready: applyAudit.state === "ready",
     },
     applyAudit,
+    repeatApplication: evaluateRepeatApplication(db, row.job_id),
```
ebarti, line 942: PR Reviewer - pr474-review-worker

Severity: Medium
Gate: Pass
Evidence: this line runs the full evaluation for every queue item on every `GET /v1/apply/review-queue` (server.ts:954), inside `GET /v1/digest` (server.ts:568 builds the queue too), and once more per job detail (read-model.ts:1181). Each evaluation re-scans all confirmed facts (three table scans) and, per fact, runs jobIdentity + canonical-identity lookup + a full `job_duplicate_links` table load plus two `job_source_observations` queries (repeat-application.ts:387-516). That is O(queue items x confirmed applications) synchronous better-sqlite3 calls on the event loop, and it grows quadratically as confirmed applications accumulate - which is the product's success path. `ensureRepeatApplicationTables` (repeat-application.ts:72) also executes the 5-statement DDL block on every call; the Python twin short-circuits when the tables exist (repeat_application.py:63) but the TS version does not.
Reproduction: Reasoned from code; not run.

Repeat evaluation is recomputed from scratch per item on hot read paths with unbounded fan-out over confirmed application history and unconditional DDL per call.

Required fix: add the table-existence short-circuit to the TS ensure, compute the confirmed-facts snapshot once per queue/digest build and share it across items, and restrict duplicate-link matching to links touching the two candidate keys instead of loading the whole table per pair.

ebarti (commented):
Deep review of the repeat-application prevention stack: TS evaluator + Python claim-boundary twin, schema v5->v6 migration, dispatch gate, Apply Review guard UI, and the cross-stack tests.

Verified during review: the fingerprint parity fixture is exercised by both languages; override consumption happens inside the worker's `BEGIN IMMEDIATE` claim with PK/UNIQUE guards (I ran `tests/test_repeat_application_prevention.py` in the PR worktree: 14/14 pass, including the concurrent-claim and unique-run-id batch tests); the approval-refusal `conn.commit()` only persists idempotent audit rows (no other writes precede it); the `run_id` decoupling from `workflow_id` preserves workflow correlation (payload `workflow_id` + reaper fallback `workflow_id or run_id`) and removes a latent `UNIQUE(tenant_id, run_id)` collision for continuous auto-apply claiming multiple jobs; `--gen` is inspection-only (no claim, `{"mcpServers": {}}`, dry-run prompt); repeat preconditions map to 409 with actionable messages; the v6 migration is additive and the test proves application facts stay byte-identical; parity breaks fail closed (worker refuses rather than submits).

No Blocker or High findings. Two Medium observations inline: (1) unverified/ambiguous prior attempts are invisible to the evaluator across canonical siblings, leaving a narrow double-submit window that the same-job-key needs_verification guard does not cover; (2) per-item repeat evaluation on review-queue/digest/job-detail reads scales O(items x confirmed priors) with unconditional DDL per call on the TS side.

PR #489, opened 2026-07-21
test(web): refresh repeat application heartbeat

ebarti:
## Summary
- refresh the test-owned worker heartbeat immediately before the post-confirmation simulated live-submit request
- retain the production worker-readiness gate and every repeat-application safety assertion

## Validation
- corepack pnpm web:check
- corepack pnpm --filter @jobctrl/web test-d
- focused Playwright: repeat-application and worker-heartbeat fixture specs (2 passed)
- git diff --check


ebarti:
Superseded by #488, which now contains the same focused heartbeat fixture fix together with the Python Playwright provisioning change required for a single green TypeScript CI gate.

PR #538, opened 2026-07-30
feat(pipeline): migrate stage-state references to stable JobIds

ebarti:
## Summary

- migrate `job_stage_states` from URL-shaped keys to `(tenant_id, job_id, stage)` in schema v13
- preserve the latest lifecycle fact while retaining maximum attempt counters and a uniform aggregate optimistic-lock version during alias collisions
- update Python and TypeScript readers/writers for stable references, with bounded API compatibility for schema v12 databases
- cover rollback/retry, UUID-shaped URL collisions, runtime identity merges, aggregate saves, API mutation/delete isolation, and projection/read-model compatibility

## Why

Stage state is the canonical pipeline lifecycle authority. Keeping it keyed by posting URL left identity aliases able to split one Job aggregate across rows and made collision cleanup unsafe. This phase attaches lifecycle state to the stable JobId while preserving existing facts and concurrency guards.

## Validation

- `workers/automation/.venv/bin/python -m pytest -q` — 2681 passed, 2 skipped
- `pnpm --filter @jobctrl/api test` — 662 passed
- `workers/automation/.venv/bin/ruff check src tests` — passed
- `pnpm --filter @jobctrl/api check` — passed
- `git diff --check` — passed
- independent Tier 3 review — Gate: PASS, 0 findings

## Stack

- stacked on #537 (`feat/job-id-scoring-references`)
- canonical product docs and cumulative product QA remain deferred to the final approved stack PR

PR #539, opened 2026-07-30
feat(artifacts): migrate registry references to stable JobIds

ebarti:
## Summary

- add schema v14 to migrate the generic `job_artifacts` registry from `job_url` to tenant-scoped `(tenant_id, job_id)` references
- preserve one current registration per stable registry key with deterministic last-write-wins collision handling, plus transactional verification and retry safety
- update Python artifact recording/projection/identity-collapse paths and TypeScript artifact reads/permanent deletion for v13/v14 compatibility
- keep UUID-shaped posting URLs URL-first across both runtimes and raise the API schema guard to the worker-owned v14 contract

## Why

Artifact registrations belong to a stable Job aggregate, not to a mutable posting URL. Keeping this registry URL-keyed allowed aliases to split one job's artifacts and made identity collapse unsafe. This phase moves only the generic registry; canonical materials, provenance, and layout references remain separate reviewable stack layers.

## Migration behavior

- stamps `PRAGMA user_version = 14` only after schema, row-count, UUID, foreign-key, and orphan verification succeed
- rolls back cleanly and can retry after an injected verification failure
- merges aliases that resolve to the same `(tenant, job, stage, type, path)` using newest `created_at`, then largest `artifact_id`, matching runtime upsert semantics
- preserves distinct registry keys and reassigns registrations safely during runtime identity collapse

## Validation

- focused Python migration/runtime/projection suite — 19 passed locally; independent reviewer rerun an expanded 34-test set
- full Python suite — 2684 passed, 2 skipped
- focused v14 artifact + schema-open Vitest suites — 12 passed
- full API suite — 661/663 passed in the restricted sandbox; the two environment-sensitive loopback/process-tree cases were rerun outside the sandbox and passed
- `corepack pnpm api:check` — passed
- `git diff --check` — passed
- independent Tier 3 review — Gate: PASS, Blocker 0 / High 0 / Medium 0 / Low 0

## Residual risk

The v14 TypeScript artifact-read SQL is statically traced through the same URL-first resolver exercised by the v14 API regression; there is not yet a separate v14 projection-read fixture.

## Stack

- stacked on #538 (`feat/job-id-stage-state-references`)
- canonical product docs and cumulative product QA remain deferred to the final approved stack PR
- core materials, provenance, and layout references follow as smaller PRs

### DIFF
--- a/apps/api/src/application-feedback.ts
+++ b/apps/api/src/application-feedback.ts
@@ -50,6 +50,7 @@ import {
   hasCompositeJobIdForeignKey,
   jobReferenceColumn,
   jobReferenceJoinToJobs,
+  jobReferencePredicateForUrl,
   tableExists,
   type SqliteDatabase,
   type SqliteValue,
@@ -2663,6 +2664,11 @@ function materialArtifactCandidates(
   }
 
   if (tableExists(db, "job_artifacts")) {
+    const artifactReference = jobReferencePredicateForUrl(
+      db,
+      "job_artifacts",
+      jobKey,
+    );
     const rows = allRows<{
       created_at: string | null;
       path: string | null;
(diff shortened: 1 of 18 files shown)


ebarti:

PR #543, opened 2026-07-30
feat(interview): migrate preparation references to stable JobIds

ebarti:
## Summary

- migrate Interview Preparation histories and item audit rows to tenant-scoped stable `JobId` references in schema v18
- preserve complete alias histories, deterministic generation ordering, accepted/superseded/failed semantics, child audit data, UUID-shaped URL compatibility, and tenant isolation
- remap immutable application-outcome generation links during both v17→v18 migration and live identity collisions; canonicalize only a losing URL-valued outcome key so it remains reachable while the outcome family stays URL-keyed until Phase 4
- keep losing-alias cleanup and rediscovery compatibility intact, and explicitly delete prep children when permanent deletion runs with SQLite FK enforcement disabled

## Validation

- Python full suite: `2700 passed, 2 skipped`
- API full suite: `675 passed`
- migration-chain suite: `56 passed`
- focused Python interview/audit suite: `14 passed`
- real `resolve_all_urls()` collision regression plus existing alias cleanup/rediscovery regression: `5 passed`
- TypeScript API check: passed
- Ruff: passed
- `git diff --check`: passed
- independent reviewer: `PASS` — Blocker 0, High 0, Medium 0, Low 0
- independent Tier-3 QA: `PASS` — Blocker 0, High 0, Medium 0, Low 0

Canonical product documentation and cumulative end-to-end QA remain intentionally deferred to the final PR of this approved unreleased stack.

### DIFF
--- a/apps/api/src/application-feedback.ts
+++ b/apps/api/src/application-feedback.ts
@@ -841,16 +841,22 @@ function resolveInterviewPrepGeneration(
   if (!tableExists(db, "job_interview_prep")) {
     throw new InputError("Interview prep generation not found.");
   }
+  const reference = jobReferencePredicateForUrl(
+    db,
+    "job_interview_prep",
+    jobUrl,
+    DEFAULT_TENANT,
+  );
   const row = getRow<{ generation: number }>(
     db,
     `SELECT generation
        FROM job_interview_prep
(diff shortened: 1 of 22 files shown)


ebarti:

PR #544, opened 2026-07-30
fix(projections): preserve tenant-scoped JobId lookups

ebarti:
## Summary

- scope projection event selection, first-run backfill, and defensive job rebuilds to the builder tenant
- keep the existing `operations_projections` watermark for the local tenant while giving non-local tenants independent watermark keys
- make every stable JobId projection lookup require an explicit tenant, removing the unsafe implicit `local` fallback
- prepare posted and market compensation projection reads for the next stable-reference schema while retaining legacy URL-table compatibility

## Why

A non-local projection builder could consume another tenant's job URL, pass it into a tenant-scoped stable-identity resolver, and fail refresh. A shared watermark could then let one tenant advance past another tenant's events. This prerequisite closes that isolation and availability gap before the compensation-reference migration lands.

## Validation

- full Python suite: `2705 passed, 2 skipped`
- focused mixed-tenant projection and compensation migration suite: `20 passed`
- Ruff: passed
- `git diff --check`: passed
- independent review: `PASS` — Blocker 0, High 0, Medium 0, Low 0
- independent Tier-3 QA: `PASS` — Blocker 0, High 0, Medium 0, Low 0

## Stack

- base: #543
- the schema-v19 compensation migration follows as a separate small PR on this branch
- canonical product docs and cumulative end-to-end QA remain deferred to the final PR of the approved unreleased stack

### DIFF
--- a/workers/automation/src/jobctrl/infrastructure/projections/projection_builder.py
+++ b/workers/automation/src/jobctrl/infrastructure/projections/projection_builder.py
@@ -680,6 +680,15 @@ class ProjectionBuilder:
         ensure_projection_tables(boot_conn)
         boot_conn.commit()
 
+    @property
+    def _watermark_name(self) -> str:
+        tenant = str(self._tenant_id)
+        return (
+            PROJECTION_NAME
+            if tenant == str(LOCAL_TENANT)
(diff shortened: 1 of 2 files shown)


ebarti:

PR #566, opened 2026-07-30
feat(identity): define historical event JobId upcast

ebarti:
## Summary

- define one deterministic, tenant-scoped historical event identity upcaster in Python and TypeScript
- resolve pre-v30 event columns and root payload identity aliases URL-first, including UUID-shaped posting URLs
- fail closed on unresolved, malformed, conflicting, or cross-source primary identity
- preserve locator fields, nested related identities, and prototype-named JSON keys
- lock both runtimes to one shared parity fixture

## Why

The v29 to v30 database migration must translate immutable historical event ownership before replacing the event table. This upcaster exists only inside that forward migration. Current runtime readers and writers accept schema v30 and never use it as a compatibility path.

## Validation

- Python shared-fixture and migration tests: 35 passed
- Full API suite including the TypeScript shared fixture: 62 files, 748 tests passed
- API TypeScript check: passed
- Focused Ruff: passed
- Diff check: passed
- Independent review gate: PASS
- Independent QA gate: PASS

## Stack

- Base: #558
- The v29 to v30 event-table migration follows in #567.
- These PRs are review slices of one cumulative local upgrade, not independently deployed application versions.

PR #567, opened 2026-07-30
feat(identity): migrate durable event ownership

ebarti:
## Summary

- add the forward-only v29 to v30 `job_events` table replacement for the local JobCtrl database
- resolve historical URL-shaped event identity through the migration-only Python/TypeScript upcast contract
- persist tenant ownership, stable `job_id`, and the upcast version directly in the durable event log
- preserve exact event IDs, ordering, idempotency keys, entity references, payload facts, indexes, foreign keys, and the pre-upgrade AUTOINCREMENT high-water
- prove failed migrations leave the v29 table and version untouched and that the pre-upgrade backup reopens through the prior schema

## Why

The legacy event table used posting URLs as durable job identity. The one-time local upgrade must replace that shape before the new runtime starts. No old and new runtime coexistence is supported: the old process is stopped, the databases are backed up, the migration runs once, and only schema v30 is accepted afterward.

## Validation

- Python event-upcast, event-migration, and backup tests: 35 passed
- Full API suite including the TypeScript upcast parity fixture: 62 files, 748 tests passed
- API TypeScript check: passed
- Focused Ruff: passed
- `git diff --check`: passed

## Stack

- Base: #566
- Runtime event writers and readers are activated in the next small stacked slices.
- These PRs are review slices of one cumulative local upgrade, not independently deployed application versions.

### DIFF
--- a/apps/api/src/event-identity-upcast.ts
+++ b/apps/api/src/event-identity-upcast.ts
@@ -29,6 +29,7 @@ const ROOT_PRIMARY_FIELDS = [
   "survivingJobId",
   "surviving_job_id",
 ] as const;
+const NON_JOB_SCOPE_REFERENCES = new Set(["pipeline"]);
 
 export class EventIdentityUpcastError extends Error {
   readonly code: string;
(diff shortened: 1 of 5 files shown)


ebarti:

PR #570, opened 2026-07-30
feat(storage): admit exact shipped v6 schema

ebarti:
## What changed

- freezes the exact fresh shipped-v6 schema as a synthetic, non-sensitive fixture
- admits the exact v1.3→v2.0.8 local upgrade-history schema as a second named core manifest
- admits the exact supported v6 runtime variants that contain the six application attestation and referral-preference profile columns
- admits only enumerated raw DDL variants for optional lifecycle, heartbeat, recovery, review, and feedback tables
- admits both TypeScript- and Python-owned v2.0.8 application outcomes variants, email evidence, outcome suggestions, and their owned indexes
- subtracts only fully validated auxiliary rows, so reserved-name collisions cannot hide triggers or other objects
- ignores only exact inert SQLite-owned objects; unknown `sqlite_%` triggers remain visible and are rejected
- performs schema admission with zero target writes

## Why

The one-shot v7 cutover must fail before mutation for unknown database shapes while still accepting real local workspaces produced by supported release histories. The profile variants are real supported v6 shapes, not a compatibility fallback.

## Validation

- `pytest -q workers/automation/tests/test_v6_to_v7_preflight.py` — 15 passed
- cumulative focused preflight/plan/copy/candidate/execute regression suite on the rebased top — 47 passed
- focused Ruff — passed
- `git diff --check` — passed

## Stack

Depends on #569. This PR performs no migration writes; migration inventory and candidate copying continue in #577–#581.

### DIFF
--- a/workers/automation/src/jobctrl/infrastructure/migrations/v6_to_v7_preflight.py
+++ b/workers/automation/src/jobctrl/infrastructure/migrations/v6_to_v7_preflight.py
@@ -23,6 +23,21 @@ _V6_CORE_MANIFESTS = frozenset(
             87,
             "ac7d39828e3ad9e0dd1983ed21180686b81fab9d62c7a75409e903ae44256e89",
         ),
+        # v6's profile runtime added these six application attestation and
+        # preference columns without changing user_version.  This fingerprint
+        # admits only that exact shipped ALTER TABLE sequence.
+        (
+            161,
+            86,
+            "17526494e8a98d0a2a166502d605f063d5c9dc57397bc1109d7d8847c8e280b5",
+        ),
+        # The same runtime ALTER sequence can run after the separately
+        # admitted v1.3-to-v2.0.8 raw-schema history.
+        (
+            162,
+            87,
+            "51da7982e874bc3883a6765ad27ef33b4aa82237da96aceb9feb8c1b03ec2252",
+        ),
     }
 )
 
--- a/workers/automation/src/jobctrl/infrastructure/migrations/v6_to_v7_preflight.py
+++ b/workers/automation/src/jobctrl/infrastructure/migrations/v6_to_v7_preflight.py
@@ -83,6 +83,69 @@ _V6_AUXILIARY_DDL = (
             updated_at                  TEXT NOT NULL,
             PRIMARY KEY (tenant_id, discover_workflow_id, discover_run_id)
         )""",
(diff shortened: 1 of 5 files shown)


workers/automation/src/jobctrl/infrastructure/migrations/v6_to_v7_preflight.py
```diff
@@ -0,0 +1,440 @@
...
+            key_digest                  TEXT NOT NULL,
+            last_error_code             TEXT,
+            updated_at                  TEXT NOT NULL,
+            PRIMARY KEY (tenant_id, discover_workflow_id, discover_run_id)
+        )""",
+    # v2.0.8's TypeScript feedback owner creates these exact tables lazily.
+    # `application_outcomes` may additionally have the later ALTER TABLE
+    # column listed below; the evidence and suggestion tables have one shipped
+    # raw definition each.
+    """CREATE TABLE application_outcomes (
```
ebarti, line 105: PR Reviewer - Registry Re-review

Severity: High
Gate: Fail
Evidence: This allowlist at workers/automation/src/jobctrl/infrastructure/migrations/v6_to_v7_preflight.py:90-148 freezes only the TypeScript owners persisted raw DDL. The supported Python owner in workers/automation/src/jobctrl/infrastructure/gmail/feedback.py:247-332 creates application_outcomes, application_email_evidence, application_outcome_suggestions, and their indexes with different persisted indentation; scan_gmail_feedback calls it at line 131, and apps/api/src/server.ts:1105-1116 directly exposes that worker through POST /v1/outcomes/gmail/scan.
Reproduction: Reproduced with actual head Git blobs in memory: executed the Python owners exact executescript constant over shipped_v6_schema.sql, then called assert_v6_migration_preflight; it raised V6MigrationPreflightError: database has an unsupported v6 optional durable-table variant. All seven owned table/index sqlite_master rows differed from the admitted TypeScript formatting.

A valid v2.0.8 workspace whose feedback tables were first created by the Python Gmail scanner cannot pass the promised supported-v6 upgrade preflight. A later TypeScript call does not repair the raw table/index DDL because CREATE IF NOT EXISTS preserves the Python-created sqlite_master text.

Required fix: Admit the exact Python-owner raw variants for outcomes both before and after the TypeScript interview_prep_generation ALTER, email evidence, suggestions, and their owned indexes. Add owner-grounded tests that reproduce both Python-first states while retaining exact drift, collision, and zero-write rejection.

PR #572, opened 2026-07-30
feat(storage): migrate event history to JobId

ebarti:
> Superseded by the source-immutable exact-v7 candidate migration stack beginning at #577. This PR rebuilt the event table inside the admitted v6 database and therefore belongs to the discarded in-place design.

## Status

Closed without merge. Event-history rewriting will be reintroduced as a candidate-only structured-copy slice; no descendant READY PR depends on this branch.

### DIFF
--- a/workers/automation/src/jobctrl/infrastructure/migrations/v6_to_v7_events.py
+++ b/workers/automation/src/jobctrl/infrastructure/migrations/v6_to_v7_events.py
@@ -1,0 +1,309 @@
+"""Explicit v6-to-v7 reconstruction of append-only job event identity."""
+
+from __future__ import annotations
+
+import json
+import sqlite3
+from typing import Any
+
+from jobctrl.infrastructure.migrations.identity_upcast import (
+    EVENT_IDENTITY_UPCAST_VERSION,
+    upcast_v6_event_identity,
+)
+from jobctrl.infrastructure.migrations.v6_to_v7_support import (
+    has_tenant_job_foreign_key,
+    table_columns,
+)
+
+_EVENT_TABLES = ("job_events",)
+_EVENT_COLUMNS = (
+    "event_id",
+    "tenant_id",
+    "job_id",
+    "identity_version",
+    "stage",
+    "event_type",
+    "level",
+    "message",
+    "occurred_at",
+    "payload_json",
+    "entity_kind",
+    "entity_ref",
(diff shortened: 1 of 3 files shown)


workers/automation/src/jobctrl/infrastructure/migrations/v6_to_v7_events.py
```diff
@@ -0,0 +1,309 @@
...
+        """
+        CREATE UNIQUE INDEX idx_job_events_idempotency_key
+        ON job_events(idempotency_key)
+        WHERE idempotency_key IS NOT NULL
+        """
+    )
+    conn.execute(
+        """
+        CREATE INDEX idx_job_events_job_time
+        ON job_events(tenant_id, job_id, occurred_at DESC, event_id DESC)
```
ebarti: PR Reviewer - review_pr572_events

Severity: High
Gate: Fail
Evidence: `workers/automation/src/jobctrl/infrastructure/migrations/v6_to_v7_events.py:126-142` creates four indexes with SQL text that differs from the canonical definitions in `schema_v7.sql`. Executing the shipped-v6 fixture through `transform_v6_root_identity` and `transform_v6_job_events`, then comparing `SELECT type,name,tbl_name,sql FROM sqlite_master WHERE tbl_name='job_events' ORDER BY type,name` against a fresh `schema_v7.sql` database returns `equal=False` for `idx_job_events_job_time`, `idx_job_events_tenant_eid`, `idx_job_events_stage_time`, and `idx_job_events_entity`. The base contract fingerprints the raw ordered `sqlite_master` tuple, so the migrated database cannot satisfy the exact v7 manifest.
Reproduction: Reproduced with two in-memory SQLite databases: one loaded from `shipped_v6_schema.sql` and transformed by this PR, the other loaded from `schema_v7.sql`; the four stored index SQL strings differ in parenthesized-column formatting.

The phase requires the shipped v6 database to become the exact v7 schema, but this transform creates only semantically equivalent indexes and `verify_v7_job_events` never compares their canonical SQL. Final manifest verification will reject an otherwise successful migration.

Required fix: Create these indexes with the exact canonical DDL from `schema_v7.sql` (ideally from one shared canonical source), and add a focused assertion that the transformed `job_events` table plus indexes match the canonical v7 `sqlite_master` rows/raw manifest.

PR #659, opened 2026-08-01
test(projections): align exact v7 runtime fixtures

ebarti:
## Summary
- seed audit and projection fixtures with exact tenant-scoped JobIds
- remove retired-column fixture writes and project only registered artifacts
- preserve projection parity behavior while making the fixtures v7-native

## Validation
- fixed-environment Python cumulative suite: 3,110 passed, 2 skipped
- Ruff on touched fixture files
- `git diff --check`


ebarti:
The projection fixture corrections now live in their mapped owner PRs (#635 and #648), so this child has no remaining diff. Closing the empty stack position.

PR #710, opened 2026-08-02
fix: restore JobId v7 continuation onto main

ebarti:
## Summary

- restores the 27 commits corresponding to #680-#701 and #703-#707 onto current `main`
- preserves the exact cumulative tree from the orphaned stack merge tip `dec43760f`
- replaces #707 because GitHub does not allow a merged pull request to be reopened

## Why

After #679 merged, #680 still targeted `feat/job-id-v7-learning-recommendation-writer`. Merging the continuation stack therefore wrote its squash commits to that already-merged branch rather than to `main`. This pull request replays those exact 27 commits onto current `main` without changing their cumulative content.

## Validation

- `gh stack view --json`: trunk `main`, base `7000f5672`, `needsRebase: false`
- `git rev-list --left-right --count origin/main...HEAD`: `0 27`
- cumulative tree matches `dec43760f` with an empty `git diff`
- the cumulative TypeScript and Python 3.11-3.13 gates previously passed on #707

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -93,6 +93,11 @@ and explain any intentional stacked deferral in the PR body.
 - **Subagent spawning:** Use subagents only for genuinely independent work or
   for the review/QA gates required by the validation tiers below. Do not add
   coordination overhead to work that is faster to complete directly.
+  Within one task, create at most one agent of each required type/role and
+  reuse that agent via follow-up turns for related work, fixes, and reruns.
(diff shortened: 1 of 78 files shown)


ebarti:
Superseded by the 27 individual recovered PRs in new GitHub Stack #738: bottom PR #711 targets main and top PR #737 preserves the former #707 layer. The cumulative recovery branch remains available as a recovery reference; no implementation content was discarded.

PR #750, opened 2026-08-06
fix(pipeline): make cancellation requests auditable

ebarti:
## Summary

-

## Validation

-

## Checklist

- [ ] I ran the relevant local checks and listed them above. Heavy CI runs automatically for ordinary same-repository pull requests and cumulative stack heads, but not public forks; maintainers run manual workflows or local validation for fork contributions after review.
- [ ] I did not include secrets, private profile data, resumes, generated application materials, raw logs, browser profiles, SQLite databases, or local paths.


workers/automation/src/jobctrl/infrastructure/projections/projection_builder.py
```diff
@@ -111,6 +111,7 @@ class _ProvenanceProjection:
 # event maps to a terminal status in the 12-state ``WORKFLOW_RUN_STATUSES``.
 WORKFLOW_EVENT_TYPES: tuple[str, ...] = (
     "WorkflowStarted",
+    "WorkflowCancellationRequested",
```
ebarti: PR Reviewer - pr-750-reviewer

Severity: High
Gate: Fail
Evidence: workers/automation/src/jobctrl/infrastructure/projections/projection_builder.py:114 with `_project_workflow_run` (initial fold state `status="in_progress"`, `started_at=None`, `input_summary={}`) and `SqliteProjectionStore.upsert_workflow_run` (full-row `ON CONFLICT ... DO UPDATE` overwrite, sqlite_projection_store.py:960-981). The read-model fold docstring in apps/api/src/read-model.ts explicitly acknowledges the affected population: "Runs with no `Workflow*` events (legacy seeds) fall back to the stored row."
Reproduction: Reproduced with a script against this branch: seeded `workflow_run_projections` rows with no `Workflow*` job_events, then called `record_workflow_cancellation_requested` (which triggers `ProjectionBuilder.refresh()`). Output:
```
BEFORE live    : {'status': 'in_progress', 'started_at': '2026-08-01T10:00:00Z', 'finished_at': None, 'input_summary': '{"jobId": "job-1"}'}
BEFORE canceled: {'status': 'canceled', 'started_at': '2026-08-01T09:00:00Z', 'finished_at': '2026-08-01T09:30:00Z', 'input_summary': '{"jobId": "job-2"}'}
AFTER  live    : {'status': 'in_progress', 'started_at': None, 'finished_at': None, 'input_summary': '{}'}
AFTER  canceled: {'status': 'in_progress', 'started_at': None, 'finished_at': None, 'input_summary': '{}'}
```

Adding `WorkflowCancellationRequested` to `WORKFLOW_EVENT_TYPES` feeds pure audit facts into `_rebuild_workflow_runs`. For a workflow whose `job_events` contain only the cancellation fact (the legacy-seed rows the TS fold explicitly preserves, or runs whose events were pruned), the fold never sees a `WorkflowStarted` or terminal event, so it returns the seed state and `upsert_workflow_run` overwrites the stored row: a canceled run flips to `in_progress` and `started_at`/`finished_at`/`input_summary` are destroyed. Because those rows have no backing events, the overwrite is unrecoverable. Note this is exactly the population `workflow_runs_missing_cancellation_audit` targets (`status = 'canceled'`), so the intended backfill consumer would corrupt every legacy canceled run it touches. Recording an audit fact must never mutate run state.

Required fix: `_project_workflow_run` must not materialize state from audit-only event groups - e.g. track whether the group contained any state-bearing event (`WorkflowStarted` or a terminal) and return `None` otherwise, keeping `WorkflowCancellationRequested` timeline-only. Add a regression test that records the fact for a projection row with no `Workflow*` events and asserts the row is unchanged.

ebarti (commented):
Reviewed adf0aec (fix/cancellation-audit-terminal-state vs main).

The event-schema plumbing (domain-types union, SSE parse, invalidation router, drawer rendering) is consistent across TS and Python, and the RPC intent-recording path is correctly ordered (record only after Temporal accepts delivery) and tested. Verified locally: `pytest tests/test_rpc_cancel_run.py` (3 passed), the two touched web suites (107 passed), and `api:check`/`web:check` typechecks (clean).

Findings: 2 High, 1 Medium, 1 Low.
- High (reproduced): recording a cancellation-audit fact for a run with no other `Workflow*` events full-row-overwrites the stored projection - a canceled legacy run flips to `in_progress` and loses `started_at`/`finished_at`/`input_summary` unrecoverably. Same defect exists independently in the TS read-model fold.
- Medium: the entire Temporal-history recovery half (history reader, source classifier, missing-audit query) is dead code with no tests, so externally initiated cancels are still not auditable despite the PR title.
- Low: an audit-write failure after successful cancel delivery surfaces as a failed cancel to the client.

apps/api/src/read-model.ts
```diff
@@ -5721,6 +5721,7 @@ function activityRowToSummary(row: Record<string, unknown>): ActivityEventSummar
  */
 const WORKFLOW_LIFECYCLE_EVENT_TYPES = [
   "WorkflowStarted",
+  "WorkflowCancellationRequested",
```
ebarti, line 5729: PR Reviewer - pr-750-reviewer

Severity: High
Gate: Fail
Evidence: apps/api/src/read-model.ts:5724; `foldWorkflowRunEvents` ignores the event (not `WorkflowStarted`, not in `WORKFLOW_TERMINAL_STATUS_BY_EVENT`) so a cancellation-only stream returns the seed fold `{status: "in_progress", startedAt: null, finishedAt: null}`, and `applyWorkflowLifecycleFold` (read-model.ts:5899-5914) unconditionally overwrites the stored row whenever a fold exists.
Reproduction: Reasoned from code; the equivalent Python-side overwrite was reproduced (see the projection_builder.py comment).

Read-side half of the same defect: before this change, `loadWorkflowRunLifecycleFolds` only produced a fold for workflows with at least one state-bearing `Workflow*` event, which is what made the documented fallback ("Runs with no `Workflow*` events (legacy seeds) fall back to the stored row") hold. Including `WorkflowCancellationRequested` in `WORKFLOW_LIFECYCLE_EVENT_TYPES` means a lone audit fact now creates a degenerate fold that serves a stored canceled/succeeded run as `in_progress` with null `startedAt`/`finishedAt` in both the runs list and the run detail. The event contributes nothing to the fold itself, so its inclusion here has no benefit, only this hazard.

Required fix: keep the fold sources state-bearing only (drop this type from `WORKFLOW_LIFECYCLE_EVENT_TYPES`), or have `loadWorkflowRunLifecycleFolds`/`foldWorkflowRunEvents` skip folds that never left the initial phase so the stored row is preserved.

workers/automation/src/jobctrl/infrastructure/temporal/cancellation_audit.py
```diff
@@ -0,0 +1,140 @@
...
+    if normalized.startswith("temporal-cli:"):
+        return "temporal_cli"
+    if normalized.startswith("temporal-web:"):
+        return "temporal_web"
+    if normalized:
+        return "temporal_external"
+    return "temporal_unknown"
+
+
+async def cancellation_request_from_history(
```
ebarti, line 57: PR Reviewer - pr-750-reviewer

Severity: Medium
Gate: Fail
Evidence: `grep -rn "cancellation_request_from_history\|workflow_runs_missing_cancellation_audit" workers/` shows no callers outside the defining modules; no test imports them either. The only wired producer is `make_cancel_run` (evidence_kind `request_intent`).
Reproduction: Reasoned from code; verified by grep on this branch.

Half of this PR's audit capability is unreachable: `cancellation_request_from_history`, `cancellation_source`, `_event_timestamp`, and `SqliteProjectionStore.workflow_runs_missing_cancellation_audit` have no callers and no unit tests, so the `temporal_history`/`recovered_temporal_history` evidence kinds can never be produced. Externally initiated cancels (Temporal CLI/Web) - the exact scenario the new WorkflowRunDrawer test showcases ("temporal-cli:tester@local via temporal_cli") - have no producing product path; that test only exercises a hand-built API mock. (Minor related drift: the fixture sets `status: "in_progress"` on the cancellation timeline entry, but the real Python-built timeline writes `status: null`, which renders as `data-event-status="unknown"`.) The untested history-parsing logic (`WhichOneof`, identity/cause extraction, protobuf timestamp conversion) is exactly the kind of code that breaks silently when finally wired.

Required fix: either wire the reconciler/backfill consumer in this PR with unit tests for `cancellation_source` and `cancellation_request_from_history` (mocked history events), or move the unused helpers and the `workflow_runs_missing_cancellation_audit` query to the PR that consumes them so this PR ships only reachable, tested behavior.

workers/automation/src/jobctrl/infrastructure/rpc/handlers.py
```diff
@@ -841,9 +841,32 @@ def make_cancel_run(canceler: WorkflowCanceler):
...
+        from jobctrl.infrastructure.temporal.cancellation_audit import (
+            record_workflow_cancellation_requested,
+        )
+        from jobctrl.state import utc_now
+
+        # A local intent becomes a cancellation request only after Temporal
+        # accepts delivery. The immutable history requester is reconciled as a
+        # separate evidence fact, so neither source can suppress the other.
         asyncio.run(canceler(run_id))
+        record_workflow_cancellation_requested(
```
ebarti: PR Reviewer - pr-750-reviewer

Severity: Low
Gate: Pass
Evidence: workers/automation/src/jobctrl/infrastructure/rpc/handlers.py:860; `record_workflow_cancellation_requested` executes SQL (projection lookup, insert, commit, projection refresh) after `asyncio.run(canceler(run_id))` has already succeeded. `JsonRpcServer.dispatch` converts handler exceptions into JSON-RPC errors (see test_cancel_run_delivery_failure_does_not_record_request_intent).
Reproduction: Reasoned from code; not run.

The delivery-before-record ordering is right and tested, but the converse failure is unhandled: if the audit write raises after Temporal accepted the cancel (locked DB, missing projection table on a fresh profile), the RPC returns an error and the UI reports a failed cancel for a run that is actually canceling - the audit feature degrades the primary action's reported outcome.

Required fix: contain audit-write failures (log at error level and still return `{"runId": ..., "status": "canceling"}`), or explicitly document that a cancel without a recorded audit fact must surface as an error.

workers/automation/src/jobctrl/infrastructure/projections/projection_builder.py
ebarti: Fixed in 010f2cc5a. `_project_workflow_run` now returns `None` for audit-only event groups (no `WorkflowStarted`/terminal event), so the stored row is preserved verbatim; `WORKFLOW_EVENT_TYPES` is split into `WORKFLOW_STATE_EVENT_TYPES` + `WORKFLOW_AUDIT_EVENT_TYPES`, and the backfill-pending detector counts state-bearing events only. Regression fixture `test_cancellation_audit_only_group_preserves_stored_rows` reproduces your exact repro (canceled legacy row + late audit fact) and fails on the unfixed code (canceled -> in_progress); a second test pins that the fact stays timeline-only for real runs.

apps/api/src/read-model.ts
ebarti: Fixed in 010f2cc5a. `foldWorkflowRunEvents` now returns `null` when the fold never left the initial phase (no state-bearing event), and `loadWorkflowRunLifecycleFolds` skips null folds, so a cancellation-only stream serves the stored row — same invariant as the Python fix. The type stays in `WORKFLOW_LIFECYCLE_EVENT_TYPES` (commented as fold-inert) so the list keeps mirroring the canonical stream. Twin regression fixture in server.test.ts ("keeps the stored canceled run when its only lifecycle stream entry is a cancellation-audit fact") verified red on the unfixed read-model and green after. No domain-types parity fixture covers workflow-run rows, so both runtimes are pinned via their own suites.

workers/automation/src/jobctrl/infrastructure/temporal/cancellation_audit.py
ebarti: Fixed in 010f2cc5a — the recovery half is now wired and tested. (1) `_reconcile_one_workflow_run` records the immutable requester from the exact observed execution when it closes a CANCELED run (evidence_kind `temporal_history`, best-effort, never settlement-blocking). (2) New `reconcile_cancellation_audit` sweep in the worker heartbeat loop consumes `workflow_runs_missing_cancellation_audit` and records `recovered_temporal_history` facts for already-canceled runs (external CLI/Web cancels, finalize-closed cooperative cancels, legacy rows); NOT_FOUND histories are skipped quietly. New `test_cancellation_audit.py` unit-tests `cancellation_source` + `cancellation_request_from_history` against mocked history events (parsing found a real bug: a blank identity classified as temporal_external; now temporal_unknown) plus integration tests proving a canceled-without-audit run gains its fact WITHOUT the row being corrupted (depends on the projection fixes, landed in the same commit). The timely path is covered in test_worker_reconciler.py. Also fixed the noted drawer-fixture drift: the timeline entry now carries `status: null` like the real Python-built timeline.

workers/automation/src/jobctrl/infrastructure/rpc/handlers.py
ebarti: Fixed in 010f2cc5a. Once `asyncio.run(canceler(run_id))` returns, the RPC always reports `{runId, status: "canceling"}`; a failing `record_workflow_cancellation_requested` degrades to `logger.warning(..., exc_info=True)`. The requester remains recoverable via the newly wired temporal_history/recovered_temporal_history reconciliation. New test `test_cancel_run_audit_write_failure_still_reports_canceling` monkeypatches the audit writer to raise and asserts the success envelope; the existing delivery-failure ordering test is unchanged.

ebarti:
Addressed all four findings from review 4886180416 in 010f2cc5a (plain commit on top of adf0aec; no rebase).

- **High (Python projection overwrite, reproduced):** audit-only event groups no longer materialise a fold — `_project_workflow_run` returns `None` when the group has no state-bearing `Workflow*` event, so recording a cancellation fact for a legacy canceled run preserves the stored status/started_at/finished_at/input_summary. The backfill-pending counter now counts state-bearing events only. Regression fixture verified red on the unfixed code (canceled -> in_progress) and green after.
- **High (TS read-model, same defect):** `foldWorkflowRunEvents` returns `null` for streams that never left the initial phase and `loadWorkflowRunLifecycleFolds` skips them, restoring the documented legacy-seed fallback with the same invariant as Python. Twin server.test.ts fixture, also verified red-on-unfixed. No domain-types parity JSON covers workflow-run rows, so each runtime pins the case in its own suite.
- **Medium (dead recovery half):** wired both consumers — the describe reconciler records `temporal_history` when closing a CANCELED execution, and a new heartbeat `reconcile_cancellation_audit` sweep backfills `recovered_temporal_history` via `workflow_runs_missing_cancellation_audit`. Unit tests for `cancellation_source` / `cancellation_request_from_history` (which surfaced and fixed a blank-identity misclassification) plus integration tests proving a canceled-without-audit run gains its fact without row corruption. Drawer fixture drift (`status: null`) also fixed.
- **Low (audit failure fails the cancel):** `cancel_run` now returns `canceling` once Temporal accepted delivery and degrades an audit-write failure to a logged warning, with a test.

Validation: `pytest test_cancellation_audit.py test_rpc_cancel_run.py test_worker_reconciler.py test_workflow_runs_projection_from_events.py` (35 passed) plus the adjacent projection suites (33 passed) and heartbeat/starter suites (13 passed); `corepack pnpm --filter @jobctrl/api check` clean and `--filter @jobctrl/api test` 741/741; touched web drawer suite 6/6.

PR #751, opened 2026-08-06
fix(enrichment): reconcile durable workflow ownership

ebarti:
## Summary

- bind selected Enrich cohorts to their exact Temporal execution and fence stale activity writers
- distinguish cancellation from timeout or worker shutdown, then reconcile abandoned ownership after restart
- rotate abandoned blocking executors so redelivery cannot silently consume worker capacity
- keep Tailor retry exhaustion in the following Materials phase so this Enrichment PR is independently importable without a partial downstream contract

## Why

Canceled or interrupted executions could leave canonical rows owned by dead workflows while the UI continued to present work as pending. The original stack split also placed Tailor consumers and tests before their owning implementation; this PR now contains only the Enrichment ownership boundary it is meant to introduce.

## Validation

- `uv --project workers/automation run --frozen --extra dev pytest workers/automation/tests/test_p1b_error_inversion.py -q` — 15 passed
- `uv --project workers/automation run --frozen --extra dev pytest workers/automation/tests --collect-only -q` — 3,319 tests collected with no import errors
- cumulative hosted checks are green on stack head #763
- lower stack layers intentionally skip the heavy hosted Python matrix; cumulative heads own that product gate

## Checklist


workers/automation/tests/test_jsonrpc_handlers.py
```diff
@@ -2258,6 +2236,100 @@ def test_selected_tailor_activity_accepts_committed_acknowledgement_replay(monke
     assert result["stages"][0]["approvedJobIds"] == [job_id]
 
 
+def test_selected_tailor_activity_uses_bounded_requested_workers(monkeypatch) -> None:
```
ebarti: PR Reviewer - pr-751-reviewer

Severity: High
Gate: Fail
Evidence: On this PR's head (5b4156f), `uv --project workers/automation run --frozen --extra dev pytest workers/automation/tests/test_jsonrpc_handlers.py -q` fails 3 tests added by this PR: `test_selected_tailor_activity_uses_bounded_requested_workers` and `test_selected_cover_activity_uses_bounded_requested_workers` fail with `AttributeError: module 'jobctrl.materials.activities' has no attribute 'ThreadPoolExecutor'`, and `test_selected_tailor_activity_keeps_item_failures_partial` fails with `assert 'failed' == 'partial'`. None of these tests exist on the base branch; the implementation they assert lives only in the next stack layer (#752, `fix/material-retry-chaos-hardening`, which adds the executor and the partial-status contract to `materials/activities.py`).
Reproduction: Reproduced with the pytest command above; 3 failed, 112 passed.

Commit 5b4156f ("fix(stack): keep tailor exhaustion in material phase") moved the Tailor/Cover consumer implementation out of this PR, but left these three consumer tests behind. That re-creates exactly the stack-split problem the PR body says it fixed ("the original stack split also placed Tailor consumers and tests before their owning implementation"): merged or imported alone, this PR leaves the suite red. The hosted Python matrix is skipped on this layer, so nothing upstream catches it.

Required fix: Move these three tests to #752 alongside the implementation they exercise (or land the minimal implementation here). After the change, `pytest workers/automation/tests/test_jsonrpc_handlers.py` must pass on this branch standalone.

workers/automation/src/jobctrl/cli.py
```diff
@@ -2677,17 +2684,144 @@ async def _reconcile_workflow_runs(temporal_client: Any, *, tenant_id: str | Non
...
+def _reconcile_canceled_enrichment_cohorts(conn, *, tenant_id: str) -> int:
+    """Terminalize unfinished Enrich rows still owned by a canceled run."""
+
+    from collections import defaultdict
+
+    from jobctrl.domain.identifiers import JobId
+    from jobctrl.domain.tenant import TenantId
+    from jobctrl.enrichment.detail import cancel_enrichment_cohort
+
+    canceled_execution_rows = conn.execute(
```
ebarti, line 2741: PR Reviewer - pr-751-reviewer

Severity: Medium
Gate: Pass
Evidence: `_reconcile_canceled_enrichment_cohorts` runs inside `_reconcile_workflow_runs`, which the worker heartbeat loop invokes every 15s (`_worker_heartbeat_loop`, `interval_seconds: float = 15.0`). This query selects every canceled run that ever claimed an enrichment lease — with no cutoff or settled marker — and the loop below calls `cancel_enrichment_cohort` for each, which re-claims the phase-3 cancellation lease (`claim_enrichment_execution_lease_for_run` → `BEGIN IMMEDIATE` + commit) and opens a second `BEGIN IMMEDIATE` transaction even when the cohort has zero unfinished rows.
Reproduction: Reasoned from code; the re-claim is idempotent (same idempotency key, `ON CONFLICT DO NOTHING` in `record_job_event`, and equal phase/attempt with the same owner passes the staleness check), so there is no unbounded event growth — but the write-lock churn is real.

Every canceled pipeline run in history permanently costs two write-lock acquisitions per 15-second heartbeat on the single-writer SQLite database, forever, contending with live enrichment/scoring writers. The work grows linearly with the lifetime count of canceled runs.

Required fix: Stop reprocessing settled cohorts — e.g., exclude runs whose lease lane already contains a phase-3 `cancellation:` claim AND have no remaining `pending/queued/running` enrich rows (an `AND NOT EXISTS` against the phase-3 lease event, or a durable `cohort_settled` marker recorded after the first successful pass).

workers/automation/src/jobctrl/enrichment/activities.py
```diff
@@ -52,14 +132,27 @@ async def enrich_activity(payload: EnrichActivityInput) -> EnrichActivityOutput:
         expected_db_path=payload.expected_db_path,
     )
 
-    cancel_event = threading.Event()
+    cancel_event = _ActivityCancellationEvent()
+    try:
+        info = activity.info()
+        activity_attempt = info.attempt
+        activity_owner_token = f"{info.activity_id}:{info.activity_run_id}:{info.attempt}"
```
ebarti: PR Reviewer - pr-751-reviewer

Severity: Low
Gate: Pass
Evidence: This line reads `info.activity_run_id`, and `_ActivityCancellationEvent.request_stop` calls `activity.cancellation_details()`. Both APIs are recent additions to the temporalio SDK (the locked version 1.26.0 has them; `Info.activity_run_id` and `cancellation_details()` do not exist in the `>=1.6` floor that `workers/automation/pyproject.toml` still declares). On an older-but-floor-compliant SDK, `info.activity_run_id` raises `AttributeError`, which the surrounding `except RuntimeError` does not catch — every workflow-driven enrich would crash.
Reproduction: Not run: uv.lock pins temporalio 1.26.0, so a `--frozen` install is safe; only a non-frozen resolve honoring the floor could produce the broken combination.

Note also that `Info.activity_run_id` is documented as "None for workflow activities", so the owner token is effectively `{activity_id}:None:{attempt}` in production — unique within the per-run lease lane, so fencing still works, but the middle component adds no entropy.

Required fix: Bump the `temporalio` floor in `workers/automation/pyproject.toml` to the minimum version that provides `activity.cancellation_details()` and `Info.activity_run_id` (anything `>=1.26` matches the lock); optionally drop the always-None `activity_run_id` component from the token.

workers/automation/src/jobctrl/infrastructure/temporal/run_in_activity.py
```diff
@@ -18,26 +18,80 @@
...
+    with _ACTIVITY_EXECUTOR_LOCK:
+        if _ACTIVITY_EXECUTOR is not abandoned_executor:
+            return False
+        max_workers = int(getattr(abandoned_executor, "_max_workers", 1))
+        replacement = ThreadPoolExecutor(
+            max_workers=max_workers,
+            thread_name_prefix="jobctrl-blocking-activity-recovery",
+        )
+        _ACTIVITY_EXECUTOR = replacement
+        _RETIRED_ACTIVITY_EXECUTORS.append(abandoned_executor)
```
ebarti: PR Reviewer - pr-751-reviewer

Severity: Low
Gate: Pass
Evidence: `_rotate_abandoned_activity_executor` appends the abandoned executor here and the list is only cleared in `shutdown_activity_executors()` at worker shutdown. Rotation triggers on every cancellation where the task has not finished at the instant of cancel (before the grace window), so a long-lived worker accumulates one retired `ThreadPoolExecutor` object per cancelled/timed-out blocking activity, including executors whose threads wound down cleanly seconds later. Separately, `int(getattr(abandoned_executor, "_max_workers", 1))` falls back to 1 if the private attribute ever disappears, silently collapsing blocking-activity concurrency to a single thread.
Reproduction: Reasoned from code; each drained retired executor is small, so this is a slow leak rather than a functional failure.

Required fix: Prune retired executors whose threads have exited (e.g., drop entries where `executor._threads` is empty on each rotation), or bound the list; and source the replacement's `max_workers` from the worker's own `activity_executor_max_workers(...)` computation instead of a private-attribute fallback of 1.

ebarti (commented):
PR review by pr-751-reviewer (automated review agent).

**Summary:** The Enrichment ownership/fencing design is sound and well-tested (cohort queue/release/settle, lease phase inversion, reconciler restart recovery, executor rotation, TS run-id vs handle separation all verified). However, this PR's own head fails 3 of the tests it adds: `test_jsonrpc_handlers.py` still contains tests for the Materials-phase implementation (`_run_selected_tailoring` partial status, bounded `requested_workers` executors) that the last commit deliberately deferred to #752 — contradicting the PR's stated goal of being independently importable. Hosted Python CI is skipped on this stack layer, so this is only visible locally.

**Verification run on head 5b4156f:**
- `pytest test_p1b_error_inversion.py test_enrichment_quarantine_gating.py test_worker_reconciler.py test_temporal_worker.py test_enrichment_detail_staleness.py test_activity_enrich.py test_enrichment_url_safety.py test_discovery_identity.py` — 108 passed
- `pytest test_discover_reliability.py` — 66 passed
- `pytest test_jsonrpc_handlers.py test_linkedin_apply_resolver.py test_linkedin_authenticated_enrichment_retry.py` — **3 failed**, 112 passed
- `vitest run` on the 4 changed apps/api test files — 277 passed

Severity: 1 High (failing tests shipped in this PR), 1 Medium (reconciler churn grows with canceled-run history), 2 Low.

workers/automation/tests/test_jsonrpc_handlers.py
ebarti: Fixed in 7af10dbe (this branch: the three Materials-phase consumer tests are removed, so pytest workers/automation/tests/test_jsonrpc_handlers.py passes standalone: 80 passed) and 44f1b072 on fix/material-retry-chaos-hardening (#752), whose own history now carries the tests beside the executor/partial-status implementation (fb1a55e7 cherry-picks the removal so the pair survives the upcoming restack; 3/3 pass on the #752 head).

workers/automation/src/jobctrl/cli.py
ebarti: Fixed in 29c32d8b. The canceled-run query now excludes runs whose lease lane already carries the terminal phase-3 cancellation claim, so settled cohorts stop costing two BEGIN IMMEDIATE transactions per heartbeat; runs with remaining pending/queued/running rows still resurface through the stage-row query, preserving idempotent crash recovery. Regression test test_reconcile_settled_canceled_cohort_performs_no_further_writes traces the second pass and asserts zero write statements (verified failing pre-fix).

workers/automation/src/jobctrl/enrichment/activities.py
ebarti: Fixed in 4b7853ac. temporalio floor raised to >=1.26 (uv.lock requires-dist specifier regenerated by uv lock; no resolution change since the lock already pins 1.26.0). The always-None activity_run_id token component is replaced with info.workflow_run_id, the scheduling execution real run id, so the owner token disambiguates executions writing into the same lease lane.

workers/automation/src/jobctrl/infrastructure/temporal/run_in_activity.py
ebarti: Fixed in 72c80d9e. Retired generations are now tracked as (executor, abandoned task) pairs and pruned on each rotation once the task has finished, so the list no longer grows until worker shutdown. Replacement sizing keeps _max_workers when present and otherwise falls back to the worker own activity_executor_max_workers(resolve_max_concurrent_activities().value) computation instead of 1. Unit tests cover both the prune and the fallback.

PR #752, opened 2026-08-06
fix(materials): bound retries and preserve recovery audit

ebarti:
## Summary

- separate inner generation retries from the durable Tailor attempt budget and terminalize the fifth durable failure as non-retryable exhaustion
- expose batch exhaustion to the Temporal boundary so a partially failed batch cannot be reported as successful
- preserve accepted materials and their audit history while retries, recovery, and replacement generation run
- use canonical JobId ownership and bounded material execution across worker restart and redelivery

## Why

Material generation could repeat an impossible repair, lose retry ownership across restarts, or hide exhaustion behind a retryable batch result. This phase owns the complete Tailor retry contract; the preceding Enrichment PR no longer carries partial consumers or tests from this layer.

## Validation

- P1b error inversion, production Tailor runtime, and pipeline workflow suites — 80 passed
- direct regressions cover inner retry exhaustion, fifth durable-attempt exhaustion, no re-entry after exhaustion, batch exhaustion count, non-retryable Temporal mapping, and preserved prior artifacts
- synchronized tree is byte-identical to previously green head `b53ac7f91`; only ancestry changed to include the corrected #751 base
- prior hosted Python 3.11/3.12/3.13, TypeScript, docs, and release checks passed; cumulative stack head #763 remains green

## Checklist

### DIFF
--- a/workers/automation/src/jobctrl/domain/materials/use_cases.py
+++ b/workers/automation/src/jobctrl/domain/materials/use_cases.py
@@ -31,7 +31,7 @@ import json
 import logging
 import re
 import uuid
-from collections.abc import Iterable, Mapping as MappingABC
+from collections.abc import Callable, Iterable, Mapping as MappingABC
 from contextlib import nullcontext
 from dataclasses import dataclass, field, replace
 from datetime import datetime, timezone
@@ -88,7 +88,12 @@ from jobctrl.domain.materials.fabrication_detector import (
     scan_prose_skill_fabrications,
     scan_resume_bullets,
 )
-from jobctrl.domain.materials.policy import LearnedTailoringRules, TailoringPolicy
+from jobctrl.domain.materials.policy import (
+    LearnedTailoringRules,
+    TailoringPolicy,
+    fingerprint_profile_snapshot,
+    fingerprint_value,
+)
 from jobctrl.domain.materials.provenance import BulletProvenance, BulletProvenanceSet
 from jobctrl.domain.materials.provenance_builder import (
     ProvenanceBindingError,
@@ -968,6 +973,7 @@ def _safe_candidate_summary(record: dict[str, Any]) -> dict[str, Any]:
         "generator": record.get("model"),
         "status": record.get("status"),
         "schema_version": record.get("schema_version"),
+        "prompt_fingerprint": record.get("prompt_fingerprint"),
         "validation": record.get("validator"),
         "judge": record.get("judge"),
         "fabrication_gate": record.get("fabrication_gate"),
@@ -976,6 +982,59 @@ def _safe_candidate_summary(record: dict[str, Any]) -> dict[str, Any]:
         "bullet_limit_overflows": record.get("bullet_limit_overflows") or [],
         "summary": _candidate_payload_summary(payload),
     }
+
+
+def _with_tailoring_attempt_audit(
+    materials: MaterialsSet,
+    *,
+    report: dict[str, Any],
+    recorded_at: str,
+    execution_id: str | None,
+    durable_attempt: int | None,
+) -> MaterialsSet:
+    """Append one durable execution's complete inner-attempt audit idempotently."""
+
+    metadata = dict(materials.metadata)
+    history: list[dict[str, Any]] = []
+    raw_history = metadata.get("tailoring_attempt_audits")
+    if isinstance(raw_history, list):
+        history = [dict(entry) for entry in raw_history if isinstance(entry, MappingABC)]
+    elif isinstance(metadata.get("tailoring_attempt_audit"), MappingABC):
+        # Preserve the singular pre-R25 record when an existing generation is
+        # first saved by the append-only writer.
+        history.append(
+            {
+                "audit_key": f"legacy:{materials.updated_at}",
+                "execution_id": None,
+                "durable_attempt": None,
+                "recorded_at": materials.updated_at,
+                "report": dict(metadata["tailoring_attempt_audit"]),
+            }
+        )
+
+    normalized_execution_id = str(execution_id or "local")
+    attempt_key = str(durable_attempt) if durable_attempt is not None else recorded_at
+    audit_key = f"{normalized_execution_id}:{attempt_key}"
+    if not any(str(entry.get("audit_key") or "") == audit_key for entry in history):
+        history.append(
+            {
+                "audit_key": audit_key,
+                "execution_id": execution_id,
+                "durable_attempt": durable_attempt,
+                "recorded_at": recorded_at,
+                "report": report,
+            }
+        )
+
+    return materials.with_metadata(
+        {
+            **metadata,
+            # Singular latest-record compatibility plus append-only history.
+            "tailoring_attempt_audit": report,
+            "tailoring_attempt_audits": history,
+        },
+        updated_at=recorded_at,
+    )
 
 
 def _voice_system_prompt() -> str:
@@ -1583,7 +1642,12 @@ class TailorResumeUseCase:
         suppress_existing_artifacts: bool = False,
         employer_analysis: EmployerAnalysis | None = None,
         requirement_fit_report: "RequirementFitReport | None" = None,
+        commit_guard: Callable[[], None] | None = None,
+        audit_execution_id: str | None = None,
+        durable_attempt: int | None = None,
     ) -> TailorOutcome:
+        if commit_guard is not None:
+            commit_guard()
         stable_job_id = job_id if job_id is not None else job.get("job_id")
         job_id = canonical_job_id(str(stable_job_id))
         # D-20: run/reuse the canonical employer analysis as the front-half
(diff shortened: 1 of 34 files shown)


workers/automation/src/jobctrl/pipeline/workflow.py
```diff
@@ -162,6 +168,7 @@ async def run(self, payload: JobPipelineWorkflowInput) -> JobPipelineWorkflowRes
                 await _check_spend(payload)
             result = await self._execute_stages(payload)
         except CancelledError:
+            await _cancel_owned_enrichment(payload)
```
ebarti, line 171: PR Reviewer - pr-752-reviewer

Severity: Medium
Gate: Pass
Evidence: `_cancel_owned_enrichment` schedules the new `cancel_enrichment_cohort` activity ahead of the pre-existing `emit_workflow_outcome` activity in the `except CancelledError:` path, with no `workflow.patched` gate. The two material-cancel insertions in this same PR are gated (`pipeline-material-cancellation-v1` at workflow.py:305, `preparation-material-cancellation-v1` in preparation/workflow.py), so the determinism hazard was recognized for the siblings but not here.
Reproduction: Reasoned from code. A pipeline canceled under the previous release has history: CancelledError -> ActivityTaskScheduled(record_workflow_outcome). If the worker restarts (e.g. the upgrade itself) inside that window before the run closes, the new worker replays the open history and now emits ScheduleActivityTask(cancel_enrichment_cohort) first -> nondeterminism error -> workflow task fails repeatedly and the canceled workflow never closes until terminated.

The window is small (seconds between the cancel handler and workflow close) but it coincides exactly with a deploy restart, and the failure mode is a permanently wedged cancellation.

Required fix: Gate this call the same way as the material-cancel path, e.g. `if not workflow.patched("pipeline-enrich-cancellation-v1"): return` at the top of `_cancel_owned_enrichment`.

workers/automation/src/jobctrl/pipeline/workflow.py
```diff
@@ -277,6 +298,30 @@ async def _recover_stage_state(
...
+        CancelPreparationStateInput(
+            tenant_id=payload.tenant_id,
+            workflow_id=workflow.info().run_id,
+            stage=stage,
+            job_ids=tuple(str(job_id) for job_id in _selected_job_ids(payload)),
+            expected_app_dir=payload.expected_app_dir,
+            expected_db_path=payload.expected_db_path,
+        ),
+        start_to_close_timeout=timedelta(seconds=30),
+        retry_policy=RetryPolicy(maximum_attempts=0),
```
ebarti: PR Reviewer - pr-752-reviewer

Severity: Low
Gate: Pass
Evidence: `RetryPolicy(maximum_attempts=0)` means unlimited attempts in temporalio (`common.py`: "If 0, the default, there is no maximum"). Same in preparation/workflow.py:243. In the same PR, `_cancel_owned_enrichment` bounds its cancel activity at `maximum_attempts=5`.
Reproduction: Reasoned from code. `assert_activity_runtime` failures are non-retryable (ApplicationError with non_retryable=True), but any persistently retryable failure of `cancel_preparation_state` (e.g. recurring sqlite I/O error) retries the 30s activity forever, so the canceled workflow never reaches its terminal state and the user's cancel hangs until a terminate.

I recognize `_recover_stage_state` already uses `maximum_attempts=0`, so this follows an existing pattern; but on the cancellation path liveness arguably matters more than at-all-costs terminalization (orphaned running rows are still recoverable later via `recover_preparation_state`/reconciliation), and the asymmetry with the enrichment cancel's `maximum_attempts=5` looks unintentional.

Required fix: Bound the attempts (e.g. `maximum_attempts=5` to match `_cancel_owned_enrichment`) here and in preparation/workflow.py `_cancel_material_stage_state`, or document why unlimited is intended for the material path but not the enrichment path.

workers/automation/src/jobctrl/domain/materials/use_cases.py
```diff
@@ -978,6 +984,59 @@ def _safe_candidate_summary(record: dict[str, Any]) -> dict[str, Any]:
...
+                "execution_id": None,
+                "durable_attempt": None,
+                "recorded_at": materials.updated_at,
+                "report": dict(metadata["tailoring_attempt_audit"]),
+            }
+        )
+
+    normalized_execution_id = str(execution_id or "local")
+    attempt_key = str(durable_attempt) if durable_attempt is not None else recorded_at
+    audit_key = f"{normalized_execution_id}:{attempt_key}"
```
ebarti: PR Reviewer - pr-752-reviewer

Severity: Low
Gate: Pass
Evidence: `audit_key = f"{normalized_execution_id}:{attempt_key}"` where CLI runs normalize execution_id to "local" and `attempt_key = str(durable_attempt)`. The append at line 1018-1027 is skipped when the key already exists.
Reproduction: Reasoned from code. Two reachable collisions drop the newest execution's report from the append-only history: (1) after `jobctrl retry tailor --reset-attempts`, the next CLI execution reuses durable_attempt=1 and collides with the pre-reset `local:1` entry; (2) a Temporal activity retry within the same run after a crash between the materials save and the stage-count increment reruns the same durable_attempt under the same run_id, so the fresh generation's report (new LLM calls, potentially different outcome) is discarded and only the crashed attempt's report stays in `tailoring_attempt_audits`. In both cases the singular `tailoring_attempt_audit` advances while the history silently diverges from it, which undercuts the audit-preservation goal of this PR (a history whose latest entry disagrees with the singular field is harder to trust than a missing one).

Required fix: Make the key unique per recorded execution (e.g. include `recorded_at` alongside execution_id/durable_attempt) or, if replay-idempotency for byte-identical reports is the goal, dedupe on report content rather than on attempt number.

ebarti (commented):
Reviewed a9476328 against base fix/enrich-recovery-ownership (5b4156f5). Verified locally: P1b error inversion, tailor/cover runtime, materials recovery/unit-of-work/repository/use-case, and pipeline+preparation workflow suites (253 passed), focused Ruff, git diff --check. The retry-budget separation, ownership fencing, and cancellation terminalization are coherent and well tested end to end. No Blocker/High findings. 1 Medium (replay-determinism gate missing on the new enrichment-cancel activity in the CancelledError path) and 2 Low (unbounded RetryPolicy on the cancel-cohort activity, audit-history key collision that can drop reports after --reset-attempts).

workers/automation/src/jobctrl/pipeline/workflow.py
ebarti: Fixed in 5e1d015cf: `_cancel_owned_enrichment` now returns early unless `workflow.patched("pipeline-enrich-cancellation-v1")`, mirroring the material-cancel patch convention, so pre-patch cancel histories replay without scheduling the new activity. Regression: `test_cancel_owned_enrichment_is_patch_gated_for_replay_safety`.

workers/automation/src/jobctrl/pipeline/workflow.py
ebarti: Fixed in 5e1d015cf: both `_cancel_material_stage_state` helpers (pipeline and preparation) now use the same bounded policy as `_cancel_owned_enrichment` (1s initial / 10s max interval / 5 attempts), with a comment noting that rows left running by an exhausted cancel remain recoverable via `recover_preparation_state`. `_recover_stage_state` was left as-is per the comment's scope. Regressions: `test_pipeline_material_cancel_activity_retries_are_bounded`, `test_preparation_material_cancel_activity_retries_are_bounded`.

workers/automation/src/jobctrl/domain/materials/use_cases.py
ebarti: Fixed in 20ec0df48: when a durable attempt is present the audit key is now `{execution_id}:{durable_attempt}:{recorded_at}` (`recorded_at` is microsecond-precision `_utc_now()`), so post-`--reset-attempts` reruns and same-run activity retries append their own report instead of dropping the newer one, keeping the singular `tailoring_attempt_audit` consistent with the history tail; identical re-merges stay idempotent. The `durable_attempt is None` key shape is unchanged (already unique per recorded execution). Docs sentence in `docs/architecture/tailoring.md` updated to match. Regression: `test_tailor_use_case_appends_rerun_of_same_durable_attempt`.

ebarti:
Test move from #751 landed here: `test_selected_tailor_activity_uses_bounded_requested_workers`, `test_selected_cover_activity_uses_bounded_requested_workers`, and `test_selected_tailor_activity_keeps_item_failures_partial` now belong to this branch's own history (fb1a55e72 cherry-picks the removal from #751, 44f1b072f re-adds them beside the implementation; the tree for `test_jsonrpc_handlers.py` is byte-identical to the previous tip). When this branch is restacked onto the new #751 head (72c80d9e7, which no longer ships these tests), the cherry-picked removal becomes an empty patch and drops out, and the re-add keeps the tests here. All 3 pass on this head: `uv --project workers/automation run --frozen --extra dev python -m pytest workers/automation/tests/test_jsonrpc_handlers.py -q` (3 selected tests: 3 passed).

PR #754, opened 2026-08-06
fix(scoring): bind fit reports to current posting analysis

ebarti:
## Summary

- resolve employer analysis through its cache-owning use case before every fresh score
- reject analyses whose posting snapshot hash no longer matches the current job
- persist requirement-fit reports against the refreshed analysis generation
- fall back to a persisted analysis for the current posting snapshot when the fresh ensemble refresh fails, instead of failing the score stage
- suppress duplicate cache-hit `EmployerAnalyzed` timeline rows on the score stage (tailor path unchanged)

## Why

Scoring could reuse the latest persisted employer analysis even after the posting changed. Tailoring then refreshed the analysis independently and received a requirement-fit report from the previous generation, producing an incoherent coverage plan.

Because the analyze use case owns the full cache identity (posting snapshot + `PROMPT_VERSION` + SDK-set version), resolving through it means a prompt or enabled-leg change intentionally invalidates cached analyses at score time: the next scoring/rescore pass re-runs the 3-SDK ensemble once per selected job (a one-time cost/latency hit after such an upgrade). When that refresh fails (e.g. provider outage), scoring degrades to a persisted analysis for the SAME posting snapshot rather than failing the stage; it still never scores against a different snapshot.

## Validation

- `uv --project workers/automation run --extra dev python -m pytest workers/automation/tests/test_scorer.py workers/automation/tests/test_v7_score_runtime.py -q` — 30 passed
- adjacent suites (`test_analyze_job_use_case`, `test_score_use_cases`, `test_activity_score`, `test_score_activity_recovery`, `test_employer_analysis_leg_config`, `test_v7_tailor_runtime`) — 81 passed
- `git diff --check` — passed

## Checklist

### DIFF
--- a/workers/automation/src/jobctrl/scoring/scorer.py
+++ b/workers/automation/src/jobctrl/scoring/scorer.py
@@ -184,7 +184,14 @@ def score_job(
         if employer_analysis_repository is None:
             employer_analysis_repository = SqliteEmployerAnalysisRepository(conn)
         if analyze_use_case is None and require_employer_analysis:
-            analyze_use_case = build_analyze_use_case(conn=conn, publisher=publisher, event_stage="score")
+            analyze_use_case = build_analyze_use_case(
+                conn=conn,
+                publisher=publisher,
+                event_stage="score",
+                # Score resolves the analysis before EVERY fresh score; cache
+                # hits must not append duplicate job_events timeline rows.
+                record_cached_hits=False,
+            )
         employer_analysis = _ensure_employer_analysis_for_job(
             repository=employer_analysis_repository,
             analyze_use_case=analyze_use_case,
@@ -250,7 +257,14 @@ def run_scoring(
     if employer_analysis_repository is None:
         employer_analysis_repository = SqliteEmployerAnalysisRepository(conn)
     if analyze_use_case is None and require_employer_analysis:
-        analyze_use_case = build_analyze_use_case(conn=conn, publisher=publisher, event_stage="score")
+        analyze_use_case = build_analyze_use_case(
+            conn=conn,
+            publisher=publisher,
+            event_stage="score",
+            # Score resolves the analysis before EVERY fresh score; cache
+            # hits must not append duplicate job_events timeline rows.
+            record_cached_hits=False,
+        )
 
     use_case = _build_use_case(
         repository=repository,
@@ -763,7 +777,14 @@ def score_job_by_id(
         if employer_analysis_repository is None:
             employer_analysis_repository = SqliteEmployerAnalysisRepository(conn)
         if analyze_use_case is None and require_employer_analysis:
-            analyze_use_case = build_analyze_use_case(conn=conn, publisher=publisher, event_stage="score")
+            analyze_use_case = build_analyze_use_case(
+                conn=conn,
+                publisher=publisher,
+                event_stage="score",
+                # Score resolves the analysis before EVERY fresh score; cache
+                # hits must not append duplicate job_events timeline rows.
+                record_cached_hits=False,
+            )
         employer_analysis = _ensure_employer_analysis_for_job(
             repository=employer_analysis_repository,
             analyze_use_case=analyze_use_case,
@@ -1100,22 +1121,44 @@ def _ensure_employer_analysis_for_job(
     # available.  Loading the latest row directly can return an analysis for an
     # older posting snapshot; scoring against that row creates a requirement-fit
     # report which Tailoring must later discard after it refreshes the analysis.
+    refresh_error: Exception | None = None
     if analyze_use_case is not None:
-        outcome = analyze_use_case.execute(job=job, tenant_id=tenant_id)
-        analysis = getattr(outcome, "analysis", None)
-        if _is_usable_employer_analysis(analysis, job):
-            return analysis
-        raise ValueError("Employer analysis did not match the current grounded job snapshot.")
-    if _is_usable_employer_analysis(existing, job):
-        return existing
-    if repository is not None:
-        loaded = _load_employer_analysis_for_job(
+        try:
+            outcome = analyze_use_case.execute(job=job, tenant_id=tenant_id)
+        except Exception as exc:  # noqa: BLE001 — degrade to a same-snapshot persisted analysis below
+            refresh_error = exc
+        else:
+            analysis = getattr(outcome, "analysis", None)
+            if _is_usable_employer_analysis(analysis, job):
+                return analysis
+            # The use case hashes the same job dict scoring validates against,
+            # so on this branch the snapshot always matches — the only reachable
+            # failure is an analysis without grounded requirements.
+            raise ValueError("Employer analysis did not produce grounded requirements for this job.")
+    # Fallback tier: a persisted analysis is acceptable only when it passes the
+    # same-snapshot usability check — never score against a different posting
+    # snapshot. This keeps scoring available when the ensemble refresh fails
+    # (e.g. provider outage) and an analysis for the CURRENT snapshot exists.
+    fallback = existing if _is_usable_employer_analysis(existing, job) else None
+    if fallback is None and repository is not None:
+        fallback = _load_employer_analysis_for_job(
             repository=repository,
             tenant_id=tenant_id,
             job=job,
         )
-        if loaded is not None:
-            return loaded
+    if fallback is not None:
+        if refresh_error is not None:
+            log.warning(
+                "Employer analysis refresh failed for tenant=%s job=%s; scoring against persisted "
+                "generation %s for the current posting snapshot: %s",
+                tenant_id,
+                job.get("job_id"),
+                fallback.generation,
+                refresh_error,
+            )
+        return fallback
+    if refresh_error is not None:
+        raise refresh_error
     if require:
         raise ValueError("Scoring requires employer analysis before a fresh score can be computed.")
     return None
--- a/workers/automation/src/jobctrl/scoring/scorer.py
+++ b/workers/automation/src/jobctrl/scoring/scorer.py
@@ -35,7 +35,8 @@ from jobctrl.domain.job_content_identity import (
     role_title_has_reference_suffix,
     role_titles_match_as_repost,
 )
-from jobctrl.domain.materials.analysis import EmployerAnalysis
+from jobctrl.domain.materials.analysis import EmployerAnalysis, compute_snapshot_hash
+from jobctrl.domain.materials.analyze_use_case import build_jd_snapshot
 from jobctrl.domain.ports.events import EventPublisher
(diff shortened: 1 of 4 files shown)


workers/automation/src/jobctrl/scoring/scorer.py
```diff
@@ -1094,6 +1095,17 @@ def _ensure_employer_analysis_for_job(
 ) -> EmployerAnalysis | None:
     """Resolve the canonical requirement source before a fresh score attempt."""
 
+    # ``AnalyzeJobUseCase`` owns the complete cache identity: posting snapshot,
+    # prompt version, and enabled SDK/model set.  Always ask it first when it is
+    # available.  Loading the latest row directly can return an analysis for an
+    # older posting snapshot; scoring against that row creates a requirement-fit
+    # report which Tailoring must later discard after it refreshes the analysis.
+    if analyze_use_case is not None:
```
ebarti, line 1125: PR Reviewer - pr-754-reviewer

Severity: Medium
Gate: Pass
Evidence: analyze_use_case.py:143-155 (cache key = snapshot_hash + prompt_version + sdk_set_version), employer_analysis.py:110-119 (sdk_set_version derived from currently enabled legs), scorer.py:252-253/765-766 (an analyze use case is always built when require_employer_analysis=True, i.e. every production path)

Moving the use case to the front makes "stale" strictly wider than the PR goal (posting-snapshot coherence). Because the cache key also includes prompt_version and sdk_set_version, any PROMPT_VERSION bump or enabled-leg toggle cache-misses every selected job, and the next scoring/rescore pass re-runs the full 3-SDK ensemble serially on the batch main thread (scorer.py:433-458, no wall-clock timeout per D-19) before any scoring starts. It also removes the previous resilience path: when the ensemble raises (provider outage), the exception propagates and the job's score stage fails even though a persisted analysis for the *current* snapshot (older prompt) exists and would pass the new hash check. Note the `existing`/repository fallbacks below are now unreachable in production, since analyze_use_case is always non-None there.

Required fix: If prompt/SDK-set invalidation at score time is intended, state it in the PR body (cost/latency implications on upgrade). Otherwise, on use-case failure fall back to a persisted analysis that passes the new snapshot-hash check before failing the stage.

workers/automation/src/jobctrl/scoring/scorer.py
```diff
@@ -1094,6 +1095,17 @@ def _ensure_employer_analysis_for_job(
...
+        outcome = analyze_use_case.execute(job=job, tenant_id=tenant_id)
+        analysis = getattr(outcome, "analysis", None)
+        if _is_usable_employer_analysis(analysis, job):
+            return analysis
+        raise ValueError("Employer analysis did not match the current grounded job snapshot.")
```
ebarti: PR Reviewer - pr-754-reviewer

Severity: Low
Gate: Pass
Evidence: analyze_use_case.py:139-155,181-196 — execute() looks up the cache by (and persists with) compute_snapshot_hash(build_jd_snapshot(job)) from the same job dict _is_usable_employer_analysis hashes at scorer.py:1141, so the snapshot check can never fail on this branch; the only reachable trigger is empty canonical.requirements (scorer.py:1151)

The new message misdiagnoses the only reachable failure. When the fresh/cached analysis has zero requirements, the user-visible stage error becomes "Employer analysis failed: Employer analysis did not match the current grounded job snapshot." — pointing operators at a posting-drift race that cannot occur on this path. The old message ("did not produce grounded requirements for this job") was accurate.

Required fix: Restore requirements-focused wording, or branch the message on which usability check actually failed.

workers/automation/src/jobctrl/scoring/scorer.py
```diff
@@ -1094,6 +1095,17 @@ def _ensure_employer_analysis_for_job(
 ) -> EmployerAnalysis | None:
     """Resolve the canonical requirement source before a fresh score attempt."""
 
+    # ``AnalyzeJobUseCase`` owns the complete cache identity: posting snapshot,
+    # prompt version, and enabled SDK/model set.  Always ask it first when it is
+    # available.  Loading the latest row directly can return an analysis for an
+    # older posting snapshot; scoring against that row creates a requirement-fit
+    # report which Tailoring must later discard after it refreshes the analysis.
+    if analyze_use_case is not None:
+        outcome = analyze_use_case.execute(job=job, tenant_id=tenant_id)
```
ebarti: PR Reviewer - pr-754-reviewer

Severity: Low
Gate: Pass
Evidence: employer_analysis.py:114 (build_analyze_use_case always wires EmployerAnalyzedEventRecorder when no publisher is passed; runner.py:2463 passes none), analyze_use_case.py:150-155 (_publish fires on cache hits too, then the recorder does record_job_event + conn.commit per event)

Previously execute() ran only when no usable analysis existed; now it runs before every fresh score computation. Every scoring pass — including per-job retries (up to the 5-attempt cap) and rescore batches — appends a duplicate "Employer analysis generation N (x/y legs)" job_events row per computed job, polluting the per-job audit timeline the dashboard renders. The message text does not distinguish cached hits (only the payload carries cached=true).

Required fix: Suppress cached=True publishes on the score-stage path, or dedupe/label cached hits in EmployerAnalyzedEventRecorder.

workers/automation/tests/test_scorer.py
```diff
@@ -817,6 +833,60 @@ def test_run_scoring_loads_persisted_employer_analysis_into_prompt(
     assert report.assessments[0].fit.kind == "missing"
 
 
+def test_run_scoring_refreshes_stale_posting_analysis_before_persisting_fit_report(
```
ebarti, line 984: PR Reviewer - pr-754-reviewer

Severity: Medium
Gate: Pass
Evidence: test_scorer.py:150-156 (autouse _stub_default_analyzer replaces build_analyze_use_case with the stub for every test) + the new analyze-first order in scorer.py:1103-1107. Verified by code path: in test_run_scoring_loads_persisted_employer_analysis_into_prompt (line 780) the stub returns a matching generation-1 analysis first, so the row saved via SqliteEmployerAnalysisRepository is never read.
Reproduction: Reasoned from code; suite run: `uv --project workers/automation run --extra dev pytest -q workers/automation/tests/test_scorer.py` — 19 passed

This PR silently neutralizes the pre-existing test test_run_scoring_loads_persisted_employer_analysis_into_prompt: it still passes, but only because the stubbed analyze use case fabricates an analysis identical to the persisted one, making it a duplicate of test_run_scoring_generates_employer_analysis_before_prompt. The repository fallback's success path (_load_employer_analysis_for_job, scorer.py:1111-1118) — still shipped and reachable via require_employer_analysis=False — now has no coverage that would fail if it broke.

Required fix: Repoint that test at the path it names (e.g. run_scoring(..., analyze_use_case=None, require_employer_analysis=False) with the persisted row, or score_job with only a repository), or rename/delete it so the coverage loss is explicit.

ebarti (commented):
Review of fix(scoring): bind fit reports to current posting analysis (head 540d8c7).

The core mechanism is sound: `AnalyzeJobUseCase.execute` keys its cache on the snapshot hash computed from the same job dict the scorer now hashes in `_is_usable_employer_analysis`, both batch (`get_jobs_by_stage`, which promotes `je_full_description` into `full_description`) and per-job (`SqlitePreparationTargetReader`) loaders feed `build_jd_snapshot` consistently, and the fit report's `employer_analysis_generation` follows the analysis actually used at compute time. The new regression test reproduces the stale-snapshot case end to end.

Findings: 2 Medium, 2 Low (inline). Summary:
- Medium: the analyze-first reorder widens invalidation beyond posting drift (prompt_version / sdk_set_version drift now re-runs the 3-SDK ensemble for every job) and drops the fallback to a same-snapshot persisted analysis when the ensemble fails.
- Medium: `test_run_scoring_loads_persisted_employer_analysis_into_prompt` is silently neutralized by the reorder (the autouse stub wins before the persisted row is read), leaving the repository fallback's success path uncovered.
- Low: the ValueError message at scorer.py:1108 misdiagnoses the only reachable failure (empty requirements) as snapshot drift.
- Low: cache-hit EmployerAnalyzed publishes now append a duplicate job_events timeline row per computed job per scoring pass.

Verification run at head: `pytest -q workers/automation/tests/test_scorer.py` — 19 passed (note: the PR body claims 69 passed for this exact command; the file collects 19 tests at this head). Adjacent suites (test_activity_score, test_score_activity_recovery, test_score_use_cases, test_v7_score_runtime, test_analyze_job_use_case) — 61 passed. Focused Ruff on both touched files — clean. `git diff --check` — clean.

No Blocker/High findings.

workers/automation/src/jobctrl/scoring/scorer.py
ebarti: Fixed in 6fe0d67. `_ensure_employer_analysis_for_job` now catches a refresh failure from the use case and falls back to `existing` / the repository row — accepted only when it passes the same-snapshot usability check — before failing the stage; with no usable same-snapshot row the original error still propagates (the existing no-persisted-row failure test is unchanged). The prompt/SDK-set invalidation at score time is intended and is now stated in the PR body with its cost/latency implication. Regression tests: `test_run_scoring_falls_back_to_persisted_analysis_when_refresh_fails` and `test_run_scoring_fails_when_refresh_fails_and_persisted_analysis_is_stale`.

workers/automation/src/jobctrl/scoring/scorer.py
ebarti: Fixed in 6fe0d67 — restored the requirements-focused wording ("Employer analysis did not produce grounded requirements for this job.") with a comment noting the snapshot cannot mismatch on that branch.

workers/automation/src/jobctrl/scoring/scorer.py
ebarti: Fixed in 6fe0d67. `EmployerAnalyzedEventRecorder` gains `record_cached_hits` and the three score-stage `build_analyze_use_case` call sites pass `record_cached_hits=False`, so cache hits no longer append `job_events` rows on scoring passes (retries/rescore included); the tailor-path default recorder is unchanged. Covered by `test_employer_analysis_recorder_skips_cache_hits_when_opted_out`, `test_employer_analysis_recorder_records_cache_hits_by_default`, and `test_run_scoring_builds_score_stage_analyzer_without_cached_event_rows`.

workers/automation/tests/test_scorer.py
ebarti: Fixed in 6fe0d67. `test_run_scoring_loads_persisted_employer_analysis_into_prompt` now runs with `require_employer_analysis=False` plus a guard patch that fails the test if any analyzer is built, so the persisted row is genuinely read through `_load_employer_analysis_for_job`'s success path; the new outage-fallback test additionally exercises that path with a (failing) use case present.

PR #755, opened 2026-08-06
fix(materials): restore requirement-led tailoring

ebarti:
## Summary

- block Tailor on Score when requirement-fit evidence is missing, cross-job, or from a different employer-analysis generation
- preserve Tailor retry counters for prerequisite failures and record auditable `StageBlocked` events in both per-job and legacy batch runners
- require exactly one bound claim mapping for every generated summary sentence, experience bullet, and complete rendered skill group
- require the model and voice pass to return an ordered `executive_profile_sentences` array whose exact one-space join reproduces the stored summary
- accept documented aggregate summary aliases only for one-sentence summaries; require indexed mappings for every sentence in multi-sentence summaries and reject extra aggregate mappings
- require exact rendered text for summary, bullet, and skill-group mappings, while rejecting missing, duplicate, partial, and out-of-range mappings before the judge
- reject outer-whitespace normalization and preserve the last accepted summary when voice output breaks sentence identity
- preserve the raw model payload for audit while normalizing only the parsed domain representation
- add end-to-end fixtures proving coherent Score evidence reaches the judge and produces approved material

## Why

Requirement-led Tailoring silently dropped stale fit evidence, built a zero-edge coverage graph, and spent durable retries on candidates that could not pass validation. Coherent jobs could also fail before the judge—or pass with incomplete audit coverage—when generated summary, bullet, and skill mappings did not match the validator's surface contract.

## Validation

- focused final Tailoring contract slice — 118 passed
- `uv --project workers/automation run python -m pytest -q` — 3,371 passed, 2 skipped
- `corepack pnpm check` — passed
- `corepack pnpm docs:build` — passed; 9,651 links resolved
- independent QA gate — passed with no findings: coverage-bearing aggregate rejection for multi-sentence summaries; indexed multi-sentence and one-sentence aggregate happy paths; byte-exact generator/voice reconstruction; education coverage; prerequisite retry preservation
- `git diff --check` — passed

## Checklist

### DIFF
--- a/workers/automation/src/jobctrl/domain/materials/use_cases.py
+++ b/workers/automation/src/jobctrl/domain/materials/use_cases.py
@@ -563,9 +563,18 @@ def _claim_mapping_binding_errors(
                 "does not exist in the generated payload."
             )
             continue
-        if not _claim_text_is_bound(actual_text, mapping.text):
+        if _skill_group_claim_location(location):
+            text_is_bound = actual_text == mapping.text
+        else:
+            text_is_bound = _claim_text_is_bound(actual_text, mapping.text)
+        if not text_is_bound:
+            relationship = (
+                "does not exactly match"
+                if _skill_group_claim_location(location)
+                else "is not present at"
+            )
             errors.append(
-                f"Generated claim {mapping.claim_id} text is not present at "
+                f"Generated claim {mapping.claim_id} text {relationship} "
                 f"generated payload location {mapping.location!r}."
             )
     return tuple(errors)
@@ -671,6 +680,17 @@ def _canonical_claim_location(location: str) -> str:
     ):
         return "executive_profile"
     return re.sub(r"\.bullet\[(\d+)\]$", r".bullets[\1]", normalized)
+
+
+def _skill_group_claim_location(location: str) -> bool:
+    """Return whether a mapping targets one complete rendered skill group."""
+    return bool(
+        re.fullmatch(
+            r"(?:skills|skill_categories|skill_category_updates)\.[^.\[]+",
+            location,
+        )
+        or re.fullmatch(r"skill_category_updates\[\d+\]", location)
+    )
 
 
 def _claim_text_is_bound(actual_text: str, mapped_text: str) -> bool:
--- a/workers/automation/src/jobctrl/domain/materials/use_cases.py
+++ b/workers/automation/src/jobctrl/domain/materials/use_cases.py
@@ -271,9 +271,10 @@ TAILORED_RESUME_RESPONSE_SCHEMA: dict[str, Any] = {
                     "location": {
                         "type": "string",
                         "description": (
-                            "Use executive_profile or executive_profile.sentence[N] for summary "
-                            "sentences, experience.<id>.bullets[N] for bullets, and skills.<id> "
-                            "for one complete rendered skill group."
+                            "Use executive_profile only for a one-sentence summary; otherwise use "
+                            "executive_profile.sentence[N] for each summary sentence. Use "
+                            "experience.<id>.bullets[N] for bullets and skills.<id> for one "
+                            "complete rendered skill group."
                         ),
                     },
                     "text": {
@@ -730,18 +731,21 @@ def _required_claim_surface_groups(
 
     groups: list[tuple[str, frozenset[str]]] = []
     executive_profile = str(payload.get("executive_profile") or "").strip()
-    for index, _sentence in enumerate(_generated_summary_sentences(executive_profile)):
+    summary_sentences = _generated_summary_sentences(executive_profile)
+    for index, _sentence in enumerate(summary_sentences):
+        locations = {f"executive_profile.sentence[{index}]"}
+        if len(summary_sentences) == 1:
+            locations.update(
+                {
+                    "executive_profile",
+                    "summary",
+                    "resume.executive_profile",
+                }
+            )
         groups.append(
             (
                 f"executive profile sentence {index}",
-                frozenset(
-                    {
-                        "executive_profile",
-                        "summary",
-                        "resume.executive_profile",
-                        f"executive_profile.sentence[{index}]",
-                    }
-                ),
+                frozenset(locations),
             )
         )
 
@@ -1431,7 +1435,8 @@ HARD RULES:
   group, emit a generated_claim_mappings entry that references valid
   coverage_edge_ids, requirement_ids, and evidence_ids; use non_requirement_reason
   only for pinned, positioning, or structure claims
-- A summary sentence location may be executive_profile.sentence[N]; N is zero-based
+- Use executive_profile only when the generated summary has exactly one sentence;
+  otherwise map every sentence with executive_profile.sentence[N], where N is zero-based
 - For a skills.<category-id> mapping, text must be the exact selected items in
   rendered order joined with ", " (comma plus one space), not the category label
 - Every achievement_evidence_id present on any COVERAGE_GRAPH edge must appear
--- a/workers/automation/src/jobctrl/domain/materials/use_cases.py
+++ b/workers/automation/src/jobctrl/domain/materials/use_cases.py
@@ -712,6 +712,10 @@ def _claim_location_requires_exact_text(location: str) -> bool:
     return bool(
         location in {"executive_profile", "summary", "resume.executive_profile"}
         or re.fullmatch(r"executive_profile\.sentence\[\d+\]", location)
+        or re.fullmatch(
+            r"(?:experience|experience_updates)(?:\.[^.\[]+|\[\d+\])\.bullets\[\d+\]",
+            location,
+        )
         or _skill_group_claim_location(location)
     )
 
--- a/workers/automation/src/jobctrl/domain/materials/use_cases.py
+++ b/workers/automation/src/jobctrl/domain/materials/use_cases.py
@@ -176,8 +176,8 @@ from jobctrl.resume_profile import (
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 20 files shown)


workers/automation/src/jobctrl/domain/materials/use_cases.py
```diff
@@ -645,7 +701,145 @@ def _generated_claim_surfaces(
...
 
 def _canonical_claim_location(location: str) -> str:
-    return re.sub(r"\.bullet\[(\d+)\]$", r".bullets[\1]", str(location or "").strip())
+    normalized = str(location or "").strip()
+    sentence_match = re.fullmatch(
+        r"(?:(?:profile\.)?(?:executive_profile|summary))(?:\.sentences?)?\[(\d+)\]",
+        normalized,
+    )
+    if sentence_match:
+        return f"executive_profile.sentence[{sentence_match.group(1)}]"
```
ebarti, line 711: PR Reviewer - pr-755-reviewer

Severity: Medium
Gate: Pass
Evidence: `bullet_id_for_claim_location` (workers/automation/src/jobctrl/domain/materials/claim_grounding.py:144) returns `None` for `executive_profile.sentence[N]`; its docstring promises to mirror "the alias forms accepted by the payload-surface validator (_generated_claim_surfaces)", which this hunk extends without mirroring. `_SUMMARY_LOCATIONS` only contains the three aggregate aliases, which this PR rejects for multi-sentence summaries.
Reproduction: Reproduced with a script in the review worktree: `bullet_id_for_claim_location("executive_profile.sentence[0]") -> None`, and `ground_claim_mappings` bound a sentence-located coverage claim `via="text_scan"` to BOTH `executive_profile#0` and an unrelated `experience:acme#0` row containing the same normalized sentence.

The prompt's default summary is 2-4 sentences and this PR rejects aggregate aliases for multi-sentence summaries, so on the common path every coverage-bearing summary claim now grounds through the "resilience fallback for location alias drift" instead of `via="location"`. Consequences: (1) the `via` audit label loses its meaning for summary claims on every generation; (2) text-scan multi-hits let `enrich_provenance_requirements` stamp summary requirement chips onto unrelated experience rows whose text happens to contain the sentence tokens, polluting the per-bullet audit chips and the grounded coverage attribution.

Required fix: teach `bullet_id_for_claim_location` to map the canonical `executive_profile.sentence[N]` form (and the alias spellings `_canonical_claim_location` accepts) to `executive_profile#0`, and add a regression test asserting a sentence-indexed summary claim grounds `via="location"`.

workers/automation/src/jobctrl/domain/materials/voice.py
```diff
@@ -155,8 +172,25 @@ def apply_voice_to_payload(tailored_payload: dict, result: VoiceResult) -> dict:
...
-    if result.executive_profile.strip():
+    source_summary_sentences = tuple(
+        str(sentence)
+        for sentence in voiced.get("executive_profile_sentences") or ()
+        if str(sentence).strip()
+    )
+    voiced_summary_sentences = tuple(
+        str(sentence) for sentence in result.executive_profile_sentences
+    )
+    if (
```
ebarti: PR Reviewer - pr-755-reviewer

Severity: Medium
Gate: Pass
Evidence: `VoicePayload.executive_profile_sentences` is declared with `default_factory=list` (voice.py:74-80), so `VoicePayload.model_json_schema()` — used verbatim as the SDK `output_format` in voice_adapter.py:167 — does not require the field. Under the v3 tailor schema the source payload always has >=1 sentence, so an omitted/mismatched result fails `len(voiced_summary_sentences) == len(source_summary_sentences)` here and the voiced summary is dropped. Nothing records that drop: `_run_voice` (use_cases.py:3537-3554) builds `VoicePassRecord` from bullet proxy deltas only, so the record can read `accepted=True` while the summary voicing was silently discarded. The new tests cover the preservation but assert no audit record.
Reproduction: Reproduced with a script: `apply_voice_to_payload` with a result carrying a voiced summary but empty `executive_profile_sentences` kept the original summary and still applied the voiced bullets, with no trace of the rejection.

Preserving the last accepted summary on identity break is the right call, but two gaps make it silently lossy: a voice model that omits the optional field permanently disables summary voicing with zero signal, and the PR-body behavior "preserve the last accepted summary when voice output breaks sentence identity" leaves no inspectable audit trail when it fires, contrary to the repo's lifecycle-labeling discipline for post-generation decisions.

Required fix: require the field in the voice output contract (e.g. `min_length=1` so the SDK json_schema marks it required), and record a summary-rejection reason on the voice audit record (VoicePassRecord or its metadata) when this gate keeps the original summary.

workers/automation/src/jobctrl/domain/materials/use_cases.py
```diff
@@ -645,7 +701,145 @@ def _generated_claim_surfaces(
...
+    return re.sub(r"\.bullet\[(\d+)\]$", r".bullets[\1]", normalized)
+
+
+def _skill_group_claim_location(location: str) -> bool:
+    """Return whether a mapping targets one complete rendered skill group."""
+    return bool(
+        re.fullmatch(
+            r"(?:skills|skill_categories|skill_category_updates)\.[^.\[]+",
```
ebarti: PR Reviewer - pr-755-reviewer

Severity: Low
Gate: Pass
Evidence: The id segment here is `[^.\[]+`, and the bullet arm of `_claim_location_requires_exact_text` uses `(?:\.[^.\[]+|\[\d+\])`. An id containing a dot (e.g. a skill category id like `node.js`) fails these fullmatches, while `_generated_claim_surfaces` and `_required_claim_surface_groups` still register `skills.node.js` as a valid surface/group location.
Reproduction: Reasoned from code; the mapping for such a location falls through to `_claim_text_is_bound` (normalized substring containment) instead of the exact-equality branch.

For dotted ids the new "complete rendered skill group binds by exact text" contract silently downgrades to substring matching, so a partial item list would pass binding while still satisfying the exactly-one-group-mapping count. Same limitation exists in claim_grounding's `_SKILLS_LOCATION_RE`/`_EXPERIENCE_LOCATION_RE`, so grounding also degrades for those ids. Low because profile ids are typically dot-free slugs.

Required fix: match dotted ids in these location regexes (the trailing component shapes are unambiguous), or reject/normalize dotted profile ids at the boundary; either way add a test with a dotted category id.

ebarti (commented):
Reviewed f2d87c7 against base fix/scoring-analysis-snapshot-coherence. The prerequisite-block wiring (both runners), retry preservation, exact-surface binding contract, and raw-payload audit preservation all check out; targeted suites pass locally (182 tests across the 6 touched files). Three findings: 2 Medium (summary-sentence claims now ground via the text_scan drift fallback because bullet_id_for_claim_location was not taught the new location form; voice summary rejection is silent and the voice output schema does not require the new sentences field) and 1 Low (dotted ids silently downgrade exact-text binding). No Blocker/High.

workers/automation/src/jobctrl/domain/materials/use_cases.py
ebarti: Fixed in 6aa13a07c. `bullet_id_for_claim_location` now maps `executive_profile.sentence[N]` and every alias spelling `_canonical_claim_location` accepts (`summary[N]`, `profile.executive_profile.sentences[N]`, ...) to `executive_profile#0` via a mirrored `_SUMMARY_SENTENCE_LOCATION_RE`. Regression tests added: `test_location_maps_summary_sentence_aliases_to_the_profile_line` and `test_summary_sentence_claim_grounds_via_location_without_cross_binding`, which reproduces your scenario (an unrelated experience row carrying the same sentence) and asserts `via="location"`, a single `executive_profile#0` binding, and that `enrich_provenance_requirements` stamps nothing onto the experience row.

workers/automation/src/jobctrl/domain/materials/voice.py
ebarti: Fixed in 6aa13a07c. `VoicePayload.executive_profile_sentences` is now required with `min_length=1`, so the SDK `output_format` schema marks it `required` with `minItems: 1` (asserted in `test_passes_voice_schema_empty_tools_and_no_turn_cap` and `test_voice_payload_schema_requires_summary_sentences`); an omission now fails `model_validate` and lands on the recorded `voice_error` path instead of silently disabling summary voicing. The identity gate is extracted into `summary_voice_rejection_reason` (same predicate, now labeled: missing / outer_whitespace / count_mismatch / join_mismatch) and `_run_voice` records it as `VoicePassRecord.summary_rejection_reason` (persisted through `to_dict`/`from_dict` and `voice_pass_json`; shared parity fixture updated). `test_summary_identity_break_is_recorded_on_the_voice_audit` reproduces your script: voiced bullets accepted, original summary shipped, and the record now carries `voiced_summary_sentence_count_mismatch`.

workers/automation/src/jobctrl/domain/materials/use_cases.py
ebarti: Fixed in 6aa13a07c. The id segments in `_skill_group_claim_location`, the experience arm of `_claim_location_requires_exact_text`, and claim_grounding's `_SKILLS_LOCATION_RE`/`_EXPERIENCE_LOCATION_RE` now use `[^\[\]]+`, so dotted ids bind exactly (the trailing `.bullets[N]`/`.items[N]` bracket shapes keep the split unambiguous). `test_claim_binding_requires_exact_text_for_dotted_ids` proves a partial `skills.node.js` item list and a partial `experience.acme.co.bullets[0]` text are rejected with "does not exactly match", and `test_dotted_skill_category_claim_grounds_via_location`/`test_location_maps_dotted_profile_ids` cover the grounding side.

PR #757, opened 2026-08-06
fix(enrichment): separate content trust from apply readiness

ebarti:
## What changed

- separates posting-description confidence from application-target readiness
- repairs legacy readable snapshots with an immutable reclassification and releases only stale Enrichment quarantine blockers
- detects LinkedIn on-site application controls without clicking them or retrying forever
- persists distinct outcomes for recovered, on-site, missing-control, missing-target, navigation, and unsafe-target cases
- projects the same privacy-safe, allow-listed outcome through both Python and TypeScript read-model writers
- keeps cancellation reconciliation outside slow authenticated browser transactions
- documents the user-facing confidence, retry, and audit contracts

## Why

A complete posting description could be labeled low confidence solely because no external application URL was available. That incorrectly blocked Tailor, and URL recovery collapsed legitimate LinkedIn on-site flows together with transient failures as a generic unresolved result.

## Validation

- independent reviewer gate: PASS (Blocker 0, High 0, Medium 0, Low 0)
- independent QA gate: PASS (Blocker 0, High 0, Medium 0, Low 0)
- full Python suite: 3383 passed, 2 skipped
- full API suite outside the localhost sandbox: 750 passed
- shared Python/TypeScript projection parity: 3 passed / 2 passed
- focused Enrichment, content-trust, and cancellation regressions: 35 passed
- focused API outcome and inactive-posting projections: 56 passed
- rendered application-target diagnostics: 17 passed
- full web unit suite: 2004 passed
- aggregate cross-stack check: passed
- documentation build and link checks: passed, 9651 references resolved
- Ruff and `git diff --check`: passed

## Checklist

- [x] Relevant local checks are listed above.
- [x] No secrets, private profile data, generated application materials, raw logs, browser profiles, SQLite databases, or local paths are included.

### DIFF
--- a/workers/automation/src/jobctrl/enrichment/detail.py
+++ b/workers/automation/src/jobctrl/enrichment/detail.py
@@ -719,7 +719,11 @@ def _apply_authenticated_linkedin_apply_url(
         else:
             return cascade_result
     except Exception as exc:  # noqa: BLE001 - resolver is best-effort
-        log.warning("LinkedIn apply URL resolver failed for %s: %s", url, exc)
+        log.warning(
+            "LinkedIn apply URL resolver failed for %s (%s)",
+            url,
+            type(exc).__name__,
+        )
         return {
             **cascade_result,
             "authenticated_apply_url_method": "resolver_error",
@@ -1675,10 +1679,6 @@ def scrape_site_batch(
                             stage_metadata["authenticatedApplyUrlMethod"] = cascade_result.get(
                                 "authenticated_apply_url_method"
                             )
-                        if cascade_result.get("authenticated_apply_url_error"):
-                            stage_metadata["authenticatedApplyUrlError"] = cascade_result.get(
-                                "authenticated_apply_url_error"
-                            )
                         stage_metadata.update(_authenticated_apply_url_outcome_metadata(cascade_result))
                         if activity_lease is not None:
                             stage_metadata.update(
@@ -1740,10 +1740,6 @@ def scrape_site_batch(
                         if cascade_result.get("authenticated_apply_url_method"):
                             completed_payload["authenticatedApplyUrlMethod"] = cascade_result.get(
                                 "authenticated_apply_url_method"
-                            )
-                        if cascade_result.get("authenticated_apply_url_error"):
-                            completed_payload["authenticatedApplyUrlError"] = cascade_result.get(
-                                "authenticated_apply_url_error"
                             )
                         completed_payload.update(_authenticated_apply_url_outcome_metadata(cascade_result))
                         if fallback_source:
@@ -2393,7 +2389,10 @@ def _authenticated_apply_url_recovery_error(
     """Translate resolver mechanics into one auditable application-target fact."""
 
     method_value = str(method or "external_url_missing")
-    detail = str(raw_error or "").strip()
+    # Resolver exceptions may contain browser-profile paths, signed URLs, or
+    # other local diagnostics. Stable outcome messages are code-owned and must
+    # never copy that raw detail into stage metadata or user-facing events.
+    del raw_error
     if method_value == "linkedin_onsite_apply":
         return EnrichmentError(
             code="APPLY_URL_LINKEDIN_ONSITE",
@@ -2407,17 +2406,18 @@ def _authenticated_apply_url_recovery_error(
             retryable=True,
         )
     if method_value == "unsafe_url":
-        suffix = f" {detail}" if detail else ""
         return EnrichmentError(
             code="APPLY_URL_UNSAFE_TARGET",
-            message=f"JobCtrl rejected the discovered application target as unsafe.{suffix}",
+            message=(
+                "JobCtrl rejected the discovered application target because it is not "
+                "a safe public HTTP(S) destination."
+            ),
             retryable=False,
         )
     if method_value in {"navigation_error", "resolver_error"}:
-        suffix = f" {detail}" if detail else ""
         return EnrichmentError(
             code="APPLY_URL_NAVIGATION_FAILED",
-            message=f"The authenticated LinkedIn page could not be inspected.{suffix}",
+            message="The authenticated LinkedIn page could not be inspected.",
             retryable=True,
         )
     if method_value == "external_url_missing":
@@ -2439,6 +2439,7 @@ def _authenticated_apply_url_outcome_metadata(result: dict) -> dict[str, object]
         return {}
     if result.get("application_url"):
         return {
+            "authenticatedApplyUrlMethod": str(method),
             "applyUrlOutcomeCode": "APPLY_URL_EXTERNAL_RECOVERED",
             "applyUrlOutcomeMessage": "An external application URL was recovered.",
             "applyUrlOutcomeRetryable": False,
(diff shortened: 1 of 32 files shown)


workers/automation/src/jobctrl/enrichment/detail.py
```diff
@@ -2526,10 +2864,16 @@ def _reset_authenticated_linkedin_retry_candidates(
...
-    if not linkedin_apply_resolver_enabled():
-        return 0
+    selected_job_ids = tuple(dict.fromkeys(canonical_job_id(str(job_id)) for job_id in job_ids))
+    _repair_legacy_missing_apply_url_content_trust_candidates(
+        conn,
+        job_ids=selected_job_ids,
+        activity_lease=activity_lease,
+        cancel_event=cancel_event,
+    )
+    resolver_enabled = linkedin_apply_resolver_enabled()
```
ebarti: PR Reviewer - pr-757-reviewer

Severity: Medium
Gate: Pass
Evidence: On the base ref, `if not linkedin_apply_resolver_enabled(): return 0` was the first statement of `_reset_authenticated_linkedin_retry_candidates`, so with the authenticated-LinkedIn capability not ready (the out-of-box state; the flag requires `browser_capability_status("authenticated-linkedin-browser").status == "ready"`) the entire pass was skipped. After this change only the enriched-row recovery is gated (`if not resolver_enabled: continue`); the failed-row reset group at the bottom of this loop now runs unconditionally, and enrich runs call this with `reset_linkedin_candidates=True` by default (detail.py:3710).
Reproduction: Reasoned from code; the new `test_legacy_missing_apply_url_snapshot_is_reclassified_for_every_source_without_browser_retry` covers only an enriched row, so no test pins the failed-row group's behavior with the resolver disabled.

Unstated behavior change: with the capability off, previously-failed LinkedIn rows (`attempt_count < 3`, retryable last error) are now reset to pending on every enrichment run. If robots blocks the anonymous refetch they flip from `failed` to `blocked`; if robots permits, each cycle records another failed attempt, and because `JobEnrichment.reset()` preserves attempt history and `_attempt_count_from_json` counts all attempts, those anonymous retries consume the same 3-attempt budget the authenticated recovery relies on — once the user later enables the capability, `attempt_count >= _MAX_AUTHENTICATED_LINKEDIN_RETRY_ATTEMPTS` permanently skips those rows.

Required fix: If unintended, gate the failed-row reset group on `resolver_enabled` while keeping the legacy content-trust repair unconditional (the point of this restructure). If intended, add a regression test pinning the resolver-disabled failed-row semantics and state the change in the PR body.

workers/automation/tests/test_linkedin_authenticated_enrichment_retry.py
```diff
@@ -402,6 +533,101 @@ def test_enriched_missing_apply_url_backfills_on_successful_recovery(
...
+        error_code="ENRICHMENT_QUARANTINED",
+        error_message="Tailoring is waiting for a trustworthy posting snapshot.",
+        retryable=True,
+        blocked_by=["enrich"],
+        validate_transition=False,
+    )
+    conn.commit()
+    monkeypatch.setattr(detail, "linkedin_apply_resolver_enabled", lambda: False)
+
+    def unexpected_resolver() -> object:
```
ebarti: PR Reviewer - pr-757-reviewer

Severity: Low
Gate: Pass
Evidence: In `_reset_authenticated_linkedin_retry_candidates`, the factory call is wrapped in `try: resolver = resolver_factory() / except Exception: log.warning(...); resolver_factory = None`. `AssertionError` subclasses `Exception`, so if the production code regressed into invoking the factory, this guard would be swallowed and logged, `resolver` would stay `None`, and the remaining assertions would still pass.

The `raise AssertionError("content-trust repair must not require browser navigation")` guard cannot actually fail this test, so the "without browser retry" half of the test name is not enforced.

Required fix: Use a recording resolver (e.g. `_RecoveryResolver`) and assert `resolver.calls == []` after the sweep, or have the factory set a flag that the test asserts is still unset.

ebarti (commented):
Reviewed bb0b86b against base fix/cover-job-fact-provenance. The content-trust/apply-readiness separation is coherent end to end: the confidence and quarantine decoupling, the immutable legacy reclassification, the allow-listed outcome projection (both writers re-derive message/retryable from the code and drop raw resolver detail), and the non-retryable on-site terminal outcome are all covered by focused tests, which I ran locally (101 Python enrichment tests, 58 API tests incl. cross-writer parity, 17 StageTimeline tests — all pass at the head SHA). Two findings below: one Medium (an unstated behavior change when the authenticated-LinkedIn capability is not ready) and one Low (an ineffective no-browser guard in the new repair regression test). No Blocker/High issues.

workers/automation/src/jobctrl/enrichment/detail.py
ebarti: Fixed in a9d424b5a. The change was unintended: the capability gate is restored as an early return placed after the source-agnostic legacy content-trust repair, so the repair stays capability-independent while both LinkedIn browser-recovery groups (enriched apply-URL recovery and the failed-row reset) are skipped entirely when the resolver capability is not ready. This restores the base-ref disabled-state semantics (no session build, no scan, no resets), removes the now-redundant per-row resolver_enabled check, and adds test_failed_rows_are_not_reset_while_authenticated_capability_is_disabled, which pins that repeated disabled-capability runs leave failed rows untouched (attempt_count unchanged, factory never invoked) and that the row is still eligible for reset once the capability is enabled.

workers/automation/tests/test_linkedin_authenticated_enrichment_retry.py
ebarti: Fixed in a9d424b5a. The raising factory is replaced with a recording factory that returns a _RecoveryResolver; the test now asserts resolver_factory_calls == [] and resolver.calls == [] after the sweep, so a regression that invokes the factory (or navigates) fails observably instead of being swallowed by production's except Exception guard.

ebarti:
Addressed both review findings in a9d424b5a:

- **Medium (detail.py)**: restored the authenticated-LinkedIn capability gate as an early return placed after the source-agnostic legacy content-trust repair. The repair remains capability-independent; both browser-recovery groups (enriched apply-URL recovery and the failed-row reset) are skipped entirely while the capability is not ready, so anonymous retries can no longer consume the bounded authenticated retry budget. New regression test: `test_failed_rows_are_not_reset_while_authenticated_capability_is_disabled`.
- **Low (retry test)**: the legacy-reclassification test's no-browser guard is now observable — a recording factory/resolver with `resolver_factory_calls == []` and `resolver.calls == []` assertions replaces the raising factory that production's `except Exception` swallowed.

Validation: `uv --project workers/automation run --extra dev python -m pytest workers/automation/tests/test_linkedin_authenticated_enrichment_retry.py workers/automation/tests/test_enrichment_aggregate.py workers/automation/tests/test_enrichment_detail_staleness.py workers/automation/tests/test_linkedin_apply_resolver.py workers/automation/tests/test_enrichment_politeness_gate.py workers/automation/tests/test_discover_reliability.py -q` → 160 passed; Ruff on touched files passed; `git diff --check` clean; no uv.lock churn committed.

PR #760, opened 2026-08-06
fix(pipeline): block terminal tailor dependents

ebarti:
## Summary

- block unstarted Cover and Apply rows when Tailor is retryably failed or durably exhausted, with exact upstream codes, dependency metadata, and recovery actions
- reconcile historical rows during database/preparation startup so old ownerless `pending` dependents become truthful
- clear only Tailor-owned dependency blocks after success and preserve concurrent queued/running claims, attempts, metadata, skips, and cancellations
- invalidate an older completed/failed Cover only after a replacement resume is accepted, using a guarded transition and rowcount-gated event

## Why

Cover and Apply could remain `pending` after Tailor had exhausted all durable attempts. That state falsely implied runnable work without a satisfiable prerequisite. An older Cover invalidation helper could also overwrite a live Cover owner during Tailor success.

## Validation

- focused Tailor, preparation, and canonical stage-state suites — 56 passed
- production Tailor-success ownership and exactly-once reset regressions — passed
- deterministic concurrent dependent-claim regression — passed
- docs build and link/redirect checks — passed (9,651 references)
- independent review gate — PASS; zero Blocker, High, or Medium findings; one documentation Low corrected before commit
- no Apply execution or application submission behavior was added

## Stack

Depends on #759.

### DIFF
--- a/workers/automation/src/jobctrl/state.py
+++ b/workers/automation/src/jobctrl/state.py
@@ -70,6 +70,9 @@ _DEPENDENCY_BLOCKER_MESSAGES: dict[str, tuple[tuple[str, tuple[str, ...]], ...]]
         ("apply", ("Materials are not ready.",)),
     ),
 }
+_DEPENDENCY_TERMINAL_ERROR_CODES: dict[str, tuple[str, ...]] = {
+    "tailor": ("UPSTREAM_TAILOR_FAILED", "UPSTREAM_TAILOR_EXHAUSTED"),
+}
 SCORE_ELIGIBILITY_BLOCKED_ERROR_CODE = "SCORE_ELIGIBILITY_BLOCKED"
 SCORE_ELIGIBILITY_BLOCKED_MESSAGE_PREFIX = "Score eligibility blocks tailoring"
 SCORE_THRESHOLD_SKIPPED_ERROR_CODE = "MIN_SCORE"
@@ -415,6 +418,13 @@ def set_stage_state(
             completed_stage=stage,
             now=now,
         )
+    elif stage == "tailor" and state in {"failed", "exhausted"}:
+        reconcile_tailor_terminal_dependents(
+            conn,
+            tenant_id=tenant_id,
+            job_id=stable_job_id,
+            now=now,
+        )
 
 
 def reconcile_dependency_blockers(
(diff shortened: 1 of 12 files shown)


workers/automation/src/jobctrl/state.py
```diff
@@ -415,6 +418,13 @@ def set_stage_state(
             completed_stage=stage,
             now=now,
         )
+    elif stage == "tailor" and state in {"failed", "exhausted"}:
```
ebarti, line 421: PR Reviewer - pr-760-reviewer

Severity: Low
Gate: Pass
Evidence: `SqlitePipelineStateRepository.save()` (workers/automation/src/jobctrl/infrastructure/pipeline/sqlite_repository.py:290-322) upserts every stage row of the aggregate through `set_stage_state`, so this new hook fires mid-loop and its sibling cover/apply writes can be overwritten by the aggregate's own (in-memory, still-pending) cover/apply upserts later in the same loop. Today that clobber does not happen only by ordering luck: `load()` selects without ORDER BY, the PK index returns stages alphabetically (Apply, Cover, ..., Tailor last), so tailor is saved after its dependents — I verified this empirically (load -> set Tailor Failed -> save leaves cover/apply blocked with UPSTREAM_TAILOR_FAILED). But `new_for_job()` builds `stages` in Stage-enum order, and any ORDER BY / stage-name / query-plan change flips the outcome to falsely-pending dependents plus orphan StageBlocked events. Notably, `save()` already defends against exactly this class of mid-loop clobber for the succeeded-hook by re-running `reconcile_dependency_blockers` after the loop (sqlite_repository.py:316-322); there is no symmetric post-loop call for `reconcile_tailor_terminal_dependents`. The repository is currently only constructed in tests (test_pipeline_use_cases.py), so this is latent, not a live regression.
Reproduction: Reproduced ordering behavior with a scratch script (init_db + ensure_job_stage_rows + repo.load/mutate/save); clobber does not fire today solely due to alphabetical iteration order.

The new invariant ("no falsely-pending dependents behind terminal Tailor") is enforced by this hook, but the aggregate save path preserves it only through an undocumented, incidental iteration order.

Required fix: mirror the existing post-loop pattern in `SqlitePipelineStateRepository.save()` — after the stage loop, call `reconcile_tailor_terminal_dependents` for any stage that persisted tailor as failed/exhausted (like `completed_stages` does for `reconcile_dependency_blockers`) — or add a regression test that pins the aggregate-save ordering assumption so a future ordering change cannot silently reintroduce pending dependents.

ebarti (commented):
Reviewed 66e2aa2 against base fix/score-threshold-stage-status. The guarded terminal-dependent blocking, exactly-once cover reset, and startup reconciliation are correct and well tested; I ran the changed suites (test_state_dashboard.py + test_v7_tailor_runtime.py: 42 passed; preparation suites: 21 passed) and verified the concurrent-claim and idempotency guards. One Low, non-blocking robustness note inline about the aggregate-repository save loop interacting with the new set_stage_state hook.

workers/automation/src/jobctrl/state.py
ebarti: Fixed in e3d8cb61f — `save()` now mirrors the post-loop pattern: it tracks whether the loop persisted tailor as failed/exhausted and, symmetric with the `reconcile_dependency_blockers` pass, calls `reconcile_tailor_terminal_dependents` after the stage loop. Added `test_save_blocks_tailor_terminal_dependents_regardless_of_stage_order`, which reorders the aggregate's stages dict so Tailor is persisted before Cover/Apply (the new_for_job enum order) instead of relying on the alphabetical row load; it fails pre-fix with falsely-pending dependents and passes post-fix.

PR #763, opened 2026-08-07
fix(pipeline): freeze global material cohorts

ebarti:
## What changed
- resolve global Tailor and Cover pickup into canonical job IDs before Temporal starts
- distinguish an explicitly empty frozen cohort from an unscoped legacy batch, completing the empty cohort as a no-op
- keep mixed Tailor/Cover/Apply material ownership separate from Apply selector semantics
- resolve global current-policy re-tailoring with its policy-aware selector before workflow dispatch
- route frozen cohorts through the existing bounded per-job fan-out, ownership fencing, and worker-wave timeout
- record the final chaos finding and completed live threshold reconciliation

## Why
A global material command left its job selection empty and therefore used the legacy unscoped batch runner. When its 30-minute activity timed out, abandoned work could finish late and race a newer selected owner. The same missing ownership distinction affected empty cohorts, mixed Apply workflows, and the global current-policy maintenance action.

## Validation
- `uv --project workers/automation run pytest workers/automation/tests/test_workflow_job_pipeline.py -q` (48 passed)
- `uv --project workers/automation run pytest workers/automation/tests/test_jsonrpc_handlers.py workers/automation/tests/test_actions.py -q` (102 passed)
- touched-file Ruff check
- `corepack pnpm docs:build` (9,651 links and redirects passed)
- `git diff --check`
- independent review and QA gates: PASS, no findings
- local Temporal, API, web, and worker stack restarted and healthy on `793e86949`

### DIFF
--- a/workers/automation/src/jobctrl/workflow_specs.py
+++ b/workers/automation/src/jobctrl/workflow_specs.py
@@ -113,6 +113,11 @@ def build_run_stage_workflow_spec(params: dict[str, Any]) -> WorkflowStartSpec:
             if apply_selector is not None and apply_selector.job_id is not None
             else _job_ids(params)
         ),
+        # ``coverJobIds`` is an internal key written only by the material
+        # cohort freeze above; ignore it on unresolved requests.
+        cover_job_ids=(
+            _job_ids(params, key="coverJobIds") if material_selection_resolved else ()
+        ),
         apply_selector_keys=apply_selector.keys if apply_selector else (),
         material_selection_resolved=material_selection_resolved,
         source_ids=_source_ids(params),
@@ -136,35 +141,62 @@ def _scope_global_material_batch(
     stages: list[str],
     tenant_id: str,
 ) -> tuple[dict[str, Any], bool]:
-    """Freeze a global Tailor/Cover pickup into the workflow input.
+    """Freeze a global Tailor/Cover pickup into per-stage workflow cohorts.
 
     Material activities need an exact cohort so their bounded per-job fan-out,
     cancellation fence, and scaled timeout apply. Leaving ``jobIds`` empty
     selects the legacy global runner, whose blocking work can outlive a Temporal
     timeout and write after a newer owner has taken over.
+
+    Each requested material stage freezes its own backlog when the command
+    starts, wherever the material stages sit in the run: the first material
+    stage's cohort rides in ``jobIds`` and a Cover stage that follows Tailor
+    keeps its independently frozen ``pending_cover`` backlog in
+    ``coverJobIds``. Queue growth inside the run flows through stage-output
+    unions instead of an unscoped fallback — Score's ``scoredJobIds`` join the
+    frozen Tailor cohort and Tailor's ``approvedJobIds`` join the frozen Cover
+    cohort. Enrich-led runs are not frozen here because the Enrich stage
+    already hands its enriched subset to every later preparation stage.
     """
 
-    if not stages or stages[0] not in {"tailor", "cover"}:
+    if not any(stage in {"tailor", "cover"} for stage in stages):
+        return params, False
+    if "enrich" in stages:
         return params, False
     if params.get("jobId") or params.get("jobIds"):
         return params, False
 
     from jobctrl.database import get_connection, get_jobs_by_stage
 
-    first_stage = stages[0]
-    jobs = get_jobs_by_stage(
-        conn=get_connection(),
-        stage="pending_tailor" if first_stage == "tailor" else "pending_cover",
-        min_score=int(params.get("minScore", 7)),
-        limit=int(params.get("limit", 0)),
-        retailor=bool(params.get("retailor", False)) if first_stage == "tailor" else False,
-    )
-    job_ids = [
-        str(canonical_job_id(str(job["job_id"])))
-        for job in jobs
-        if str(job.get("tenant_id") or tenant_id) == tenant_id
-    ]
-    return {**params, "jobIds": list(dict.fromkeys(job_ids))}, True
+    conn = get_connection()
+
+    def frozen_cohort(queue_stage: str, *, retailor: bool) -> list[str]:
+        jobs = get_jobs_by_stage(
+            conn=conn,
+            stage=queue_stage,
+            min_score=int(params.get("minScore", 7)),
+            limit=int(params.get("limit", 0)),
+            retailor=retailor,
+        )
+        return list(
+            dict.fromkeys(
+                str(canonical_job_id(str(job["job_id"])))
+                for job in jobs
+                if str(job.get("tenant_id") or tenant_id) == tenant_id
+            )
+        )
+
+    scoped = dict(params)
+    if "tailor" in stages:
+        scoped["jobIds"] = frozen_cohort(
+            "pending_tailor",
+            retailor=bool(params.get("retailor", False)),
+        )
+        if "cover" in stages:
+            scoped["coverJobIds"] = frozen_cohort("pending_cover", retailor=False)
+    else:
+        scoped["jobIds"] = frozen_cohort("pending_cover", retailor=False)
+    return scoped, True
 
 
 def build_pipeline_workflow_spec(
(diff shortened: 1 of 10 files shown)


workers/automation/src/jobctrl/workflow_specs.py
```diff
@@ -120,6 +130,43 @@ def build_run_stage_workflow_spec(params: dict[str, Any]) -> WorkflowStartSpec:
...
+        return params, False
+    if params.get("jobId") or params.get("jobIds"):
+        return params, False
+
+    from jobctrl.database import get_connection, get_jobs_by_stage
+
+    first_stage = stages[0]
+    jobs = get_jobs_by_stage(
+        conn=get_connection(),
+        stage="pending_tailor" if first_stage == "tailor" else "pending_cover",
```
ebarti: PR Reviewer - pr-763-reviewer

Severity: High
Gate: Fail
Evidence: `_scope_global_material_batch` freezes only `stages[0]`'s queue (`pending_tailor` for a `["tailor", "cover"]` run), but `material_selection_resolved` then governs both material stages in `JobPipelineWorkflow._execute_stages` (workers/automation/src/jobctrl/pipeline/workflow.py:228-234), and the cover stage runs over `approvedJobIds` derived from the tailor result (workflow.py:276-279). Before this PR, the cover stage of a global multi-stage run selected all `pending_cover` jobs via the legacy runner.
Reproduction: Reproduced with a direct `_execute_stages` call in the head worktree: `JobPipelineWorkflowInput(stages=["tailor", "cover"], material_selection_resolved=True, job_ids=())` returns `stages_completed == ["tailor", "cover"]` with zero `_execute_stage` dispatches — cover never consults `pending_cover`.

A global `run tailor cover` (CLI `jobctrl run tailor cover`, or the JSON-RPC `run` method with `stages`) regresses in two ways:

1. If `pending_tailor` is empty but `pending_cover` has eligible jobs (e.g. prior cover failures below the 5-attempt cap, or a run canceled between tailor and cover), the whole workflow now completes as a no-op and those jobs silently never get cover letters. Legacy behavior covered them.
2. If `pending_tailor` is non-empty, cover runs only over this run's approved tailor subset (`derived_preparation_job_ids`), so pre-existing `pending_cover` stragglers outside the tailor cohort are silently dropped.

In both cases the command reports success while skipping work it previously performed; the only recovery is knowing to run a bare `run cover` afterward.

Required fix: Freeze each requested material stage against its own queue — e.g. also freeze a `pending_cover` cohort at spec-build time for a trailing cover stage and union it with the approved tailor subset before cover dispatch (at minimum, do not no-op the cover stage based on the tailor queue's emptiness). Add a regression test for the empty-tailor/non-empty-cover multi-stage case. If the narrowing in (2) is intentionally accepted, state it explicitly in docs/local-reliability-qa.md.

workers/automation/src/jobctrl/workflow_specs.py
```diff
@@ -120,6 +130,43 @@ def build_run_stage_workflow_spec(params: dict[str, Any]) -> WorkflowStartSpec:
...
+) -> tuple[dict[str, Any], bool]:
+    """Freeze a global Tailor/Cover pickup into the workflow input.
+
+    Material activities need an exact cohort so their bounded per-job fan-out,
+    cancellation fence, and scaled timeout apply. Leaving ``jobIds`` empty
+    selects the legacy global runner, whose blocking work can outlive a Temporal
+    timeout and write after a newer owner has taken over.
+    """
+
+    if not stages or stages[0] not in {"tailor", "cover"}:
```
ebarti: PR Reviewer - pr-763-reviewer

Severity: Medium
Gate: Fail
Evidence: The freeze triggers only when `stages[0]` is a material stage. For a global `["score", "tailor", "cover"]` run (`MAINTENANCE_STAGE_ORDER` in workers/automation/src/jobctrl/pipeline/runner.py:78, reachable via `jobctrl run score tailor cover` and the JSON-RPC `run` method), `_execute_stages` derives no cohort from the score result — only enrich and tailor set `derived_preparation_job_ids` (workflow.py:272-279) — so tailor dispatches with empty `job_ids` and `material_selection_resolved=False`, and `tailor_activity` falls through to the legacy unscoped `_run_stage_observed("tailor", _run_tailor, ...)` batch runner (workers/automation/src/jobctrl/materials/activities.py:170).
Reproduction: Reasoned from code; the fallthrough is unconditional once `job_ids` is empty and neither `current_policy_only` nor `material_selection_resolved` is set.

The abandoned-timeout ownership race this PR fixes (chaos finding 20: a 30-minute activity timeout abandons the unscoped batch, whose late writes race a newer selected owner) therefore remains reachable through score-first global maintenance runs. The PR body and the docs/local-reliability-qa.md addition state that global material pickup now always uses the bounded per-job fan-out and ownership fence, which overstates the coverage.

Required fix: Either close the path (e.g. have the score stage return its downstream-eligible job ids so trailing tailor/cover run scoped, mirroring `enrichedJobIds`), or explicitly record the residual unscoped path in docs/local-reliability-qa.md and the chaos plan so the gap is tracked rather than implied closed.

ebarti (commented):
PR Reviewer - pr-763-reviewer

Reviewed head a7f6ec0 (`fix/material-batch-ownership`) against `fix/cover-skill-grounding`.

The freeze mechanism is sound and well-tested for single-stage global commands, the empty-cohort no-op, mixed batch-Apply semantics, and the frozen current-policy path. I re-ran the new tests in a clean worktree at the head SHA: 5 passed in `test_workflow_job_pipeline.py` (freeze/no-op/apply-selector), 1 passed in `test_jsonrpc_handlers.py`.

Two findings:

- **High** — a multi-stage global `run tailor cover` freezes only the `pending_tailor` queue, so pre-existing `pending_cover` jobs are silently dropped, and an empty tailor cohort no-ops the cover stage entirely (reproduced with a direct `_execute_stages` call).
- **Medium** — score-first global runs (`run score tailor cover`) never trigger the freeze, so the legacy unscoped material batch and its abandoned-timeout ownership race remain reachable.

Counts: Blocker=0, High=1, Medium=1, Low=0.

workers/automation/src/jobctrl/workflow_specs.py
ebarti: Fixed in 5955d22354d3fdbe7657bfee7080df19064fb283.

Each requested material stage now freezes its own backlog at command start: the first material cohort stays in `jobIds` and a Cover stage that follows Tailor keeps an independently frozen `pending_cover` cohort (`coverJobIds` / `cover_job_ids`). Cover dispatches the union of its frozen backlog and the subset Tailor approved in this run, so (a) an empty Tailor cohort no longer no-ops a non-empty Cover backlog and (b) pre-existing `pending_cover` stragglers are no longer dropped. The spend gate counts a non-empty frozen Cover cohort. Regressions: `test_empty_tailor_cohort_still_covers_frozen_cover_backlog`, `test_cover_unions_frozen_backlog_with_approved_tailor_subset`, `test_multi_stage_global_material_run_freezes_each_queue_independently`, `test_resolved_empty_multi_stage_material_run_is_a_no_op`; the PR's existing freeze/no-op/apply tests are unchanged and green.

workers/automation/src/jobctrl/workflow_specs.py
ebarti: Fixed in 5955d22354d3fdbe7657bfee7080df19064fb283.

The freeze now triggers whenever Tailor/Cover appear anywhere in a global run, so `run score tailor cover` no longer reaches the legacy unscoped material runner. Semantics for cohorts affected by an earlier stage: within a run a material queue only grows through upstream stage output, so Score's newly scored ids (`scoredJobIds`, now reported by `run_scoring` through the score stage result) join the frozen Tailor cohort and Tailor's `approvedJobIds` join the frozen Cover cohort — newly qualified jobs are still processed in the same command. Score itself keeps its legacy global sweep with an explicitly unscoped selection, and Enrich-led runs keep the existing enriched-subset handoff (documented in `_scope_global_material_batch`). The overstated text in docs/local-reliability-qa.md and chaos finding 20 now describes this behavior. Regressions: `test_score_first_global_run_freezes_material_cohorts`, `test_score_first_resolved_run_keeps_score_unscoped_and_feeds_tailor`, plus `scoredJobIds` assertions in test_scorer.py.

PR #765, opened 2026-08-07
fix(ci): stop API schema seeding from timing out its hook

ebarti:
## What changed

- `apps/api/test/v7-schema.ts` — build the exact-v7 database **once per worker process** and copy the closed file, instead of re-executing 2,651 lines of DDL on every seeding call.
- `apps/api/vitest.config.ts` — `hookTimeout: 15_000`, matching the existing `testTimeout`.
- `.github/workflows/typescript.yml` — add `workers/**` and `scripts/**` to the **`push`** path filter only.
- `apps/api/test/typescript-ci-workflow-contract.test.ts` — two trigger tests.

## Why: the #749 failure

#749 deletes only markdown, and TypeScript CI failed on it. Re-running the identical SHA passed — 55/55 in 92s wall / 210s tests, against 679s / 1571s for the failure. A 7.5× runner swing on unchanged code.

The `beforeEach` in `apps/api/test/server.test.ts:117` re-executed the full `schema_v7.sql` — 110 tables, 108 indexes, 24 triggers — once per test, across 30 test files. Vitest defaults `hookTimeout` to 10s while this project sets `testTimeout` to 15s, so the setup hook carried a *tighter* deadline than the tests it existed to set up, and it was the first thing to blow on a loaded runner.

`schema_v7.sql` is pure DDL: no seed rows, no PRAGMAs. So the built database is a single self-contained file and a byte copy is indistinguishable from a fresh build, `user_version` included. Building it once and copying takes the local full-suite run from **174.38s to 33.72s** of test time (~72ms saved per hook). Reaching the 10s hook deadline now requires roughly a 140× slowdown rather than a 7.5× one.

Measurement note: I never instrumented the hook directly. The defensible number is that controlled local A/B, not a per-hook profile.

## Correction: the trigger scoping in this PR was wrong

The first commit on this branch added a `paths:` filter to `pull_request`. **That was a mistake and the third commit reverts it.**

`scripts/stacked-ci-workflows.test.mjs:44` requires `typescript.yml` to have `on.pull_request == null`, and PR #644 deleted precisely this filter after #640 introduced it. [GitHub's stacked-pull-request documentation](https://docs.github.com/en/pull-requests/how-tos/merge-and-close-pull-requests/optimizing-ci-for-stacked-pull-requests) gives the reason: a workflow must instantiate for **every** layer of a stack so each layer produces a check record, and cost is meant to be gated on a job-level `if:` over `stack.position` — not on the trigger. A filtered-out workflow reports *nothing*, where a job-gated one reports a skip.

CI did not catch this here, only because `launcher.yml`'s own path filter does not match this changeset. Running `node --test scripts/stacked-ci-workflows.test.mjs` on the pre-revert commit fails assertion 1; it passes now.

**The `push` filter additions are kept**, because they are a real and separate gap that the stacked-PR contract does not constrain. `workers/**` supplies `schema_v7.sql`, `schema_manifest.py`, and `config.py` to the cross-runtime parity and schema-guard tests; `scripts/**` supplies the install script the install contract test executes. Both sit outside `apps/` and `packages/`, so a push touching either skipped the suite covering it.

## Validation

- `pnpm --filter @jobctrl/api check` → clean.
- `pnpm --filter @jobctrl/api test` → **55 files, 742 tests passed**, 9.40s wall / 41.46s tests.
- `git diff --check` clean.
- **Three mutation tests** against the committed baseline, each firing its own assertion: adding a `paths` filter to `pull_request`, adding `branches:` to `pull_request`, and dropping `workers/**` from the `push` list.

### DIFF
--- a/apps/api/test/typescript-ci-workflow-contract.test.ts
+++ b/apps/api/test/typescript-ci-workflow-contract.test.ts
@@ -40,7 +40,15 @@ describe("TypeScript CI trigger contract", () => {
 
   it("leaves pull requests entirely unfiltered", () => {
     const block = onBlock();
-    const pullRequest = block.slice(block.indexOf("\n  pull_request:"));
+    const triggerStart = block.indexOf("\n  pull_request:");
+
+    // Deleting the trigger outright is the strongest violation of this
+    // invariant — the workflow would stop instantiating for pull requests
+    // altogether — so an absent trigger must fail here, not pass vacuously
+    // (indexOf would return -1, slicing the block's last character, and
+    // triggerPaths would report an absent trigger as "no filter").
+    expect(triggerStart).toBeGreaterThanOrEqual(0);
+    const pullRequest = block.slice(triggerStart);
 
     // Every layer of a GitHub stack must instantiate this workflow so that layer
     // gets a check record; a filtered-out workflow reports nothing at all rather
--- a/apps/api/test/typescript-ci-workflow-contract.test.ts
+++ b/apps/api/test/typescript-ci-workflow-contract.test.ts
@@ -4,6 +4,79 @@ import { describe, expect, it } from "vitest";
 
 const workflowPath = new URL("../../../.github/workflows/typescript.yml", import.meta.url);
 
+/**
+ * Collects the `paths` filter of a single `on:` trigger without a YAML parser,
+ * which `apps/api` does not depend on. Returns an empty list when the trigger is
+ * absent or carries no filter, so a bare `pull_request:` fails the contract.
+ */
+function triggerPaths(onBlock: string, trigger: string): string[] {
+  const triggerStart = onBlock.indexOf(`\n  ${trigger}:`);
+  if (triggerStart < 0) {
+    return [];
+  }
+  const rest = onBlock.slice(triggerStart + 1);
+  const nextTrigger = rest.slice(1).search(/\n {2}\S/);
+  const section = nextTrigger === -1 ? rest : rest.slice(0, nextTrigger + 1);
+  const pathsStart = section.indexOf("\n    paths:");
+  if (pathsStart < 0) {
+    return [];
+  }
+  const entries: string[] = [];
+  for (const line of section.slice(pathsStart + 1).split("\n").slice(1)) {
+    const entry = /^ {6}- "(.+)"$/.exec(line)?.[1];
+    if (entry === undefined) {
+      break;
+    }
+    entries.push(entry);
+  }
+  return entries;
+}
+
+describe("TypeScript CI trigger contract", () => {
+  const onBlock = () => {
+    const workflow = readFileSync(workflowPath, "utf8");
+    return workflow.slice(workflow.indexOf("\non:"), workflow.indexOf("\njobs:"));
+  };
+
+  it("filters pull requests to the same paths as pushes", () => {
+    const block = onBlock();
+    const pullRequestPaths = triggerPaths(block, "pull_request");
+
+    // A bare `pull_request:` trigger ran the full suite - Playwright E2E and
+    // Storybook included - on documentation-only pull requests, exposing them to
+    // failures they could not have caused.
+    expect(pullRequestPaths).not.toHaveLength(0);
+    expect(pullRequestPaths).toEqual(triggerPaths(block, "push"));
+  });
+
+  it("runs whenever a file the suite reads can change", () => {
+    const pullRequestPaths = triggerPaths(onBlock(), "pull_request");
+
+    // `workers/**` supplies schema_v7.sql, schema_manifest.py, and config.py to
+    // the cross-runtime parity and schema-guard tests; `scripts/**` supplies the
+    // install script the install contract test executes. Dropping either would
+    // let a change that breaks this suite merge without ever running it.
+    for (const required of [
+      "apps/**",
+      "packages/**",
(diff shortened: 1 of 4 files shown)


ebarti:
**Converted to draft: this PR violates an existing repository contract.**

`scripts/stacked-ci-workflows.test.mjs:44` asserts that `python.yml` and `typescript.yml` must both have `on.pull_request == null` — no `branches` filter and **no `paths` filter**:

```
python must not suppress stacked PRs by their direct base branch or changed paths
```

That contract is deliberate and was built over four PRs (#640, #641, #642, #644). The design is *instantiate always, gate at the job level*: the workflow triggers on every pull request so every stacked layer produces a check-run record, and the job-level `if:` (`trustedTopOfStack`) decides whether to actually spend the runner. A `paths` filter defeats it, because a workflow that never triggers reports nothing at all rather than reporting a skip.

CI caught this on #766 (`Darwin launcher (vet + race)` → `not ok 122`). It did **not** catch it on #765 only because `launcher.yml`'s own path filter does not match that changeset; running `node --test scripts/stacked-ci-workflows.test.mjs` on the #765 branch fails the same assertion.

Held pending a decision on the right approach. The path-audit work in this PR (which directories the suite genuinely reads) stays valid regardless — it just needs to be applied at the job level rather than the trigger.

apps/api/test/typescript-ci-workflow-contract.test.ts
```diff
@@ -4,6 +4,74 @@ import { describe, expect, it } from "vitest";
...
+    const pullRequest = block.slice(block.indexOf("\n  pull_request:"));
+
+    // Every layer of a GitHub stack must instantiate this workflow so that layer
+    // gets a check record; a filtered-out workflow reports nothing at all rather
+    // than reporting a skip. Cost belongs at the job level, on the trusted
+    // cumulative-head `if:`, which is the pattern GitHub documents for stacks.
+    // scripts/stacked-ci-workflows.test.mjs asserts the same invariant by
+    // parsing the YAML; this keeps it visible to the suite that owns the file.
+    expect(pullRequest).not.toContain("branches:");
+    expect(triggerPaths(block, "pull_request")).toHaveLength(0);
```
ebarti, line 60: PR Reviewer - pr-review-worker

Severity: Medium
Gate: Pass
Evidence: `block.indexOf("\n  pull_request:")` returns -1 when the trigger is absent, so `block.slice(-1)` yields the block's last character and `.not.toContain("branches:")` passes; `triggerPaths` documents "Returns an empty list when the trigger is absent" and returns `[]`, so `toHaveLength(0)` passes too. Deleting the trigger is the strongest violation of the invariant this test guards, and it is also the one no other gate catches on the changesets that matter: launcher.yml's `pull_request`/`push` paths do not include `.github/workflows/typescript.yml`, so `scripts/stacked-ci-workflows.test.mjs` (whose strict `assert.equal(on.pull_request, null)` would fail on `undefined`) never runs for a typescript.yml-only change — and post-merge, the push-triggered run of this very test passes vacuously.
Reproduction: Reproduced with a scratch harness replicating both test bodies against the head workflow file — mutations adding `paths:`/`branches:` to `pull_request` and dropping `workers/**` from `push` all FAIL as intended, but deleting the `  pull_request:` line entirely PASSES both tests.

The test named "leaves pull requests entirely unfiltered" is satisfiable by removing the `pull_request` trigger altogether, which would silently stop instantiating TypeScript CI for every PR layer — the exact outcome this contract exists to prevent.

Required fix: Assert the trigger exists before asserting it is unfiltered, e.g. capture `const start = block.indexOf("\n  pull_request:")` and `expect(start).toBeGreaterThanOrEqual(0)` (or `expect(block).toContain("\n  pull_request:")`) before slicing.

ebarti (commented):
Reviewed the diff at 0c6e2df against main (946d6ef). Verified locally: `pnpm --filter @jobctrl/api check` clean; `pnpm --filter @jobctrl/api test` 55 files / 742 tests passed (8.9s wall, 39.2s tests); `node --test scripts/stacked-ci-workflows.test.mjs` 4/4 passed at head, confirming the pre-revert contract violation is gone. Also re-ran the three claimed mutations against the committed assertions — all caught. The v7 template-copy approach is sound: schema_v7.sql verified pure DDL (no PRAGMAs, no inserts), every caller seeds into a fresh mkdtemp dir, and under Vitest 4 forks isolation the exit-handler/tmpdir lifecycle is once per short-lived process.

One Medium finding (inline): the new unfiltered-pull-request contract test passes vacuously when the `pull_request:` trigger is deleted outright — the one violation none of the other gates would catch, given launcher.yml's path filters exclude `.github/workflows/typescript.yml`. No Blocker/High findings.

apps/api/test/typescript-ci-workflow-contract.test.ts
ebarti: Fixed in 5bf76bde5. The test now captures `const triggerStart = block.indexOf("\n  pull_request:")` and asserts `expect(triggerStart).toBeGreaterThanOrEqual(0)` before slicing, so deleting the trigger fails the test instead of passing vacuously. Mutation-checked: removing the `pull_request:` line from typescript.yml fails with `expected -1 to be greater than or equal to 0`; restored baseline passes 3/3, `stacked-ci-workflows.test.mjs` still 4/4, api `check` clean.

PR #766, opened 2026-08-07
fix(ci): run Python CI on pushes that touch the files its suite reads

ebarti:
## What changed

- `.github/workflows/python.yml` — add `packages/**`, `packaging/**`, `.github/workflows/**`, the outreach sources scanned by the INV-1 guard (`apps/web/src/contexts/outreach/**`, `apps/web/src/views/outreach/**`, `apps/api/src/outreach.ts`), and root `package.json`, `README.md`, `LICENSE`, `NOTICE` to the **`push`** path filter. `pull_request` stays completely unfiltered.
- `scripts/check-python-workflow-contract.mjs` — parse the workflow as real YAML (Ruby `YAML.safe_load`, the same mechanism as `scripts/stacked-ci-workflows.test.mjs`; python.yml's contract step installs no Node dependencies, so js-yaml is not an option there) and assert `on.pull_request` is exactly `null` — no `branches`, no `paths`, in any YAML spelling — plus assert the `push` list covers every directory and file the suite reads.

## Why

The Python suite reads real files from outside `workers/` and `scripts/`, and the `push` filter did not list them:

| Path | Read by |
| --- | --- |
| `packages/**` | 11 parity tests resolve `packages/domain-types/test/fixtures/…` via `parents[3]`; `test_domain_event_parity.py:22` reads `packages/domain-types/src/events` |
| `packaging/**` | `test_provider_packs.py:145,446` reads `provider-packs.lock.json`; `test_doctor_playwright.py:23` reads `capability-policy.json` |
| `.github/workflows/**` | `test_release_check.py:1195` reads `release-check.yml` |
| `apps/web/src/contexts/outreach/**`, `apps/web/src/views/outreach/**`, `apps/api/src/outreach.ts` | `test_outreach_no_send_transport.py:22-31` scans these sources for transport-shaped symbols — the only cross-runtime source-level INV-1 enforcement |
| `package.json`, `README.md` (root) | `test_release_check.py:20-30` reads both through `release_check.ROOT` |
| `LICENSE`, `NOTICE` | `test_python_package_attribution.py:18-19` reads both verbatim |

A push to `main` touching only one of these could break the suite — or, for the outreach paths, land an INV-1 violation — without Python CI ever running.

**Correction (review follow-up):** an earlier revision of this description claimed `apps/**` was "deliberately excluded" because `test_release_check.py:189,196` only passes `apps/web/public/demo/*.pdf` paths to `scan_name()` as strings. That statement was true of `test_release_check.py` but the audit behind it was incomplete: `test_outreach_no_send_transport.py` does read apps outreach sources. The three outreach paths are now in the filter; the rest of `apps/**` remains excluded because nothing else in the suite reads it.

## Correction: the trigger scoping in the first commit was wrong

The first commit also applied this list to `pull_request`. **That was a mistake and the second commit reverts it.**

`scripts/stacked-ci-workflows.test.mjs:44` requires `python.yml` to have `on.pull_request == null`, and PR #644 deleted precisely this filter after #640 introduced it. [GitHub's stacked-pull-request documentation](https://docs.github.com/en/pull-requests/how-tos/merge-and-close-pull-requests/optimizing-ci-for-stacked-pull-requests) gives the reason: a workflow must instantiate for **every** layer of a stack so each layer produces a check record, and cost is meant to be gated on a job-level `if:` over `stack.position` — not on the trigger. A filtered-out workflow reports *nothing*, where a job-gated one reports a skip.

CI caught it: `Darwin launcher (vet + race)` → `not ok 122 - stacked pull requests always instantiate the Python and TypeScript admission workflows`. It now passes.

`push` filters are unconstrained by that contract, so the genuine gap above is still fixed.

## Correction: the first contract check was format-dependent (review follow-up)

The initial `triggerPaths` helper string-matched one canonical spelling (`\n    paths:` + six-space quoted entries). Flow-style `paths: ["workers/**"]` or unquoted block entries under `pull_request:` parsed as zero entries and false-passed, even though GitHub would honor the filter — and this script is the sole PR-time guard for python.yml-only changes (launcher.yml's filter does not match them). The third commit replaces the string matching with a real YAML parse and a strict `on.pull_request === null` assertion, which also fails when the trigger is absent entirely.

## Validation

- `node scripts/check-python-workflow-contract.mjs` → exit 0.
- `python3 scripts/release_check.py` → `Release check passed.`
- Ruby `YAML.safe_load` diff of the workflow before/after the third commit: `pull_request` and `workflow_dispatch` still `null`, `push.branches` unchanged, `push.paths` gained exactly the 7 new entries, all jobs byte-identical.
- `git diff --check` clean.
- **Eight mutation tests** against the contract check, each run on a sandboxed copy: flow-style `paths` under `pull_request` → fail; unquoted block `paths` under `pull_request` → fail; canonical quoted `paths` under `pull_request` → fail; `pull_request` trigger absent → fail; `branches:` under `pull_request` → fail; dropping `apps/api/src/outreach.ts` from the `push` list → fail; unmutated baseline → pass; `push.paths` rewritten flow-style with the same entries → pass.

### DIFF
--- a/.github/workflows/python.yml
+++ b/.github/workflows/python.yml
@@ -1,14 +1,31 @@
 name: Python CI
 
+# The Python suite reads real files from outside workers/ and scripts/:
+# `packages/**` holds the shared cross-runtime parity fixtures and the
+# domain-event sources, `packaging/**` holds provider-packs.lock.json and
+# capability-policy.json, and the release-scan tests read workflow files under
+# `.github/workflows/**`. Both lists must stay in sync and must keep those
+# directories, or a change that breaks Python CI could merge without ever
+# running it. No `branches` filter on `pull_request`: stacked PRs target their
+# parent branch, not main.
 on:
   push:
     branches: ["main"]
     paths:
       - "workers/**"
       - "scripts/**"
+      - "packages/**"
+      - "packaging/**"
       - "pyproject.toml"
-      - ".github/workflows/python.yml"
+      - ".github/workflows/**"
   pull_request:
+    paths:
+      - "workers/**"
+      - "scripts/**"
+      - "packages/**"
+      - "packaging/**"
+      - "pyproject.toml"
+      - ".github/workflows/**"
   workflow_dispatch:
 
 jobs:
--- a/.github/workflows/python.yml
+++ b/.github/workflows/python.yml
@@ -3,7 +3,10 @@ name: Python CI
 # `push` is filtered to the files this suite actually reads. Beyond workers/ and
 # scripts/, `packages/**` holds the shared cross-runtime parity fixtures and the
 # domain-event sources, `packaging/**` holds provider-packs.lock.json and
-# capability-policy.json, and the release-scan tests read workflow files under
+# capability-policy.json, the INV-1 no-send-transport guard
+# (test_outreach_no_send_transport.py) scans the TypeScript outreach sources
+# under apps/, the byline/attribution tests read root package.json, README.md,
+# LICENSE, and NOTICE, and the release-scan tests read workflow files under
 # `.github/workflows/**`. A push that changed any of those without these entries
 # would skip the suite covering it.
 #
@@ -22,6 +25,13 @@ on:
       - "scripts/**"
       - "packages/**"
       - "packaging/**"
+      - "apps/web/src/contexts/outreach/**"
+      - "apps/web/src/views/outreach/**"
+      - "apps/api/src/outreach.ts"
+      - "package.json"
+      - "README.md"
+      - "LICENSE"
+      - "NOTICE"
       - "pyproject.toml"
       - ".github/workflows/**"
   pull_request:
--- a/.github/workflows/python.yml
+++ b/.github/workflows/python.yml
@@ -1,13 +1,19 @@
 name: Python CI
 
-# The Python suite reads real files from outside workers/ and scripts/:
-# `packages/**` holds the shared cross-runtime parity fixtures and the
+# `push` is filtered to the files this suite actually reads. Beyond workers/ and
+# scripts/, `packages/**` holds the shared cross-runtime parity fixtures and the
 # domain-event sources, `packaging/**` holds provider-packs.lock.json and
 # capability-policy.json, and the release-scan tests read workflow files under
-# `.github/workflows/**`. Both lists must stay in sync and must keep those
-# directories, or a change that breaks Python CI could merge without ever
-# running it. No `branches` filter on `pull_request`: stacked PRs target their
-# parent branch, not main.
+# `.github/workflows/**`. A push that changed any of those without these entries
+# would skip the suite covering it.
+#
+# `pull_request` deliberately takes no filter at all — no `branches`, no
+# `paths`. Every layer of a GitHub stack must instantiate this workflow so it
+# produces a check record for that layer; a filtered-out workflow reports
+# nothing rather than reporting a skip. Cost is controlled at the job level
+# instead, by the `if:` below that admits only a trusted cumulative head, which
+# is the pattern GitHub documents for stacked pull requests. Enforced by
+# scripts/stacked-ci-workflows.test.mjs — do not add a filter here.
 on:
   push:
     branches: ["main"]
@@ -19,13 +25,6 @@ on:
       - "pyproject.toml"
       - ".github/workflows/**"
   pull_request:
-    paths:
-      - "workers/**"
-      - "scripts/**"
-      - "packages/**"
-      - "packaging/**"
-      - "pyproject.toml"
-      - ".github/workflows/**"
   workflow_dispatch:
 
 jobs:
--- a/scripts/check-python-workflow-contract.mjs
+++ b/scripts/check-python-workflow-contract.mjs
@@ -1,66 +1,71 @@
 import assert from "node:assert/strict";
+import { execFile } from "node:child_process";
 import { readFile } from "node:fs/promises";
+import { fileURLToPath } from "node:url";
+import { promisify } from "node:util";
 
-const workflow = await readFile(new URL("../.github/workflows/python.yml", import.meta.url), "utf8");
+const workflowPath = fileURLToPath(new URL("../.github/workflows/python.yml", import.meta.url));
+const workflow = await readFile(workflowPath, "utf8");
 
-const onBlock = workflow.slice(workflow.indexOf("\non:"), workflow.indexOf("\njobs:"));
-
-/**
- * Collects the `paths` filter of a single `on:` trigger. Returns an empty list
- * when the trigger is absent or carries no filter, so a bare `pull_request:`
- * fails the contract.
- */
-function triggerPaths(trigger) {
-  const triggerStart = onBlock.indexOf(`\n  ${trigger}:`);
-  if (triggerStart < 0) {
-    return [];
-  }
-  const rest = onBlock.slice(triggerStart + 1);
-  const nextTrigger = rest.slice(1).search(/\n {2}\S/);
-  const section = nextTrigger === -1 ? rest : rest.slice(0, nextTrigger + 1);
-  const pathsStart = section.indexOf("\n    paths:");
-  if (pathsStart < 0) {
-    return [];
-  }
-  const entries = [];
-  for (const line of section.slice(pathsStart + 1).split("\n").slice(1)) {
-    const entry = /^ {6}- "(.+)"$/.exec(line)?.[1];
-    if (entry === undefined) {
-      break;
-    }
-    entries.push(entry);
-  }
-  return entries;
-}
+// Parse the real YAML the same way scripts/stacked-ci-workflows.test.mjs does
+// (Ruby ships on the hosted runners; this step installs no Node dependencies),
+// so the trigger contract holds for what GitHub will actually evaluate rather
+// than for one canonical spelling. A string match false-passes on flow-style
+// lists, unquoted entries, or reordered keys.
+const execFileAsync = promisify(execFile);
+const rubyYamlToJson = String.raw`
+document = YAML.safe_load(
+  File.read(ARGV.fetch(0)),
+  permitted_classes: [],
+  permitted_symbols: [],
+  aliases: false,
+)
+events = document.delete("on") || document.delete(true)
+document["on"] = events
+puts JSON.generate(document)
+`;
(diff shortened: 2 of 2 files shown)


ebarti:
**Converted to draft: this PR violates an existing repository contract.**

`scripts/stacked-ci-workflows.test.mjs:44` asserts that `python.yml` and `typescript.yml` must both have `on.pull_request == null` — no `branches` filter and **no `paths` filter**:

```
python must not suppress stacked PRs by their direct base branch or changed paths
```

That contract is deliberate and was built over four PRs (#640, #641, #642, #644). The design is *instantiate always, gate at the job level*: the workflow triggers on every pull request so every stacked layer produces a check-run record, and the job-level `if:` (`trustedTopOfStack`) decides whether to actually spend the runner. A `paths` filter defeats it, because a workflow that never triggers reports nothing at all rather than reporting a skip.

CI caught this on #766 (`Darwin launcher (vet + race)` → `not ok 122`). It did **not** catch it on #765 only because `launcher.yml`'s own path filter does not match that changeset; running `node --test scripts/stacked-ci-workflows.test.mjs` on the #765 branch fails the same assertion.

Held pending a decision on the right approach. The path-audit work in this PR (which directories the suite genuinely reads) stays valid regardless — it just needs to be applied at the job level rather than the trigger.

.github/workflows/python.yml
```diff
@@ -1,13 +1,29 @@
...
   push:
     branches: ["main"]
     paths:
       - "workers/**"
       - "scripts/**"
+      - "packages/**"
+      - "packaging/**"
       - "pyproject.toml"
-      - ".github/workflows/python.yml"
+      - ".github/workflows/**"
```
ebarti, line 36: PR Reviewer - pr-766-reviewer

Severity: Medium
Gate: Pass
Evidence: workers/automation/tests/test_outreach_no_send_transport.py:22-32 scans real source files under apps/web/src/contexts/outreach/**, apps/web/src/views/outreach/**, and apps/api/src/outreach.ts (the INV-1 no-send-transport guard). None of these match the push filter. typescript.yml does run on apps/** pushes, but its INV-1 coverage (apps/api/test/outreach.test.ts:462) only asserts the /send route 404s -- it never scans sources for transport symbols, so this Python test is the only cross-runtime source-level enforcement.
Reproduction: Reasoned from code; grepped the suite for repo-root reads (parents[3]/parents[1] resolution).

The PR body says "apps/** is deliberately excluded" based on test_release_check.py:189,196 passing demo-PDF paths as strings, but that audit missed test_outreach_no_send_transport.py. A push to main touching only apps outreach code (e.g. introducing a nodemailer/sendMail usage) skips Python CI and lands an INV-1 violation ungated -- the exact gap class this PR exists to close. Lesser instances of the same class: test_release_check.py:185-196 reads root package.json and README.md (in practice redundantly covered because release-check.yml push-triggers unfiltered), and test_python_package_attribution.py:22-23 reads root LICENSE and NOTICE (content stays self-consistent via the pyproject force-include, but deletion/rename would break the suite with no CI run).

Required fix: add "apps/web/src/contexts/outreach/**", "apps/web/src/views/outreach/**", and "apps/api/src/outreach.ts" (or "apps/**") to the push paths and to the contract script's required list; make an explicit call on LICENSE/NOTICE/package.json/README.md and correct the PR body's apps/** rationale.

scripts/check-python-workflow-contract.mjs
```diff
@@ -2,6 +2,73 @@ import assert from "node:assert/strict";
...
+}
+
+// Every layer of a GitHub stack must instantiate this workflow so that layer
+// gets a check record; a filtered-out workflow reports nothing at all rather
+// than reporting a skip. Cost belongs on the job-level trusted-cumulative-head
+// `if:`, which is the pattern GitHub documents for stacks.
+// scripts/stacked-ci-workflows.test.mjs asserts the same invariant by parsing
+// the YAML; this keeps it enforced by the workflow's own contract step too.
+assert.equal(
+  triggerPaths("pull_request").length,
```
ebarti: PR Reviewer - pr-766-reviewer

Severity: Medium
Gate: Pass
Evidence: triggerPaths only recognizes entries matching /^ {6}- "(.+)"$/ beneath a "\n    paths:" block; any other valid YAML spelling parses as zero entries, so this assertion passes while a real filter exists. launcher.yml's path filter (scripts/**, launcher/**, ...) does not include .github/workflows/python.yml, so a python.yml-only PR never runs the YAML-parsing stacked-ci-workflows.test.mjs -- this script is the sole PR-time guard in exactly that scenario (the same mechanism that let #765 slip, per the draft-conversion comment on this PR).
Reproduction: Reproduced with commands -- copied .github/ + scripts/ to a temp dir and inserted, under pull_request:, first `paths: ["workers/**"]` (flow style) and then an unquoted block entry (`- workers/**`). node check-python-workflow-contract.mjs exited 0 both times, while Ruby YAML.safe_load shows on.pull_request == {"paths"=>["workers/**"]} (non-null, i.e. the stacked test would have failed it).

The new pull_request assertion false-passes on non-canonical formats, defeating its purpose as the last line of defense for python.yml-only edits. Secondary: the triggerPaths doc comment says a bare `pull_request:` "fails the contract" -- inverted; returning [] is exactly what makes this assertion pass. A fully absent pull_request: trigger also passes both new assertions (indexOf -1 slices to the last character), leaving only the stacked test's strict null check to catch it, when it runs.

Required fix: make the check format-independent -- e.g. assert.doesNotMatch on the pull_request section for /paths:/, mirroring the branches: check (or parse real YAML as stacked-ci-workflows.test.mjs does), and fix the inverted doc comment.

ebarti (commented):
Reviewed 73f39a88fa184e2c7891771b5763199ec7356661 against main. The push-filter gap this PR fixes is real and every cited read verified: packages/** (test_domain_event_parity.py:22 plus the parents[3] fixture parity tests), packaging/** (test_provider_packs.py:145,446; test_doctor_playwright.py:23), .github/workflows/** (test_release_check.py:1195 and the HOMEBREW/RELEASE workflow-path reads at test_release_check.py:710-1015). pull_request stays null, and both enforcement layers pass locally: node scripts/check-python-workflow-contract.mjs exit 0; node --test scripts/stacked-ci-workflows.test.mjs 4/4.

Two Medium findings, no Blocker/High:
1. The path audit missed real suite reads under apps/** (the INV-1 outreach source scan) and root files, and the PR body's "apps/** deliberately excluded" claim only audited test_release_check.
2. The new contract assertion is format-dependent and demonstrably false-passes on flow-style or unquoted pull_request paths filters -- in the exact python.yml-only scenario where the stacked YAML test does not run.

.github/workflows/python.yml
ebarti: Fixed in 92a659cbb.

- Added `apps/web/src/contexts/outreach/**`, `apps/web/src/views/outreach/**`, and `apps/api/src/outreach.ts` to the push filter and to the contract script's required list — the specific paths rather than `apps/**`, matching the filter's files-the-suite-reads philosophy (nothing else under apps/ is read).
- Explicit call on the lesser reads: **all four added** (`package.json`, `README.md`, `LICENSE`, `NOTICE`), each verified as a real read at head (test_release_check.py:20-30 via release_check.ROOT; test_python_package_attribution.py:18-19). The release-check.yml redundancy only covers the script run, not the pytest assertions, and LICENSE/NOTICE deletion/rename had no trigger at all. Swept every repo-root join in workers/automation/tests — all remaining ones target packages/ or packaging/, so this is now the complete set.
- PR body corrected: the "apps/** is deliberately excluded" rationale is replaced with the accurate account (release_check only handles demo-PDF strings, but the INV-1 guard does read apps outreach sources).

Verified: `node scripts/check-python-workflow-contract.mjs` exit 0; Ruby YAML diff shows push.paths gained exactly the 7 entries with pull_request/workflow_dispatch/jobs untouched; mutation dropping `apps/api/src/outreach.ts` now fails the contract check.

scripts/check-python-workflow-contract.mjs
ebarti: Fixed in 92a659cbb.

Replaced the string-matching `triggerPaths` helper with a real YAML parse — Ruby `YAML.safe_load` via execFile, the exact mechanism of scripts/stacked-ci-workflows.test.mjs (js-yaml is only a pnpm override here, and python.yml's contract step installs no Node dependencies, so importing it would break the CI step; Ruby ships on the hosted runners). The check now asserts `parsed.on.pull_request === null` (strict), so any filter in any YAML spelling — and an absent trigger, since `undefined !== null` — fails. The inverted doc comment is gone with the helper. The push-paths assertion reads `on.push.paths` from the parsed document, so it is format-independent in both directions.

Mutation matrix on sandboxed copies (script + workflow copied to a temp tree, workflow mutated, node run):
- flow-style `paths: ["workers/**"]` under pull_request → exit 1
- unquoted block entry under pull_request → exit 1
- canonical quoted `paths:` under pull_request → exit 1
- pull_request trigger absent → exit 1
- `branches:` under pull_request → exit 1
- push list missing a required entry → exit 1
- unmutated baseline → exit 0
- push paths rewritten flow-style with the same entries → exit 0

`node --test scripts/stacked-ci-workflows.test.mjs` still passes (4/4).

ebarti:
Addressed both review findings in 92a659cbb:

1. **Push filter completeness** — added the INV-1 outreach sources (`apps/web/src/contexts/outreach/**`, `apps/web/src/views/outreach/**`, `apps/api/src/outreach.ts`) plus root `package.json`, `README.md`, `LICENSE`, `NOTICE` to python.yml's push paths and to the contract script's required list; swept all repo-root joins in the test suite to confirm this is now the complete set. PR body's `apps/**` rationale corrected.
2. **Contract check hardened** — python.yml triggers are now parsed as real YAML (Ruby `YAML.safe_load`, same as stacked-ci-workflows.test.mjs) with a strict `on.pull_request === null` assertion, so narrowing fails regardless of YAML style; the inverted triggerPaths doc comment is gone with the helper.

Validation: contract check exit 0 at head; `node --test scripts/stacked-ci-workflows.test.mjs` 4/4; `python3 scripts/release_check.py` passed; Ruby YAML diff confirms only the 7 intended push-path additions (pull_request/workflow_dispatch/jobs untouched); 8-case mutation matrix (flow-style, unquoted, canonical paths, absent trigger, branches filter, dropped entry → all fail; baseline and flow-style-equivalent push list → pass); `git diff --check` clean.

PR #768, opened 2026-08-07
ci: scan every pull request with the release privacy gate

ebarti:
## What changed

`.github/workflows/release-check.yml`: removed `branches: ["main"]` from the `pull_request` trigger. Added `test_release_privacy_workflow_scans_every_pull_request` to `workers/automation/tests/test_release_check.py`.

## What the gate is, and why its trigger matters

This repository is public and is developed against real personal data. `scripts/release_check.py` is what stands between that and a published commit: forbidden path names and file suffixes (browser profile artifacts, private runtime files), PII and secret needles scanned in source *and inside minified build output and demo build artifacts* so a bundled chunk cannot smuggle them, non-placeholder secret assignments, "private planning corpus must not be tracked", demo PDF fixtures pinned to content digests so a real resume cannot be swapped for the demo one, the old-product-name gate, version parity, and release-distribution/PyPI/Homebrew safety. The second workflow step re-runs the scanner's own tests so the gate cannot be quietly weakened.

It is the most consequential check in the repository and it takes 12s. Its trigger should be the widest one here, not a narrow one.

## The observation

The Release Privacy Gate produced **no workflow run at all** on two recent pull requests:

| PR | Base | Stack | `Release check` |
| --- | --- | --- | --- |
| #747 | `chore/upgrade-codex-sdk` | — | **absent** |
| #741 | `main` | `stack != null`, not top (Python/TypeScript both `skipped`) | **absent** |
| #765–#768 | `main` | not stacked | present, 12–17s |

Every workflow in this repository *without* a `branches` filter (`python`, `typescript`, `launcher`, `demo-site`, `dco`) ran on #741. The two *with* one (`release-check`, `docs-site`) did not — though `docs-site` is separately explained by its `paths` filter, so `release-check` is the only clean data point.

## Honest limitation

[GitHub's documentation](https://docs.github.com/en/pull-requests/how-tos/merge-and-close-pull-requests/optimizing-ci-for-stacked-pull-requests) states that each PR in a stack "is treated as if it targets the base of the stack, such as `main`", so a `branches: ["main"]` filter should match *every* layer. That contradicts the #741 observation, and I cannot reconcile the two. Stacked pull requests are in public preview and flagged subject to change.

**The change is correct under either reading.** Removing `branches` makes the trigger a strict superset of what it matched before: a no-op if the documentation holds, and a real fix if the observed behavior is what actually happens. It also brings this workflow in line with the five that already carry no `branches` filter, and with the direction #644 took for `python.yml` and `typescript.yml`.

## On the absence of a `paths` filter

This PR deliberately does not add one, and the new test forbids it. Any file can carry PII or a secret, so every excluded path is a path a leak lands in unscanned. That reasoning is now a comment on the trigger plus an assertion, so it does not get "optimized" back in.

## Validation

- `ruff check workers/automation/src workers/automation/tests` with the pinned `ruff==0.15.8` from `workers/automation/pyproject.toml:59` → `All checks passed!`. (Unpinned 0.16.0 reports pre-existing findings in untouched files; none from this change.)
- `pytest -q --noconftest workers/automation/tests/test_release_check.py` → **53 passed**, the exact command in the workflow's self-test step.
- `python3 scripts/release_check.py --strict-prompt` → `Release check passed.`
- PyYAML parse confirming `pull_request` resolves to `None`, `push` still `{'branches': ['main']}`, all 5 steps unchanged.
- `git diff --check` clean.
- **Two mutation tests** against the committed baseline: restoring `branches: ["main"]` and adding a `paths` filter each fail the new test.

### DIFF
--- a/workers/automation/tests/test_release_check.py
+++ b/workers/automation/tests/test_release_check.py
@@ -1199,13 +1199,20 @@ def test_release_privacy_workflow_enforces_strict_prompt_gate() -> None:
 
 
 def test_release_privacy_workflow_scans_every_pull_request() -> None:
-    """The privacy gate must not be narrowed by base branch or by path.
+    """The privacy gate's ``pull_request`` trigger must stay exactly bare.
 
     A ``branches: ["main"]`` filter skipped the scan on every stacked pull
     request, because a stacked layer targets its parent branch rather than
     main. A ``paths`` filter would be self-defeating: any file in the
     repository can carry owner PII or a secret, so every excluded path is a
-    path a leak can land in unscanned.
+    path a leak can land in unscanned. Requiring the block to be exactly bare
+    also rejects the ``branches-ignore``/``paths-ignore`` spellings and
+    ``types`` narrowing such as ``types: [opened]``, which would leave every
+    later push to an open pull request unscanned. A bare ``pull_request:`` key
+    parses to YAML null, the same invariant ``scripts/stacked-ci-workflows.test.mjs``
+    asserts as ``on.pull_request === null``; string equality is used here
+    because this suite also runs in the gate workflow itself, whose
+    environment installs only pytest and has no YAML parser.
     """
     workflow = (
         release_check.ROOT / ".github/workflows/release-check.yml"
@@ -1215,8 +1222,11 @@ def test_release_privacy_workflow_scans_every_pull_request() -> None:
         on_block.index("  pull_request:") : on_block.index("  workflow_dispatch:")
     ]
 
-    assert "branches:" not in pull_request
-    assert "paths:" not in pull_request
+    assert pull_request == "  pull_request:\n", (
+        "the pull_request trigger must stay exactly bare (YAML null): any "
+        "branches/paths/types filter or -ignore variant narrows which pull "
+        "requests or files the privacy gate scans"
+    )
 
 
 def test_old_product_name_gate_blocks_shipping_surfaces(tmp_path: Path) -> None:
--- a/workers/automation/tests/test_release_check.py
+++ b/workers/automation/tests/test_release_check.py
@@ -1198,6 +1198,27 @@ def test_release_privacy_workflow_enforces_strict_prompt_gate() -> None:
     assert "python3 scripts/release_check.py --strict-prompt" in workflow
 
 
+def test_release_privacy_workflow_scans_every_pull_request() -> None:
+    """The privacy gate must not be narrowed by base branch or by path.
+
+    A ``branches: ["main"]`` filter skipped the scan on every stacked pull
+    request, because a stacked layer targets its parent branch rather than
+    main. A ``paths`` filter would be self-defeating: any file in the
+    repository can carry owner PII or a secret, so every excluded path is a
+    path a leak can land in unscanned.
+    """
+    workflow = (
+        release_check.ROOT / ".github/workflows/release-check.yml"
+    ).read_text(encoding="utf-8")
+    on_block = workflow[workflow.index("\non:") : workflow.index("\njobs:")]
+    pull_request = on_block[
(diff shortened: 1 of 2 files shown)


workers/automation/tests/test_release_check.py
```diff
@@ -1198,6 +1198,27 @@ def test_release_privacy_workflow_enforces_strict_prompt_gate() -> None:
...
+    workflow = (
+        release_check.ROOT / ".github/workflows/release-check.yml"
+    ).read_text(encoding="utf-8")
+    on_block = workflow[workflow.index("\non:") : workflow.index("\njobs:")]
+    pull_request = on_block[
+        on_block.index("  pull_request:") : on_block.index("  workflow_dispatch:")
+    ]
+
+    assert "branches:" not in pull_request
+    assert "paths:" not in pull_request
```
ebarti: PR Reviewer - pr-768-reviewer

Severity: Medium
Gate: Pass
Evidence: These two asserts check only the exact spellings `branches:` and `paths:`. Mutation check against f6d8715: inserting `paths-ignore: ["docs/**", "**.md"]`, `branches-ignore: [...]`, or `types: [opened]` under `pull_request:` each passes this test while narrowing the trigger — `paths-ignore` excludes paths from scanning, `branches-ignore` excludes base branches, and `types: [opened]` leaves every subsequent push to an open PR unscanned.
Reproduction: Reproduced with a script that applies the three mutations to the committed workflow and re-evaluates this test's slice-and-assert logic: all three bypass it, while restoring `branches: ["main"]` (the control) is caught.

The docstring says the gate "must not be narrowed by base branch or by path", and the PR body positions this test as the reason the trigger "does not get 'optimized' back in" — but the assertions miss the most likely optimization spellings (`paths-ignore` with a docs/markdown exclusion is the classic variant). As written, the tamper-guard only catches the one historical regression it was derived from.

Required fix: Assert the `pull_request` block is exactly bare, e.g. `assert pull_request == "  pull_request:\n"` — this also catches `types:` narrowing and matches the `on.pull_request === null` precedent in `scripts/stacked-ci-workflows.test.mjs:52-56`. At minimum, drop the colons (`assert "branches" not in pull_request` / `assert "paths" not in pull_request`) to catch the `-ignore` variants.

ebarti (commented):
Reviewed the trigger change and the guard test. The workflow edit itself is verified correct: `pull_request` parses to `None` (all PRs, default activity types), `push` stays scoped to `main`, all 5 steps unchanged, and the 53-test self-suite passes. One Medium finding: the new tamper-guard test only catches the exact spellings `branches:`/`paths:` and is bypassed by `branches-ignore:`, `paths-ignore:`, and `types:` narrowing, so it does not fully deliver the "cannot be optimized back in" guarantee the PR body relies on. No Blocker/High findings.

workers/automation/tests/test_release_check.py
ebarti: Fixed in 6f74418db. The test now asserts the sliced block equals the bare `"  pull_request:\n"` line exactly, per your primary suggestion — the string form of the `on.pull_request === null` invariant in `scripts/stacked-ci-workflows.test.mjs`. Raw-text equality (rather than PyYAML) is deliberate and now documented in the docstring: the gate workflow's self-test step installs only pytest (`python -m pip install pytest`), so a `yaml` import would break — or an importorskip would skip — this guard exactly where it must run. Mutation-checked against the committed workflow: `branches-ignore: ["tmp/**"]`, `paths-ignore: ["docs/**", "**.md"]`, `types: [opened]`, `branches: ["main"]`, and `paths: ["src/**"]` each fail the test; the bare trigger passes (full suite 53 passed via `pytest -q --noconftest workers/automation/tests/test_release_check.py`).

PR #770, opened 2026-08-08
refactor(pipeline): delete the dead legacy in-process pipeline

ebarti:
## Summary
Deletes the dead pre-Temporal in-process pipeline: `run_discovery_legacy_once` (~437 lines), the `_STAGE_RUNNERS` map, `_build_stage_kwargs`, `_resolve_stages`, and the discovery enrichment worker-thread helpers. Zero production callers existed — the Temporal `DiscoverWorkflow` path replaced this orchestrator (2026-07-03 ADRs) and the subtree survived only through tests that exercised the dead entrypoint itself. Also stops re-exporting `_PENDING_SQL`/`_count_pending`/`_resolve_stages` as package API (tests import the runner module directly), preserves the `_StageRunner` alias with its one live consumer (`_run_stage_observed`), and deletes `_stage_status`'s discover-only escalation branch, which lost its last caller with the orchestrator.

## Why
Wave-1 simplification: the deletion test — removing this concentrates nothing anywhere else; it is pure dead weight that misleads readers and agents.

## Validation
`uv run --frozen --extra dev pytest` on all six touched/affected test files (score/materials queue selectors, pipeline observability, discovery preparation orchestration, tailor retailor, enrichment queue selectors) — green at the layer and at the stack tip; ruff clean; review follow-up commit migrated the two remaining package-private test imports the first pass missed.

PR #771, opened 2026-08-08
refactor(preparation): delete the stillborn work-item machinery

ebarti:
## Summary
Deletes the stillborn preparation work-item machinery: the `PreparationWorkItemState` enum, `PreparationWorkItem` dataclass, its repository port, and `SqlitePreparationWorkItemRepository` — none had a production caller and no code ever updated the table's state column (the real durable lifecycle is `DiscoveryExecutionWorkPlanState` + the #743 recovery modules). `work_items.py` shrinks to the live vocabulary (`PreparationWorkItemKind`, `make_preparation_idempotency_key` — the identity of `prep-{key}` workflow ids). The live `SqlitePreparationTargetReader` stays. Also deletes the never-raised `AuthenticationError`/`BrowserTransientError` classes and the consumed-but-never-produced "authentication" entry from the four workflow non-retryable lists (nothing can produce that type — verified by grep and by the taxonomy test).

## Why
The decoy state machine sat beside the real one; the error classes advertised behavior production never exercises.

## Deferred (tracked)
The `preparation_work_items` table stays in the sealed v7 DDL (removal = cross-runtime manifest resync, next schema revision). Two architecture-doc taxonomy tables still list the deleted error classes — stack-final docs pass.

## Validation
v6→v7 work-items copy suite, plan/population/candidate migration suites, p1b error inversion, workflow preparation, job repository, scorer, tailor-retailor — 126 tests green; ruff clean; migration validates states against a frozen local vocabulary matching the deleted enum exactly.


ebarti:
PR Reviewer - pr-reviewer

Severity: Low
Gate: Pass
Evidence: The error-taxonomy tables in docs/architecture/runtime.md:623-626 and docs/architecture/pipeline/envelope.md:324-328 still list `AuthenticationError` (`authentication`) and `BrowserTransientError` (`browser_transient`), both deleted here. Also cosmetic: three test-harness RetryPolicies still carry the now-dead "authentication" literal (tests/test_workflow_job_preparation.py:559 and :803, tests/test_p1b_error_inversion.py:67) — harmless, since they are test-local policy overrides, but the vocabulary no longer exists anywhere in src.
Reproduction: Reasoned from code (grep at the stack tip; `grep -rn '"authentication"' workers/automation/src` returns nothing, confirming the four-list removal is behavior-neutral).

Everything else verified clean: `SqlitePreparationTargetReader` intact with its 4 live consumers (interview, scorer, tailor, cover_letter); the migration's frozen state vocabulary matches the deleted enum exactly (queued/running/completed/failed) and `state.value` -> `state` is correct; no schema or manifest file touched; sealed schema_v7.sql still carries the `preparation_work_items` table and claim index; migration copy tests pass (tests/test_v6_to_v7_work_items_copy.py, 7 passed). The Python CI failure on this PR is fully inherited from the PR #770 blocker (same 6 AttributeError failures), not from this layer.

Required fix: Remove the two deleted classes from both taxonomy tables — either here or, if canonical docs are deferred to the stack-final PR per repo policy, track it explicitly so the docs pass does not miss it. Optionally drop the dead "authentication" literal from the three test-harness policies.

PR #772, opened 2026-08-08
fix(projections): make the TS watermark monotonic and bound read-path folds

ebarti:
## Summary
Two fixes to the shared-watermark discipline in `apps/api/src/projections.ts`: (1) `setWatermark` becomes monotonic — `MAX(last_event_id, excluded)` with `updated_at` moving only on a real advance — mirroring the documented Python setter (`infrastructure/events/watermark.py`) that the TS twin had skipped; a slow TS refresh could previously rewind an advance made by the worker's builder. (2) The per-request catch-up scan gains `REFRESH_EVENT_BATCH_LIMIT` (1000, matching the SSE tick's bound): the first read after a large discovery run no longer folds the entire backlog synchronously; the watermark only advances to the last event actually scanned, so capped passes resume correctly.

## Why
A live one-sided mirror drift on the projection substrate, and an unbounded hot-path fold — both found in the logic/process architecture review.

## Validation
Two new tests pin the behaviors (stale write is a no-op preserving updated_at; a limit+2 backlog folds across exactly two refreshes with exact stop points); full projections + server suites green (274); api typecheck green.

### DIFF
--- a/apps/api/src/projections.ts
+++ b/apps/api/src/projections.ts
@@ -28,7 +28,7 @@ import type {
   MarketCompensationEstimateResponse,
   PostedCompensationFactResponse,
 } from "./contracts.js";
-import { allRows, getRow, tableExists, type SqliteDatabase, type SqliteValue } from "./db.js";
+import { allRows, getRow, openDatabase, tableExists, type SqliteDatabase, type SqliteValue } from "./db.js";
 import { normalizeJobLocation } from "./location-normalization.js";
 import { getMarketCompensationEstimate } from "./market-compensation-estimates.js";
 import { getPostedCompensationFact } from "./posted-compensation-facts.js";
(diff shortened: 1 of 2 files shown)


ebarti:
PR Reviewer - pr-reviewer

Severity: Low
Gate: Pass
Evidence: The PR title is "fix/monotonic watermark bounded fold" (the branch name), which does not follow Conventional Commits — an explicit repo rule ("PR titles must follow Conventional Commits"), and the squash-merge commit on main would inherit it. The underlying commit message is fine: "fix(projections): make the TS watermark monotonic and bound read-path folds".
Reproduction: Not run: metadata-only finding.

Code verified clean: the setter SQL mirrors workers/automation/src/jobctrl/infrastructure/events/watermark.py exactly (MAX + updated_at CASE); the LIMIT'd fold only advances `maxEventId` over scanned rows so the watermark can never pass unscanned events; the backfill detectors are count/existence-based and unaffected by the cap (uncounted events stay above the watermark for the next pass); `setWatermark` has a single production caller (projections.ts:804); REFRESH_EVENT_BATCH_LIMIT (1000) matches event-stream.ts EVENT_BATCH_LIMIT (1000) as the comment claims. Both new tests pin the exact behaviors and pass locally, plus the full server+projections suites (274 passed).

Required fix: Retitle the PR to the commit message (or equivalent Conventional Commits form) before merge.

apps/api/src/projections.ts
```diff
@@ -640,8 +658,8 @@ export function refreshProjections(db: SqliteDatabase, tenantId = "local"): void
   {
     const rows = allRows<{ event_id: number; tenant_id: string; job_id: string | null; event_type: string }>(
       db,
-      "SELECT event_id, tenant_id, job_id, event_type FROM job_events WHERE event_id > ? ORDER BY event_id ASC",
-      [watermark],
+      "SELECT event_id, tenant_id, job_id, event_type FROM job_events WHERE event_id > ? ORDER BY event_id ASC LIMIT ?",
```
ebarti, line 704: **High — guarantee the capped backlog is drained.** With 1,001 pending events, an existing job whose canonical update is represented by event 1,001 remains stale after this read: the watermark stops at 1,000 and the request returns. A new SSE subscriber initializes its cursor to the current maximum event ID, so event 1,001 is not replayed to force another refetch; without another user read, the stale projection can persist indefinitely. I reproduced `afterOne="Old title", wmOne=1000`, then only a manual second refresh produced `afterTwo="New title", wmTwo=1001`. Please retain the per-pass budget while arranging an automatic/background drain (or another guaranteed follow-up refresh), and add a regression proving the remainder catches up without a manual second read.

ebarti (commented):
Found one High-severity freshness regression in the bounded catch-up path; details inline.

apps/api/src/projections.ts
ebarti: Fixed in the follow-up commit on this branch: a capped pass now schedules a background drain (one REFRESH_EVENT_BATCH_LIMIT batch per event-loop tick, own connection, in-flight guard per db+tenant, stops on error and hands back to the next foreground read). The per-request bound is unchanged. Regression added with your exact scenario: 1000 fillers + canonical rename as event 1001 → single refresh stops at the cap with the stale title asserted synchronously, then vi.waitFor proves watermark and title converge with no second read.

PR #773, opened 2026-08-08
refactor(apply): single-source the approval-gate vocabulary

ebarti:
## Summary
Single-sources the apply-review approval-gate vocabulary. `packages/contracts` now exports `APPLY_REVIEW_APPROVAL_GATE_REASONS` as the runtime constant with the `ApplyReviewApprovalGateReason` type derived from it; the server's precondition set builds its gate subset from the typed vocabulary; the demo executor's thrown literals are compile-fenced with `satisfies`; Python gains `APPROVAL_GATE_REFUSAL_REASONS` in `domain/apply/value_objects.py` (exported via `__all__`), pinned to the shared fixture `apply_approval_gate_reasons.json` together with a source-scan test over `_approval_refusal_reason`'s return literals.

## Why
The gate reasons were spelled independently in five places across two runtimes with no cross-check. The new pin proved its worth immediately: the launcher's refusal vocabulary is the full seven reasons (including `awaiting_dry_run` and `override_evidence_invalid`), not the five the earlier review pass had catalogued — the fixture now records the truth.

## Validation
TS parity test (constant == fixture) and Python tests (constant == fixture; every launcher return literal inside the constant) green; server precondition set unchanged as a set; contracts/api/web typechecks green.

### DIFF
--- a/packages/contracts/src/schemas.ts
+++ b/packages/contracts/src/schemas.ts
@@ -872,14 +872,22 @@ export interface RepeatApplicationOverrideResponse {
 }
 
 export type ApplyReviewDryRunCoverage = "full" | "partial";
-export type ApplyReviewApprovalGateReason =
-  | "awaiting_approval"
-  | "awaiting_dry_run"
-  | "approval_stale_materials"
-  | "approval_stale_profile"
-  | "approval_stale_url"
-  | "approval_stale_email_candidate"
-  | "override_evidence_invalid";
+/**
+ * The apply-review approval-gate vocabulary. This constant is the single
+ * TypeScript source for the gate reasons; the Python launcher's refusal
+ * reasons are pinned to the same fixture
+ * (packages/domain-types/test/fixtures/apply_approval_gate_reasons.json).
+ */
+export const APPLY_REVIEW_APPROVAL_GATE_REASONS = [
+  "awaiting_approval",
+  "awaiting_dry_run",
+  "approval_stale_materials",
+  "approval_stale_profile",
+  "approval_stale_url",
+  "approval_stale_email_candidate",
+  "override_evidence_invalid",
+] as const;
+export type ApplyReviewApprovalGateReason = (typeof APPLY_REVIEW_APPROVAL_GATE_REASONS)[number];
 
 export interface ApplyReviewDryRunEvidence {
   runId: string;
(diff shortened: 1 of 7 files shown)


workers/automation/tests/test_apply_approval_vocabulary.py
```diff
@@ -0,0 +1,40 @@
...
+    fixture = _fixture()
+    assert APPROVAL_GATE_REFUSAL_REASONS == frozenset(fixture["launcherRefusalReasons"])
+    assert APPROVAL_GATE_REFUSAL_REASONS <= frozenset(fixture["reasons"])
+
+
+def test_launcher_refusal_literals_stay_inside_the_vocabulary() -> None:
+    source = inspect.getsource(launcher._approval_refusal_reason)
+    literals = set(re.findall(r'return "([a-z_]+)"', source))
+    assert literals, "expected _approval_refusal_reason to return literal reasons"
+    unknown = literals - APPROVAL_GATE_REFUSAL_REASONS
```
ebarti, line 39: **Medium — make the launcher pin bidirectional.** `launcherRefusalReasons` is presented as the launcher’s exact vocabulary, but this only proves `literals ⊆ APPROVAL_GATE_REFUSAL_REASONS`. If a refusal branch is removed or starts returning `None`, the fixture and exported constant remain stale while both new tests pass. Assert equality (with missing/unknown diagnostics) so the pin catches removals as well as additions.

packages/contracts/src/schemas.ts
```diff
@@ -872,14 +872,22 @@ export interface RepeatApplicationOverrideResponse {
...
+export const APPLY_REVIEW_APPROVAL_GATE_REASONS = [
+  "awaiting_approval",
+  "awaiting_dry_run",
+  "approval_stale_materials",
+  "approval_stale_profile",
+  "approval_stale_url",
+  "approval_stale_email_candidate",
+  "override_evidence_invalid",
+] as const;
+export type ApplyReviewApprovalGateReason = (typeof APPLY_REVIEW_APPROVAL_GATE_REASONS)[number];
```
ebarti, line 890: **Medium — fence the user-facing reason map too.** `apps/web/src/contexts/apply/components/ApplyReviewDecisionControls.tsx` still declares `GATE_REASON_LABELS` as `Record<string, string>`. Adding a value to this canonical list therefore typechecks without a label and the UI falls back to raw snake_case, leaving another independent copy of the vocabulary. Type that table with `ApplyReviewApprovalGateReason` (or an exhaustive `satisfies` map) so this source actually protects the consumer.

ebarti (commented):
Two drift-prevention gaps remain in the new approval-gate vocabulary pin.

PR #775, opened 2026-08-08
refactor(contracts): share the jobs-query semantics between API and demo

ebarti:
## Summary
Extracts the jobs-list query semantics — `filterJob`, `compareJobs` (with a `normalizeSqlText` option), the compensation sort values, `STATE_RANK` (typed `Readonly<Record<StageState, number>>`), `compareValues`, timestamp helpers, and the `paginate` envelope — into `packages/contracts/src/jobs-query.ts`, consumed by both the API read model and the demo adapter. The demo keeps only a thin `filterJob` wrapper for the `deleted` facet it must emulate client-side (the read model resolves it in SQL) and passes `normalizeSqlText` per sort field exactly as before.

## Why
The two hand-copied TypeScript twins were kept aligned only by hand-re-encoded test expectations; a filter or sort fix in one could silently miss the other (the demo drifted this way historically). One module, two adapters.

## Validation
Reviewer confirmed byte-level parity of the shared functions against both deleted twins under both collation modes. Demo adapter suite (14, including the matches-production filter/sort tests — now true by construction), read-model-v7 (5), server+projections (274); contracts/api/web typechecks green. Demo adapter shrank 309 lines.

### DIFF
--- a/packages/contracts/src/jobs-query.ts
+++ b/packages/contracts/src/jobs-query.ts
@@ -1,0 +1,267 @@
+/**
+ * Shared jobs-list query semantics: filtering, sorting, and pagination over
+ * `JobSummary` rows.
+ *
+ * Two adapters consume this module: the local API's read model (SQLite-backed
+ * rows) and the demo workspace's in-memory adapter. Both previously carried
+ * hand-copied twins of these functions, with agreement asserted only by
+ * hand-re-encoded test expectations; a filter or sort fix now lands once.
+ *
+ * The demo emulates two SQL-side effects the API gets from SQLite itself:
+ * the `deleted` facet (applied in SQL WHERE by the read model) stays in the
+ * demo's wrapper, and `normalizeSqlText` reproduces the read model's
+ * case-insensitive ORDER BY collation for text sort fields.
+ */
+
+import type { JobCompensationSummary, JobListQuery, JobSummary, PaginatedResponse } from "./schemas.js";
+
+/** Terminal-first ranking used by the current_state sort arm. */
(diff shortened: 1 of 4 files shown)


packages/contracts/src/jobs-query.ts
```diff
@@ -0,0 +1,267 @@
...
+ * The demo emulates two SQL-side effects the API gets from SQLite itself:
+ * the `deleted` facet (applied in SQL WHERE by the read model) stays in the
+ * demo's wrapper, and `normalizeSqlText` reproduces the read model's
+ * case-insensitive ORDER BY collation for text sort fields.
+ */
+
+import type { JobCompensationSummary, JobListQuery, JobSummary, PaginatedResponse } from "./schemas.js";
+
+/** Terminal-first ranking used by the current_state sort arm. */
+export const STATE_RANK: Readonly<Record<string, number>> = {
```
ebarti: PR Reviewer - pr-reviewer

Severity: Low
Gate: Pass
Evidence: The API's deleted twin typed this table `Record<StageState, number>` (apps/api/src/read-model.ts pre-diff), so adding a new `StageState` without a rank was a compile error. The shared version widens to `Readonly<Record<string, number>>`, so a future state silently falls to the 999 fallback both here (compareJobs current_state arm) and in the read model's SQL rank CASE built from these entries (apps/api/src/read-model.ts:355).
Reproduction: Reasoned from code; behavior today is identical (all 11 entries and values match both deleted twins).

Required fix: Type the constant `Readonly<Record<StageState, number>>` (StageState is exported from ./schemas.js) to restore compile-time exhaustiveness for both consumers.

PR #777, opened 2026-08-08
fix(api): render resume PDFs through the JSON-RPC worker seam

ebarti:
## Summary
Resume PDF rendering moves from a synchronous `execFileSync` Python+Chromium subprocess (up to 120s inside the API's event loop) to a `render_resume_pdf` sync JSON-RPC method on the API's long-lived `jobctrl rpc` child. `ResumeHtmlPdfRenderer` becomes async; the templates render-only refresh and the drafts flow await it with database rows persisted only after artifact files exist. `renderResumeReviewDraft` renders **before** its transaction, re-validates the draft and the materials generation inside a short write transaction, and removes orphaned artifact files when the commit fails.

## Why
Every render froze the whole API — every route, `/v1/health`, the SSE tick — for its duration, and the draft path additionally wrapped the subprocess in a deferred-BEGIN transaction whose write upgrade failed with `SQLITE_BUSY_SNAPSHOT` whenever the worker heartbeat committed mid-render (500 + stranded files). Architecture review R1 finding.

## Validation
api/web/contracts typechecks; 233-test server suite; resume-templates, resume-review-drafts, qa-workflow, projections suites; new Python handler tests (param validation, adapter dispatch, sync registration). Tier-2 product QA (real end-to-end render through the rpc child + health-during-render probe + orphan check) runs as a separate gate on this stack.

### DIFF
--- a/apps/api/src/resume-pdf-render.ts
+++ b/apps/api/src/resume-pdf-render.ts
@@ -23,7 +23,16 @@ export class ResumeRenderError extends Error {
 
 export function createResumeHtmlPdfRenderer(dispatcher: JsonRpcDispatcher): ResumeHtmlPdfRenderer {
   return async ({ htmlPath, pdfPath }) => {
-    const response = await dispatcher.call(RpcMethods.RenderResumePdf, { htmlPath, pdfPath });
+    let response: Awaited<ReturnType<JsonRpcDispatcher["call"]>>;
+    try {
+      response = await dispatcher.call(RpcMethods.RenderResumePdf, { htmlPath, pdfPath });
+    } catch (error) {
+      // Transport-level failures (spawn/write errors, the request timeout,
+      // child exit) reject instead of returning a JSON-RPC error envelope.
+      throw new ResumeRenderError(
+        `Resume HTML-to-PDF render failed: ${error instanceof Error ? error.message : String(error)}`,
+      );
+    }
     if (response.error) {
       // The RPC server wraps handler exceptions as a generic "Internal
       // error" message and puts the real cause in error.data.
--- a/apps/api/src/resume-pdf-render.ts
+++ b/apps/api/src/resume-pdf-render.ts
@@ -1,73 +1,31 @@
-import { execFileSync } from "node:child_process";
-import { defaultSourcePythonRuntime, type PythonRuntimeCommandResolver } from "./python-runtime.js";
-
-// The render is synchronous so it can run inside the better-sqlite3 write
-// transaction. Bound the subprocess so a hung runtime resolve or Chromium launch
-// cannot freeze the Node event loop indefinitely; on timeout execFileSync throws
-// and the caller's catch turns it into a preserved-generation render failure.
-// Generous enough for a cold Chromium start.
-const RESUME_RENDER_TIMEOUT_MS = 120_000;
-const RESUME_RENDER_MAX_BUFFER_BYTES = 64 * 1024 * 1024;
+import type { JsonRpcDispatcher } from "./json-rpc-adapter.js";
+import { getDefaultJsonRpcDispatcher } from "./json-rpc-adapter.js";
+import { RenderResumePdfResultSchema, RpcMethods } from "./contracts.js";
 
 export interface ResumeHtmlPdfRenderInput {
   readonly htmlPath: string;
   readonly pdfPath: string;
 }
 
-export type ResumeHtmlPdfRenderer = (input: ResumeHtmlPdfRenderInput) => void;
+export type ResumeHtmlPdfRenderer = (input: ResumeHtmlPdfRenderInput) => Promise<void>;
 
-export const RESUME_HTML_PDF_SCRIPT = `
-import sys
-from pathlib import Path
-
-from jobctrl.infrastructure.materials.html_resume_pdf import render_resume_html_to_pdf
-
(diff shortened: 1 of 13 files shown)


apps/api/src/resume-review-drafts.ts
```diff
@@ -495,65 +488,101 @@ export function seedResumeReviewCommentThreads(
...
+        layoutBoxCount: rendered.layoutBoxes.length,
       };
+    });
+    return tx();
+  } catch (error) {
+    // The files rendered but the rows did not commit: remove the orphans so
+    // a failed render leaves no artifacts without database rows.
+    for (const orphan of [rendered.textPath, rendered.htmlPath, rendered.pdfPath]) {
+      try {
+        fs.rmSync(orphan, { force: true });
```
ebarti, line 589: **High — make conflict cleanup ownership-safe.** Two concurrent render requests for the same draft/revision compute the same generation and deterministic artifact suffix before awaiting, so they use identical `.txt`/`.html`/`.pdf` paths. One request can commit those paths; the loser then hits the state conflict and this cleanup deletes the winner’s registered files, leaving a successful generation with missing artifacts. Use per-attempt temporary paths plus atomic promotion, or single-flight the draft, and add a concurrent-render regression.

apps/api/src/resume-templates.ts
```diff
@@ -935,7 +935,18 @@ function persistRenderOnlyRefresh(
 
   fs.writeFileSync(textPath, text, "utf8");
   fs.writeFileSync(htmlPath, htmlForTemplateRefresh(text, effective), "utf8");
-  renderPdf({ htmlPath, pdfPath });
+  try {
+    await renderPdf({ htmlPath, pdfPath });
```
ebarti: **High — revalidate and commit the refresh after the async gap.** Generation, source material, and effective template are captured before this await; another material write, refresh, or template assignment can commit while rendering. The code can then publish stale-template output or collide during the separate autocommit inserts, while cleanup only covers renderer failures, leaving orphan files or a partial generation. Recheck the source/template/generation in a short transaction and clean up attempt-owned files on any post-render failure; cover overlapping refreshes/material writes.

apps/api/src/resume-pdf-render.ts
```diff
@@ -1,73 +1,44 @@
...
-        env: command.env,
-        stdio: ["ignore", "pipe", "pipe"],
-        timeout: RESUME_RENDER_TIMEOUT_MS,
-        maxBuffer: RESUME_RENDER_MAX_BUFFER_BYTES,
-      });
-    } catch (error) {
-      throw new Error(resumeRenderErrorMessage(error));
+export function createResumeHtmlPdfRenderer(dispatcher: JsonRpcDispatcher): ResumeHtmlPdfRenderer {
+  return async ({ htmlPath, pdfPath }) => {
+    const response = await dispatcher.call(RpcMethods.RenderResumePdf, { htmlPath, pdfPath });
```
ebarti: **Medium — normalize dispatcher rejections too.** `dispatcher.call()` rejects (rather than returning `response.error`) when the child cannot spawn/write, closes, or reaches its request timeout. Those render failures bypass `ResumeRenderError`, so `withWritableDb` reports `500 db_write_failed` even though the database was not the failing boundary. Catch transport rejections here, rethrow a sanitized `ResumeRenderError`, and add a rejecting-dispatcher test.

ebarti (commented):
Three actionable issues: two high-severity concurrency/data-integrity blockers in the async render paths, plus one medium-severity transport-error classification bug.

apps/api/src/resume-review-drafts.ts
ebarti: Fixed: each attempt now renders to `.<attemptId>.tmp` files and promotes them to the deterministic finals via renameSync only inside the commit transaction, after the staleness guards — a loser never renames and its cleanup removes only its own attempt-suffixed temps (finals are added to cleanup only when the failure happened after promotion, where the guards make them attempt-owned). Regression added: two concurrent renders of the same revision — winner commits and its files survive, loser gets 409 resume_render_conflict, zero .tmp strays.

apps/api/src/resume-templates.ts
ebarti: Fixed: persistRenderOnlyRefresh now renders to attempt-suffixed temps, then a single short db.transaction revalidates nextMaterialGeneration, the source material generation, and the effective templateVersionId+templateHash before promoting the temps and persisting job_materials + both artifacts + layout boxes together; any post-render failure cleans the attempt's temps (plus finals when promoted) and surfaces as the recorded failed attempt. Regression: a material write injected mid-render yields status=failed with a retry message, zero .tmp strays, and a clean forced retry completes.

apps/api/src/resume-pdf-render.ts
ebarti: Fixed: dispatcher.call is wrapped — transport rejections (spawn/write failure, request timeout, child exit) rethrow as ResumeRenderError with the sanitized message, so withWritableDb maps them to 502 resume_render_failed instead of 500 db_write_failed. Tests added for rejecting, error-envelope (error.data preferred), and invalid-result dispatchers; the stale execFileSync-era tests in resume-pdf-render.test.ts (which this seam had silently broken) are rewritten for the dispatcher design.

PR #778, opened 2026-08-08
fix(api): run the Gmail outcome scan through the JSON-RPC worker seam

ebarti:
## Summary
The Gmail outcome scan moves from a bespoke one-shot Python spawn (stdout line-scraping, no timeout, no overlap protection) to a `gmail_feedback_scan` sync JSON-RPC method on the long-lived rpc child. Auth failures still travel as `ok:false + message` inside the result, so the route's 400/503/500 status mapping is byte-identical; unset scan parameters keep their Python-side defaults; the route's injected-scanner test seam is untouched.

## Why
This was the last bespoke transport to Python. With the render seam (#777), the API now invokes Python exactly one way: the JSON-RPC adapter.

## Validation
New Python handler tests (camelCase mapping, unset-param defaults, ok:false passthrough, sync registration) — 92 handler-suite tests green; api/contracts typechecks; 233-test server suite green (the scan route's stub seam is unchanged).

### DIFF
--- a/apps/api/src/gmail-feedback-worker.ts
+++ b/apps/api/src/gmail-feedback-worker.ts
@@ -32,7 +32,9 @@ export class GmailFeedbackScanError extends Error {
 // the status mapping below stays behavior-identical.
 export function createWorkerGmailFeedbackScanner(dispatcher: JsonRpcDispatcher): GmailFeedbackScanner {
   return async (input, context) => {
-    const response = await dispatcher.call(RpcMethods.GmailFeedbackScan, {
+    let response: Awaited<ReturnType<JsonRpcDispatcher["call"]>>;
+    try {
+      response = await dispatcher.call(RpcMethods.GmailFeedbackScan, {
       expectedAppDir: context.appDir,
       expectedDbPath: context.dbPath,
       ...(input.recipientEmail !== undefined ? { recipientEmail: input.recipientEmail } : {}),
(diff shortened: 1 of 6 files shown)


apps/api/src/gmail-feedback-worker.ts
```diff
@@ -34,22 +26,40 @@ export class GmailFeedbackScanError extends Error {
...
-      appDir: context.appDir,
-      dbPath: context.dbPath,
-      pythonRuntime,
+// Runs the bounded Gmail outcome scan through the long-lived JSON-RPC worker
+// child instead of a bespoke one-shot subprocess with stdout line-parsing.
+// Auth failures still travel as { ok: false, message } inside the result so
+// the status mapping below stays behavior-identical.
+export function createWorkerGmailFeedbackScanner(dispatcher: JsonRpcDispatcher): GmailFeedbackScanner {
+  return async (input, context) => {
+    const response = await dispatcher.call(RpcMethods.GmailFeedbackScan, {
```
ebarti: High: wrap `dispatcher.call()` rejections in `GmailFeedbackScanError` and pin this with a rejecting-dispatcher route test. `SubprocessJsonRpcAdapter.call()` rejects on timeout, spawn/write failure, or child exit; those cases do not return a JSON-RPC `error` envelope. This `await` therefore bypasses the route's typed error handling and Fastify emits its generic 500 shape, whereas the deleted subprocess close/non-zero path returned the stable `gmail_feedback_scan_failed` response.

ebarti (commented):
One High transport-error regression needs fixing before this seam is behavior-preserving.

apps/api/src/gmail-feedback-worker.ts
ebarti: Fixed: dispatcher.call is wrapped in try/catch — transport rejections (timeout, spawn/write failure, child exit) rethrow as GmailFeedbackScanError through the same message-based status mapping, so the route keeps returning the stable gmail_feedback_scan_failed shape. Route test added: a real createWorkerGmailFeedbackScanner over a rejecting dispatcher → 500 {ok:false, error:"gmail_feedback_scan_failed", message:"Gmail feedback scan failed: JSON-RPC request timed out…"}.

PR #779, opened 2026-08-08
fix(web): improve job detail action layout

ebarti:
## Summary

- move the focused job-detail actions into the available header space
- group secondary preparation and application commands in an accessible popover
- keep artifact Open controls readable and contained in the diagnostics column
- clarify that compensation refresh and its optional observation path apply only to the current job
- add responsive regression coverage at 1440px, 1080px, and 390px

## Root cause

The artifact row inherited a generic mini-row grid whose final columns could collapse to zero width in the narrow inspector. The job action disclosure was also hidden at wider viewports, forcing every workflow command into one crowded inline strip while the header's first row remained underused. The compensation control called the current-job endpoint correctly, but its copy did not communicate that scope.

## Impact

Job details now keep primary navigation and review actions in the header, expose secondary commands through clearly labeled Preparation and Application groups, and preserve readable artifact controls without horizontal overflow. The compensation action explicitly says “Refresh this job”; the separate Jobs-page action remains the all-jobs refresh path.

## Validation

- `corepack pnpm web:check`
- `corepack pnpm web:test` — 2,007 tests passed
- `corepack pnpm web:build`
- `corepack pnpm web:storybook:build`
- `corepack pnpm web:storybook:test` — 453 tests passed
- targeted Chromium responsive E2E — 1440×900, 1080×900, and 390×844
- independent reviewer gate — PASS
- independent QA gate — PASS
- `git diff --check`

### DIFF
--- a/apps/web/src/styles/redesign-route-workspaces.css
+++ b/apps/web/src/styles/redesign-route-workspaces.css
@@ -90,7 +90,7 @@
 }
 
 .job-detail-workspace__header {
-  grid-template-columns: minmax(0, 1fr);
+  grid-template-columns: minmax(0, 1fr) auto;
 }
 
 .job-detail-workspace__header > .workspace-back,
@@ -100,6 +100,21 @@
 .workflow-run-workspace__header > .workspace-back,
 .job-run-workspace__header > .workspace-back {
   grid-column: 1 / -1;
+}
+
+.job-detail-workspace__header > .workspace-back {
+  grid-column: 1;
+  grid-row: 1;
+}
+
+.job-detail-workspace__header > .job-overview {
+  grid-column: 1 / -1;
+  grid-row: 2;
+}
+
+.job-detail-workspace__header > .job-detail-top-actions {
+  grid-column: 2;
+  grid-row: 1;
 }
 
 .job-overview {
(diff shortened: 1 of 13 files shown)


apps/web/src/styles/redesign-route-workspaces.css
```diff
@@ -102,6 +102,21 @@
...
+}
+
+.job-detail-workspace__header > .job-overview {
+  grid-column: 1 / -1;
+  grid-row: 2;
+}
+
+.job-detail-workspace__header > .job-detail-top-actions {
+  grid-column: 2;
+  grid-row: 1;
```
ebarti, line 117: The new grid rows paint these actions before `.job-overview`, but the JSX still renders `JobOverview` first and it contains the focusable “Open original posting” link. At desktop widths the Tab sequence is therefore Back → posting link (visual row 2) → Apply Review / Evidence map / More actions (visual row 1), making keyboard focus jump backward through the header. Please preserve DOM and visual order at this breakpoint and add an E2E assertion for the sequential focus order.

ebarti (commented):
Found one accessibility regression in the desktop header focus order.

apps/web/src/styles/redesign-route-workspaces.css
ebarti: Fixed: the top-actions block now renders before JobOverview in the DOM. Desktop visuals are unchanged (the grid placement is explicit), and the mobile flow order follows the DOM, so Back → actions → overview holds in DOM, visual, and Tab order at both breakpoints — the mobile actions strip keeps its separator by flipping to a bottom border in its new position. E2E added in jobs-drawer.spec.ts: desktop viewport, Tab from Back asserts Apply Review → Evidence map → More job actions → Open original posting in sequence. Also ran the PR's other touched suites (decision-workflows-responsive, mobile-shell-width, materials): 10/10 pass.

PR #781, opened 2026-08-09
fix(gmail): preserve canonical feedback identity and transactional ingestion

### LINKED ISSUE #780: bug(gmail): outcome scan uses retired job identity joins
Found by QA during the wave-1 stack gate (pre-existing on main, verified untouched by the stack).

**Symptom:** any Gmail outcome scan on a canonical v7 database fails before Gmail auth is ever consulted: `_outcome_anchors` in `workers/automation/src/jobctrl/infrastructure/gmail/feedback.py` (~line 642) selects `o.job_key` from `application_outcomes`, but the sealed v7 schema (`infrastructure/migrations/schema_v7.sql`) gives that table `job_id` (FK to jobs). SQLite raises `OperationalError: no such column: o.job_key`.

**Impact:** the scan's auth-guidance path (503) is unreachable on v7 whenever the outcomes table exists — every scan fails with the column error instead. This predates and is independent of PR #778's transport change; the JobId v7 migration train (#568–#668) evidently missed this reader.

**Fix shape:** migrate `_outcome_anchors` (and any sibling queries in feedback.py) to `job_id` + the canonical join, mirroring how the other v7 readers were cut over (e.g. #625 keyed feedback by canonical job id).

**Repro:** exact v7 DB with one applied-anchor job and a row in `application_outcomes`; `POST /v1/outcomes/gmail/scan` or direct `scan_gmail_feedback(db_path=...)`.

YiFanWangSCU:
Gmail outcome scans now carry canonical `jobId` values through the scanner, response contracts, stored events, and web invalidation. Application URLs remain search hints sourced from tenant-scoped `job_enrichments` on the exact-v10 schema.

Each linked message writes evidence, its suggestion, and the event atomically. A short transaction after Gmail I/O rechecks job existence and duplicate messages; notifications publish only after commit. This prevents stale private records after concurrent deletion, duplicate writes, and projection lock contention. Application/ATS search hints retain priority under the six-hint cap.

The contributor history and regressions are preserved. The branch is integrated with current main, including v10 runtime admission and canonical URL authority. Refs #780 (the original anchor-query failure was already fixed by #927).

Validation on the integrated candidate:
- 85 focused Python Gmail, event, and projection tests passed.
- 339 API tests, 183 domain-type tests, and 106 web invalidation tests passed.
- Domain types, contracts, API client, API, and web type checks passed.
- Worker Ruff, web boundary lint, web build, docs build/link checks, and diff check passed.
- Synthetic fixtures cover tenant isolation, concurrent deletion/duplicates, post-commit projection updates, schema rejection, and private-body exclusion. No real mailbox or personal database was used.
- Independent review and QA passed with no findings. QA exercised 11 scenarios through 17 production JSON-RPC subprocess calls, including injected suggestion/event/commit failures and successful retries.


workers/automation/src/jobctrl/infrastructure/gmail/feedback.py
```diff
@@ -642,16 +660,25 @@ def _job_anchors(conn: sqlite3.Connection) -> list[ApplicationAnchor]:
...
+    # older (v2.0.8/v6) databases still use the application job URL. job_key is
+    # taken from jobs.url, so an outcome whose job no longer exists is dropped
+    # instead of producing an empty-query anchor.
+    outcome_job_column = "job_id" if "job_id" in outcome_columns else "job_key"
     columns = _columns(conn, "jobs")
     title_expr = _column_expr(columns, "title", prefix="j")
     company_expr = _first_column_expr(columns, ["company", "site", "employer"], prefix="j")
     application_url_expr = _column_expr(columns, "application_url", prefix="j")
+    job_join_expr = (
+        "j.job_id = o.job_id" if outcome_job_column == "job_id" else "j.url = o.job_key"
```
ebarti: Scope these v7 anchors by the full tenant identity. `job_id` is unique only with `tenant_id`, but this join uses only `job_id`, while `_apply_run_anchors` also omits an `a.tenant_id` filter and tenant join. With the same UUID in `local` and another tenant, a local outcome produced anchors for both tenants, and an other-tenant apply run did the same. Please join on both `tenant_id` and `job_id`, filter every anchor source to `TENANT_ID`, and add a shared-JobId cross-tenant regression.

workers/automation/src/jobctrl/infrastructure/gmail/feedback.py
```diff
@@ -435,11 +435,18 @@ def _store_linked_message(
     to_addresses = _email_addresses(_text(full_message.get("to") or metadata.get("to")))
     thread_id = _nullable_text(full_message.get("threadId") or metadata.get("threadId"))
     linked_at_text = _iso(linked_at)
+    evidence_columns = _columns(conn, "application_email_evidence")
+    evidence_job_column = "job_id" if "job_id" in evidence_columns else "job_key"
```
ebarti: Keep this runtime on the exact-v7 identity contract. Probing for `job_id` and falling back to `job_key` reintroduces runtime table-shape detection and internal URL resolution, both explicitly prohibited after the v7 cutover. It also leaves `ApplicationAnchor.job_key` as `jobs.url`, so the scan response emits URL-shaped `jobKey` values rather than canonical JobIds. Please carry canonical `job_id` in the anchor, keep URL as a separate Gmail-search hint, and read/write/return `job_id` unconditionally; v6 handling belongs only in migration fixtures.

workers/automation/src/jobctrl/infrastructure/gmail/feedback.py
```diff
@@ -534,7 +548,11 @@ def _record_safe_event(
         "linkSignals": list(signals),
     }
     values: dict[str, Any] = {
+        "tenant_id": TENANT_ID,
         "job_url": job_key,
+        "job_id": _job_id_for_url(conn, job_key),
```
ebarti: Emit a canonical v7 event payload here. Although this adds the `job_id` column, the payload still contains legacy `jobKey: <URL>`; `canonicalSseEnvelope` rejects any payload with that alias, so this event is silently dropped and open clients do not invalidate after the scan. Emit canonical `jobId, preferably through the canonical event helper, and add a scanner-to-SSE regression.

ebarti (commented):
Thanks for this, it is a genuinely strong first contribution. You found the exact root cause of #780, and the tests are precisely how I like regressions pinned here: seeded against the exact v7 candidate schema, asserting the canonical writes end-to-end.

The three inline comments converge on one simplification that makes your diff smaller than what you wrote: this runtime only ever opens exact-v7 databases, so instead of probing schemas and branching, the code can assume v7 unconditionally:

1. Carry the canonical job_id on ApplicationAnchor (your anchor queries already join jobs, so it's one more selected column), keep the URL as a separate field used only as the Gmail-search hint, and drop _job_id_for_url entirely. Besides the identity-contract point in the inline comment, the helper hides a crash: it returns None when the URL doesn't resolve (e.g. the anchor's job was deleted mid-scan), and both application_email_evidence.job_id and application_outcome_suggestions.job_id are NOT NULL on v, so the scan would die on an IntegrityError, which is the same class of failure you're fixing.
2. Scope every join and filter by (tenant_id, job_id) per the inline cross-tenant comment, that's the composite key everywhere in this schema.
3. Emit the canonical event payload (jobId, no legacy jobKey alias) per the inline SSE comment, so the scan's invalidation event actually reaches open clients.

Two small things while you're in there: identity_version is hardcoded 1 with a commen, import the constant if it's exported rather than mirroring it; and don't worry about this module's pre-existing column probing (_column_expr etc.) or its legacy-shaped ensure_*/test, that's older debt, we'll track a full v7 cleanup of this module as a follow-up issue, it's not on you.

I'll approve the CI run on your branch after the fixes. Happy to merge once the v7-only shape lands. Thanks again; this fixes a real breakage and the test discipline made it easy to review.

YiFanWangSCU:
Thanks for the detailed review. I’ve addressed the v7 identity issues in a follow-up commit:

- `ApplicationAnchor` now carries the canonical `job_id`; the job URL is kept only as a Gmail search hint.
- Outcome and apply-run anchors join and filter by `(tenant_id, job_id)`, with shared-JobId cross-tenant regressions.
- Evidence, suggestions, scan summaries, contracts, and web invalidation use `jobId` end to end; the URL lookup helper and runtime schema fallback added by this PR are removed.
- Gmail feedback events now go through `record_job_event()`. Paired regressions cover the runtime boundary: the production Python scanner persists the canonical event shape, and the API streams that stored shape through SSE without legacy identity aliases.

The focused Python, domain-types, web invalidation, API, RPC contract, typecheck, Ruff, and diff checks all pass.

workers/automation/src/jobctrl/infrastructure/gmail/feedback.py
```diff
@@ -521,33 +522,23 @@ def _record_safe_event(
...
-        "occurred_at": _iso(occurred_at),
-        "payload_json": json.dumps(payload, sort_keys=True),
-    }
-    names = [name for name in values if name in columns]
-    if not names:
-        return
-    conn.execute(
-        f"INSERT INTO job_events ({', '.join(names)}) VALUES ({', '.join('?' for _ in names)})",
-        tuple(values[name] for name in names),
+    record_job_event(
```
ebarti, line 569: PR Reviewer - pr-reviewer

Severity: High
Gate: Fail
Evidence: `record_job_event()` (workers/automation/src/jobctrl/state.py:1274-1288) synchronously publishes on the process-wide bus while this scan's *private* `sqlite3.connect` handle (feedback.py:128) still holds the uncommitted evidence/suggestion/event INSERTs (`conn.commit()` only runs at feedback.py:221). The `jobctrl rpc` child that serves `GmailFeedbackScan` runs `cli._bootstrap()` (cli.py:2178 -> cli.py:121-126), which wildcard-subscribes a `ProjectionBuilder(conn_factory=get_connection)`. Its `_on_event` therefore refreshes on a *different* connection than the writer, hits the WAL write lock held by this scan, waits the full `busy_timeout=10000`, then fails with `sqlite3.OperationalError: database is locked` (logged as `ProjectionBuilder failed to refresh after ApplicationEmailFeedbackIngested`). The builder's own comment at projection_builder.py:1027-1034 (`defer_commit`) documents that the subscriber is expected to share the producer's connection; main's direct INSERT never published, so this stall is new.
Reproduction: Reproduced with command/steps — sealed v9 DB (`create_exact_v9_database`), one applied job, `ProjectionBuilder(conn_factory=get_connection).subscribe_to(get_default_publisher())` exactly as `_bootstrap` does, then `scan_gmail_feedback(..., client=FakeGmailClient(2 linked messages))`: scan took 21.4 s for 2 linked messages (~10 s per message), two `database is locked` tracebacks (`replace_evidence_usage_rows` DELETE, then `_rebuild_job`/`upsert_job_list`), and the projection refresh for both events failed. Rows were persisted, so the cost is per-linked-message latency (a 20-link scan adds ~200 s), ERROR tracebacks in the worker log, and job projections that only catch up on the next unrelated event.

Every linked Gmail message now blocks the RPC worker for the busy timeout and logs a lock error, because the event is published from a foreign connection mid-transaction.

Required fix: Publish from the same handle the wildcard subscriber uses, or publish only after commit. Concretely, either run the scan on `jobctrl.database.get_connection(db_path)` (the thread-local handle `_on_event` reuses, so the refresh joins the open transaction under `defer_commit` and the scan's `commit()` flushes both), or keep the private connection but pass an explicit `publisher` to `record_job_event` and dispatch the event after `conn.commit()`. Add a regression that wildcard-subscribes a `ProjectionBuilder` like `_bootstrap` and asserts the scan completes with no `database is locked` error (and that the builder sees the new event).

workers/automation/src/jobctrl/infrastructure/gmail/feedback.py
```diff
@@ -706,6 +711,7 @@ def _anchor_query(anchor: ApplicationAnchor) -> str:
     hints: list[str] = []
     hints.extend(_name_tokens(anchor.company))
     hints.extend(_title_tokens(anchor.title))
+    hints.extend(_application_url_tokens(anchor.job_url))
```
ebarti: PR Reviewer - pr-reviewer

Severity: Medium
Gate: Pass
Evidence: This line is new (main's `_anchor_query` never used the job/listing URL). `_safe_query_hints` in gmail/client.py:252-263 keeps only the first 6 hints, and `_application_url_tokens` keeps dotted hostnames whole, so the listing-site tokens inserted here displace the application-domain and ATS hints that follow. Computed on head vs main for a LinkedIn-discovered job (`https://www.linkedin.com/jobs/view/4123456789/`, applied via `https://jobs.lever.co/acme/1b2c3d4e`): main sends `['Acme','Staff','jobs.lever.co','1b2c3d4e','lever']`; head sends `['Acme','Staff','www.linkedin.com','view','4123456789','jobs.lever.co']`. The `lever` ATS hint and the posting id drop out; for the test fixture itself `greenhouse` drops out. Listing-host tokens (`www.linkedin.com`, `view`, a numeric id) never appear in employer mail, so this only removes useful OR terms.
Reproduction: Reasoned from code, hint lists computed by calling `_anchor_query` + `_safe_query_hints` on both trees.

Adding the job URL as a search hint ahead of the application URL / ATS hints silently changes every Gmail query under the 6-hint cap and lowers recall for the emails the linker is designed to catch.

Required fix: Either drop the job-URL hint (it was never part of the search on main, and the owner's ask was only that the URL not be identity) or append `_application_url_tokens(anchor.job_url)` after the application-URL and ATS hints so it only fills leftover slots. Add a unit test for `_anchor_query` that pins the hint order under the cap.

workers/automation/tests/test_gmail_feedback.py
```diff
@@ -12,15 +12,21 @@
...
 
 from jobctrl.infrastructure.gmail.feedback import (
+    _apply_run_anchors,
+    _outcome_anchors,
     classify_outcome,
     ensure_application_feedback_tables,
     scan_gmail_feedback,
 )
+from jobctrl.infrastructure.migrations.schema_v7 import (
+    create_unstamped_exact_v7_candidate,
```
ebarti: PR Reviewer - pr-reviewer

Severity: Low
Gate: Pass
Evidence: The runtime opens only exact v9 (`open_exact_v9_database` in workers/automation/src/jobctrl/database.py:229-243 raises `SchemaMigrationRequiredError` for v6/v7/v8 and asserts `EXACT_V9_MANIFEST`). All new regressions and `seed_feedback_db` build the database with `create_unstamped_exact_v7_candidate`. Today the `jobs`, `application_outcomes`, `application_email_evidence`, `application_outcome_suggestions`, `apply_run_projections`, and `job_events` DDL is byte-identical between the v7 and v9 constructors (checked via both frozen constructors), so the tests are valid now, but they pin a schema the runtime refuses to open.

The #780 regressions should prove the scan on the sealed schema the runtime actually opens, so future v10 drift in these tables is caught.

Required fix: Import `create_unstamped_exact_v9_candidate` from `jobctrl.infrastructure.migrations.schema_v9` for the new tests and `seed_feedback_db` (rename the `*_v7_*` test names accordingly).

ebarti (commented):
Follow-up review of 2ee90c2 (cumulative diff vs `main` @ ed5e530). All three owner threads are resolved in this revision: `ApplicationAnchor` carries canonical `job_id` with the URL as a separate field, no `_job_id_for_url`/schema fallback remains; all three anchor sources join and filter on `(tenant_id, job_id)` with shared-JobId cross-tenant regressions; the event goes through `record_job_event()` and both the Python and SSE tests pin the alias-free `jobId` payload.

Local verification in a detached worktree at the head SHA: `test_gmail_feedback.py` + `test_domain_event_parity.py` 20 passed; 14 related Python modules 219 passed; ruff clean; the new tests fail against `main` for the right reasons (`no such column: o.job_key`, and the apply-run cross-tenant test returns 2 anchors); domain-types 179 passed, api `server`/`application-feedback`/`rpc-contracts`/`endpoint-client` 318 passed, web `invalidation-router` 101 passed; tsc clean for domain-types, contracts, api-client, api, web; `git diff --check` clean; both commits carry `Signed-off-by`.

Three new findings inline: one High (the `record_job_event` publish now runs a `ProjectionBuilder` refresh on a foreign connection mid-transaction in the `jobctrl rpc` child, reproduced as a ~10 s `database is locked` stall per linked message), one Medium (job-URL search hint displaces ATS hints under the 6-hint cap), one Low (regressions pin the v7 candidate while the runtime opens only exact v9).

Two non-blocking notes: the PR description still describes the first commit's schema-probing/fallback approach and should be rewritten for the v7-identity shape; and `docs/api/complete-contract.md` (line 967) still says the scan response returns "job keys" where it now returns canonical `jobId` values.

YiFanWangSCU:
Thanks for the follow-up review. I addressed all three findings in `88fc43bf`:

- The Gmail scanner now buffers each canonical domain event, commits the evidence/suggestion/event transaction, and only then publishes to the default bus. The regression uses the production topology—scanner-private SQLite connection plus `ProjectionBuilder(conn_factory=get_connection)` on another connection—and asserts two linked messages publish twice with no `database is locked`, a materialized job projection, and a watermark caught up to the latest event.
- Application URL and ATS hints now precede the listing URL, with `_anchor_query()` tested through the real six-item cap.
- All Gmail scanner fixtures and versioned test names now use the current exact-v9 candidate schema.

During final fault-injection review, I also found and fixed a related mid-scan deletion race. The scanner now enables and verifies SQLite foreign keys, keeps Gmail I/O outside the write transaction, then uses a short `BEGIN IMMEDIATE` transaction to recheck the tenant-scoped job and provider message before writing. Regressions cover a permanent delete during body retrieval (including same-JobId tenant isolation and no retained private body/orphans) and two scanners that concurrently pass the initial duplicate check.

I also updated the API documentation to say canonical `jobId` values and rewrote the PR description to match the current implementation.

Validation is green: the full Python worker suite passed 3,716 tests with 2 expected opt-in system-browser skips, plus 52 focused Python tests, 179 domain-types tests, 318 API tests, 101 web invalidation tests, five TypeScript package checks, full-worker Ruff, and `git diff --check`. The two concurrency regressions passed 100 independent rounds and both fail against the previous local head for the expected stale-write and uniqueness reasons; the projection regression was likewise verified against the previous public head, where it reproduces the lock error and missing projection/watermark.

PR #830, opened 2026-08-17
fix: preserve achievement fidelity in resume tailoring

ebarti:
## Summary

- make the configured per-role bullet limit a hard ceiling and select the smallest sufficient evidence-backed achievement set
- bind each tailored bullet and numeric claim to one exact source achievement, preventing cross-achievement metric borrowing
- preserve agency, scope, causality, outcomes, and professional register through voice edits, then rerun final provenance, fabrication, quality, and judge gates
- remove the separate editable metric field while preserving unmatched legacy values as non-authoritative compatibility data
- project the final judge and voice-audit details through the API/read model and document the updated tailoring contract

## Root cause

Requirement coverage and positioning logic could promote too many achievements, including overflow beyond the configured maximum. Numeric grounding used a profile-wide compatibility corpus, so a number present in one achievement could incorrectly authorize another. The voice pass could also rewrite already-clean text without rerunning every acceptance gate against the final artifact.

## User impact

Tailoring now favors a concise, job-relevant subset instead of treating the bullet maximum as a target. Metrics stay attached to the achievement that proves them, explicit required bullets remain the only hard pins, and post-voice artifacts must retain the source meaning before they can be accepted.

## Validation

- `corepack pnpm check`
- full API suite: 816 tests passed
- full web suite: 2,037 tests passed; production build passed
- profile editor E2E: 4 tests passed
- exact-current profile/validator worker suites: 117 tests passed
- data-integrity regressions: 11 tests passed
- focused independent review matrix: 349 Python and 237 API tests passed
- documentation build passed
- independent reviewer and QA gates: PASS with no findings

The broader local Python run reported 3,694 passed and 2 skipped; its only failure was a cutoff-policy check caused by an unrelated unstaged `workers/automation/uv.lock` edit that is not part of this PR.


apps/api/src/read-model.ts
```diff
@@ -3512,6 +3512,9 @@ function parseVoicePass(value: string | null): VoicePassAudit | null {
     promptVersion: metadataText(record.prompt_version, 64) ?? "",
     proxyDelta,
     reason: metadataText(record.reason, 600) ?? "",
+    summaryRejectionReason: metadataText(record.summary_rejection_reason, 600) ?? "",
+    scopeViolations: metadataTextList(record.scope_violations).slice(0, 20),
+    finalJudge: metadataRecord(record.final_judge),
```
ebarti: PR Reviewer - pr-reviewer

Severity: High
Gate: Fail
Evidence: `metadataRecord` is a raw passthrough (`apps/api/src/read-model.ts:3811`), unlike every neighboring audit field which is bounded/redacted (`metadataText` 120-600 chars; adversarial prompts go through `parseAdversarialAudit` -> `metadataPromptText` with a 2400-char cap and secret redaction, read-model.ts:3699-3729). The Python side persists `VoicePassRecord.final_judge` containing `_judge_record(...)` plus `adversarial_review.to_dict()` when the post-voice adversarial review runs (`use_cases.py:3916-3933`), and `AdversarialReviewRecord.to_dict()` embeds `llm_audit.prompt_messages` with the FULL prompt - `tailoring_plan.to_prompt_dict()`, the rendered resume, and the job text (`adversarial.py:283-303`, `build_adversarial_review_prompt`). That blob is serialized into `voice_pass_json`, denormalized onto every provenance row (`bullet_provenance_repository.py:110-151`), preserved verbatim by `normalizeVoicePassJson` (`projections.ts`, `final_judge: record(parsed.final_judge)`), and served raw here via `explanation.voicePass` (read-model.ts:2841).
Reproduction: Reasoned from code; the parity fixture only exercises `final_judge: {}` so no test covers a populated record.

docs/architecture/tailoring.md (retained by this PR) states the requirement-led audit surface "must not expose raw prompts, full profile payloads, full job descriptions...". For a high-fit job with a voice pass, `voicePass.finalJudge.adversarial_review.llm_audit.prompt_messages` now carries exactly that - unbounded and unredacted - bypassing the read model's established bounding discipline, and duplicating a multi-KB prompt blob onto every provenance row.

Required fix: strip `llm_audit` (or the whole `adversarial_review`) before storing it into `VoicePassRecord.final_judge` in `_voice_and_audit`, and/or sanitize here: bound scalar judge fields with `metadataText`/`metadataTextList` and route `final_judge.adversarial_review` through the existing `parseAdversarialReview`/`parseAdversarialAudit` sanitizers. Add a guard test asserting no prompt text appears under `voicePass.finalJudge` in the artifact explanation.

workers/automation/src/jobctrl/domain/materials/services.py
```diff
@@ -231,12 +233,14 @@ def _validate_master_json_fields(
...
-            bullet_count=len(bullets),
-        ):
+        effective_bullets = [str(bullet).strip() for bullet in bullets if str(bullet).strip()]
+        seen_effective = {_normalize_output_term(bullet) for bullet in effective_bullets}
+        for required_bullet in required_bullets_by_entry.get(entry_id, ()):
+            normalized_required = _normalize_output_term(required_bullet)
+            if normalized_required and normalized_required not in seen_effective:
+                effective_bullets.append(required_bullet)
+                seen_effective.add(normalized_required)
+        if len(effective_bullets) > max_bullets:
```
ebarti, line 243: PR Reviewer - pr-reviewer

Severity: Medium
Gate: Pass
Evidence: `effective_bullets` counts every configured required bullet whether or not the model emitted it, then hard-errors above `max_experience_bullets`. Verified against this branch with the real `ContentValidator`: a profile with 5 required bullets and `max_experience_bullets: 4` fails with "Experience update 'acme_swe' exceeds 4 bullets" both when the payload includes all 5 required bullets and when it includes only 4 (the 5th is appended and counted). The renderer additionally ships only 4 of the 5 user-required bullets (`tailored_experience_bullets` -> `kept_required[:max_bullets]`).
Reproduction: Reproduced with an inline script against this worktree (both payload shapes fail; renderer output showed 4 of 5 required bullets).

When `len(required_bullets) > max_experience_bullets` for a role, the generation contract is unsatisfiable: no model output can pass, yet the attempt loop still burns `max_retries+1` attempts across every candidate model before reporting a per-candidate "exceeds N bullets" error that reads like a model failure rather than a profile configuration conflict. Before this PR such profiles were servable via the mandatory-overflow path, so this is a newly unwinnable state.

Required fix: add a deterministic feasibility pre-check (before any LLM call) that fails fast with an explicit configuration error naming the conflicting role, required-bullet count, and ceiling - and/or reject the conflicting combination at profile save. The same pre-check should cover pins + mandatory coverage jointly exceeding the ceiling for one role.

workers/automation/src/jobctrl/resume_profile.py
```diff
@@ -508,16 +484,11 @@ def tailored_experience_bullets(entry: dict, update: dict, profile: dict) -> lis
...
-    # trim requirement-covered or pinned bullets after that validator allows it.
-    if isinstance(update, dict) and update.get(_MANDATORY_BULLET_OVERFLOW_KEY):
-        return bullets
-
     required_norm = {_normalize_text(bullet) for bullet in required}
     kept_required = [bullet for bullet in bullets if _normalize_text(bullet) in required_norm]
     kept_other = [bullet for bullet in bullets if _normalize_text(bullet) not in required_norm]
     if len(kept_required) >= max_bullets:
-        return kept_required
+        return kept_required[:max_bullets]
```
ebarti, line 549: PR Reviewer - pr-reviewer

Severity: Medium
Gate: Pass
Evidence: `experience_updates_by_id` no longer injects the mandatory-overflow allowance and this cap now trims unconditionally. Legacy artifacts approved before this PR under the mandatory-overflow contract persisted a reviewed `.txt` and provenance rows carrying more than `max_experience_bullets` bullets. Any post-upgrade render from the stored payload - `RenderPdfUseCase.execute` (use_cases.py:4688, runs whenever `resume_pdf` is missing/unapproved) or the template-refresh `resume_pdf` path - now re-assembles through this cap and silently drops the overflow bullets.
Reproduction: Reasoned from code; the flipped tests (`test_provenance_caps_covered_bullets_to_match_shipped_resume`, `test_html_pdf_renderer_caps_requirement_covered_overflow_like_txt`) prove trimming but only exercise same-generation parity, not a legacy artifact whose persisted `.txt`/provenance predate the ceiling.

For those legacy generations the submitted PDF becomes weaker than the reviewed plain-text/HTML artifact and than the persisted provenance rows - the exact "submitted PDF weaker than the reviewed .txt" divergence class the regression catalog row (updated in this PR) says must not happen. The Apply Review rename to "Legacy bullet-limit violation" labels the state but does not prevent the silent trim on re-render.

Required fix: keep render-only refresh faithful to the accepted artifact for legacy generations - e.g. render legacy payloads from the persisted provenance/text bullet set, or version tailored payloads and preserve the overflow allowance for generations accepted before the hard ceiling; alternatively force re-tailor (never silent re-render) for over-cap legacy payloads, and add a regression test for a legacy overflow payload.

apps/api/src/profile-store.ts
```diff
@@ -691,7 +696,7 @@ function isMaterializedLegacyBulletEvidence(evidence: Record<string, unknown>, e
   return text(evidence.action).trim() === sourceText
     && text(evidence.outcome).trim() === sourceText
     && asTextArray(evidence.tools).length === 0
-    && arraysEqual(asTextArray(evidence.metrics), extractedLegacyBulletMetrics(sourceText))
+    && arraysEqual(asTextArray(evidence.metrics), extractedAchievementMetrics(sourceText))
```
ebarti: PR Reviewer - pr-reviewer

Severity: Medium
Gate: Pass
Evidence: `isMaterializedLegacyBulletEvidence` now requires stored `metrics` to equal the NEW extractor's output. Rows materialized before this PR were written with the old `LEGACY_BULLET_METRIC_PATTERN`, which extracted no bare numbers, no count units (users/customers/teams/...), no GBP/EUR currency, no `24/7`, no `+` suffixes. For any such bullet (e.g. "Scaled the platform to 200 users.": old metrics `[]`, new extraction `["200 users"]`) the stored row fails this equality, is treated as user-authored evidence, and `achievementEvidenceForEntry` keeps it verbatim forever instead of re-deriving it. The Python mirror has the same break: `_is_materialized_legacy_bullet_evidence` compares against `_legacy_bullet_metrics(source_text)` which now calls `extract_achievement_metrics` (workers/automation/src/jobctrl/resume_profile.py:384/440).
Reproduction: Reasoned from code; existing rederive tests only cover rows created under the new pattern.

Consequence: when the user later edits that bullet, the stale evidence row (old source_text, old metrics) is preserved as claim authority - its text stays in the fabrication corpus and its tokens stay claimable via `merge_achievement_metrics` - directly contradicting this PR's invariant that "stale derived values disappear when their achievement changes" (docs/api/profile-and-settings.md).

Required fix: recognize legacy materialized rows by accepting metrics equal to EITHER the new extraction OR the old pattern's extraction of the same source_text (keep the old regex for recognition only), or run a one-time re-materialization of auto-derived evidence rows on load/save; apply the same fix to the Python `_is_materialized_legacy_bullet_evidence`.

workers/automation/src/jobctrl/domain/materials/use_cases.py
```diff
@@ -3522,18 +3747,24 @@ def _voice_and_audit(
...
-        adopted when it both improves voice AND survives the SAME grounding gate, so
-        a voice edit that injects an unsourced metric/title/date/employer (VOICE-03)
-        or regresses voice never reaches the user.
+        adopted only when it edits an eligible line, reduces buzzwords, and survives
+        mapping, rendering, quality, provenance, fabrication, fit, judge, and any
+        applicable adversarial gate. Semantic or register regressions therefore
+        fall back to the already accepted candidate.
         """
+        if base_verdict is None:
+            base_verdict = JudgeVerdict.failed(notes="selected candidate has no judge verdict")
```
ebarti: PR Reviewer - pr-reviewer

Severity: Low
Gate: Pass
Evidence: `base_verdict` is None only on already-failed paths (failed_validation / parse errors / exhausted retries; every accepted candidate carries a real verdict, and lenient mode synthesizes a PASS at use_cases.py:3152-3153). This synthetic `JudgeVerdict.failed(...)` then flows back through `execute` (`verdict = final_verdict`, line 2270) into `judge_record = self._judge_record(verdict)` and is persisted as `metadata["judge"]`, `metadata["final_judge"]`, and `report["final_judge"]` - including `judge_model` filled from the configured policy (use_cases.py:3422).

Before this PR these failed artifacts persisted `judge: null`. Now they carry a judge FAIL record naming a judge model that never evaluated the candidate, with the placeholder note surfaced as an "issue". On an audit surface governed by the repo's judge-audit-fidelity rules, a fabricated-looking judge verdict is a truthfulness regression, even if confined to failed artifacts.

Required fix: keep `judge_record = None` when no judge ran (persist null as before), or mark the synthetic record unambiguously (e.g. `verdict: "NOT_RUN"`, no `judge_model`) so the read model cannot render it as a real judge decision.

ebarti (commented):
5 findings (1 High, 3 Medium, 1 Low): unbounded post-voice final_judge audit (including full adversarial prompt messages) passes unsanitized through the voicePass read model; a required-bullets-over-ceiling profile now deterministically fails every attempt with no fail-fast; legacy approved overflow artifacts re-render weaker than the reviewed material; the metric-pattern swap breaks recognition of previously materialized bullet evidence in both runtimes so stale evidence stops re-deriving; failed paths persist a synthetic judge FAIL record for a judge that never ran.

PR #831, opened 2026-08-18
feat(observability): expose provider failure diagnostics

ebarti:
## Summary

- preserve bounded, privacy-safe provider failure details before SDK exceptions collapse them
- distinguish provider failures from JSON parse failures in tailoring audit records and existing telemetry spans
- correlate diagnostics to the exact Temporal execution and expose aggregated counts and latest failure details in run views
- distinguish automatic retryability from operator-initiated Tailor recovery
- provide a narrowly labeled compatibility bridge for historical structured-output builder failures

## Why

Provider turn failures were flattened into generic runtime exceptions, then recorded as parse errors. The audit records were not projected into workflow-run details, leaving operators with only a terminal attempt-budget error and no explanation of the underlying repeated failures.

## User impact

Run details now show the provider failure count, safe diagnostic dimensions, structured-details availability, and whether manual recovery is available. Prompts, outputs, arbitrary exception messages, paths, URLs, and secrets remain excluded.

## Validation

- 177 focused worker tests passed
- 9 API read-model tests passed
- 9 run-drawer tests passed
- contracts, API, and web typechecks passed
- targeted Ruff checks passed
- documentation build passed
- production web build passed
- independent review and QA gates passed with no Blocker or High findings

## Stack

This PR is based on #830 and should be reviewed after it.

PR #832, opened 2026-08-19
feat(profile): extend Plate editing, ordering, and PDF export

ebarti:
## Stack

Depends on #831 and remains based directly on `fix/provider-error-observability` at the parent branch head.

## Summary

- Add a shared **Export PDF** control to every production Plate editor: Profile baseline, Preferences template preview, and Apply Review tailored resume.
- Export the currently mounted browser document while preserving active typography, colors, margins, and A4 or Letter geometry.
- Use the browser-rasterized page as the visible PDF layer so Unicode glyphs, punctuation-adjacent spacing, line wrapping, and active font metrics are not re-typeset through jsPDF built-in fonts.
- Add a separate invisible normalized text layer so the visually faithful PDF remains searchable and extractable.
- Remove editor, audit, comment, and selection chrome from the export clone; prevent off-page audit overflow from changing PDF scale; avoid blank trailing pages and text-line clipping at page boundaries.
- Keep PDF export browser-local behind `PdfExportPort`; the export action does not regenerate or replace an artifact, call a mutation API, or change approval state.
- Project direct Profile Plate edits with an unambiguous canonical owner—full name, executive profile, position summary, and experience bullet—into the existing validated Profile form draft, including its normal undo, save, and autosave path. Reconcile split Plate bullet blocks as ordered canonical insertions without replacing existing bullets or dropping non-rendered trailing bullets.
- Present relative font size as a high-contrast percentage (`100%` at the resume default) instead of exposing the internal unitless scale.
- Make the saved Candidate Profile experience sequence the single order used by baseline and tailored resume text/HTML/PDF.
- Add accessible move-up/move-down controls plus an explicit newest-first action; all reorder operations use the existing validated Profile save/autosave path.
- Replace flattened ultra-wide experience rows with bounded, numbered role cards whose controls remain visually attached to their entry.
- Use `Company - Job Title` for each experience-card label, place role actions in one full-width compact row, and keep Start, End, Present, and Clear in one compact date row with narrow-container wrapping.
- Render no-bullet roles without an empty list or the full bullet-bearing entry gap.
- Render education as institution/year above the degree with baseline-aligned dates.
- Add an optional per-position summary to structured Profile experience entries and render it below the role/date row and before achievement bullets in Plate, submitted HTML/PDF, and canonical resume text.
- Preserve exact Apply Review source-field mapping and fabrication grounding for position summaries so reviewed text, generated artifacts, provenance, and downstream cover-letter safety checks stay aligned.
- Advance the runtime contract additively to exact schema v9. Exact-v8 installs add the summary column directly; retained v6/v7 installs compose through private intermediates under the existing stopped-runtime paired-backup lifecycle.
- Remove v9 candidates, private intermediates, nested intermediates, and SQLite sidecars during normal rollback and interrupted-transition recovery.
- Update the owning runtime, storage, Profile, frontend-state, PDF-export, QA, requirements, and architecture-decision documentation.

### Current head `a06f42110`

- `corepack pnpm --filter @jobctrl/web check`: PASS.
- Focused Profile form, editor, accessibility, and layout-contract regressions: PASS (63 tests across 5 files), including repeated-input Plate bullet insertion, direct text edits, non-rendered bullet preservation, card labels, saved reordering, full-width compact action rows, and full-width compact date rows.
- `corepack pnpm --filter @jobctrl/web build`: PASS.
- `git diff --check`: PASS.
- Browser E2E, browser screenshots, and manual PDF comparison were intentionally not run; final visual/product QA is user-owned.

### Earlier cumulative gates

- Focused PDF adapter and Apply Review export regressions: PASS (5 tests across 2 files; 54 unrelated tests skipped by the focused filter).
- Focused Profile editor and PDF adapter regressions: PASS (13 tests across 2 files).
- `corepack pnpm docs:build`: PASS; 9,653 emitted references resolve and redirect checks pass.
- `node scripts/distribution-manifest.mjs audit`: PASS after recording the explicit browser-rasterizer runtime dependency.
- Focused Python Ruff gate: PASS.
- Resume renderer and assembler regressions: PASS (28 tests), including saved-sequence parity, compact no-bullet HTML, and institution/year-above-degree HTML/text parity.
- Earlier PDF adapter and all three Plate editor integration suites: PASS (65 focused tests), including A4/Letter geometry, inherited template styles, audit-chrome removal, overflow-stable scaling, live content, failure handling, and no trailing blank page.
- API Apply Review source-field regression: PASS (38 tests); `corepack pnpm api:check`: PASS.
- Exact-v9 migration/profile persistence regressions: PASS (37 tests).
- Fabrication grounding and exact-schema regressions: PASS (81 tests), including a summary-only metric that previously false-failed.
- Domain type mirror: PASS (`@jobctrl/domain-types` check and 179 tests).
- Launcher v9 promotion/rollback/recovery tests under the race detector: PASS; `go vet ./...`: PASS.
- Independent code review: PASS on prior cumulative head `7c014ab0d`. Later UI/PDF fidelity deltas are not represented as independently reviewed.

Fresh exact-head GitHub CI is running. Final product QA remains explicitly user-owned and is not represented as complete here.

## Local-worktree note

A pre-existing local-only `workers/automation/uv.lock` policy edit remains unstaged and is excluded from every commit. The committed lockfile used by clean CI is unchanged.

## Dependency audit

The repository-wide production audit still reports advisories inherited from the parent branch. The PDF path resolves to the patched DOMPurify version pinned by this PR; no profile, resume, PDF, database, browser data, logs, or local paths are committed or included in PR metadata.

## Checklist

- [x] Conventional commits and PR title.
- [x] Relevant focused checks are listed above.
- [x] No secrets or generated/private user artifacts are included.
- [ ] Exact-head CI complete.
- [ ] Current delta independent review complete.
- [ ] Final product QA complete (user-owned).


apps/web/src/contexts/materials/components/ResumeAuditPins.tsx
```diff
@@ -2713,6 +2745,9 @@ function ResumePlateDocument({
           data-rendered-line-count={lines.length}
           role="textbox"
           spellCheck={false}
+          onInput={(event: FormEvent<HTMLDivElement>) => {
+            onSemanticTextChange?.(resumeSemanticTextChangesFromElement(event.currentTarget));
```
ebarti: PR Reviewer - pr-832-reviewer

Severity: High
Gate: Fail
Evidence: This `onInput` handler is the sole trigger for the Plate-to-Profile projection (`onSemanticTextChange` has no other production call site). slate-react 0.124.2 (the version resolved by this PR's pnpm-lock.yaml, bundled via platejs 53.1.2) handles editing through `onDOMBeforeInput` and calls `event.preventDefault()` for every input type except a single-character `insertText` matching `/[a-z ]/i` at a collapsed selection with `anchor.offset !== 0` and no active marks (the `_native` fast path in slate-react `dist/index.js`, `onDOMBeforeInput`). Per the Input Events spec, a canceled `beforeinput` produces no DOM mutation and no native `input` event, so React's `onInput` never fires for: backspace/delete, digits, punctuation, non-ASCII characters, paste, cut, Enter (the bullet-split gesture), the first character typed into an empty split node (offset 0), or typing with bold/italic marks active.
Reproduction: Reasoned from code (slate-react 0.124.2 source inspected in the resolved dependency; not reproduced in a live browser). The passing unit tests do not cover this because jsdom lacks `InputEvent.prototype.getTargetRanges`, so slate takes its legacy path there and `fireEvent.input`/`user.type` synthesize the very events production suppresses; the new e2e test exercises PDF export only, not projection.

In a real browser, an edit session whose final edits are deletion-only, digit-only, or punctuation-only (e.g. correcting a bullet metric "40%" to "45%": backspace fires `deleteContentBackward` and "5" is not `[a-z ]`, both preventDefault'ed) never runs the projection: the boxed field keeps the old text, the form never becomes dirty, autosave/save never fire, and the Plate edit silently vanishes on the next preview refresh. That contradicts the PR's core contract that direct canonical Plate edits "enter the same validated form draft and follow its normal autosave", and it fails silently with no unsaved-changes indicator.

Required fix: drive the projection from Plate's model-change pipeline instead of native DOM `input` — e.g. run `resumeSemanticTextChangesFromElement` in an effect keyed on the committed Plate value (the component already receives `onValueChange`), or on the editor's change event after React commits. Add a browser E2E (profile-edit.spec.ts) that deletes text and types digits in a bullet, then asserts the boxed field and the Unsaved-changes state update.

apps/web/src/contexts/profile/forms/profile-form.tsx
```diff
@@ -147,10 +264,43 @@ export function ProfileForm({
...
   });
 
+  const applyPlateTextChanges = useCallback(
+    (changes: readonly ProfilePlateTextChange[]) => {
+      const currentProfileText = form.state.values.profileText;
+      const previousProjection = plateProfileProjectionRef.current;
+      // Reapply Plate's complete semantic snapshot to one stable baseline.
+      // Otherwise every keystroke in an inserted split node would insert it again.
+      const baselineProfileText =
+        previousProjection?.lastAppliedProfileText === currentProfileText
```
ebarti: PR Reviewer - pr-832-reviewer

Severity: Medium
Gate: Fail
Evidence: profile-form.tsx:264-281 with profileTextWithPlateChanges (lines ~108-215). When `currentProfileText !== previousProjection?.lastAppliedProfileText` — i.e. the boxed Profile editor changed the draft since the last Plate projection — the baseline becomes the current draft, and the next Plate input re-applies Plate's complete semantic snapshot onto it. The snapshot values come from the mounted preview HTML, which only re-renders after a successful profile mutation (`useProfileHtmlPreviewUrl` is keyed on mutation count), so between a boxed edit and its autosave (5s debounce, longer if the mutation fails) the snapshot is stale relative to the draft.
Reproduction: Reasoned from code; not run. Both panes are simultaneously editable in the split Profile view.

Two concrete lost-update paths: (1) edit "Full name" (or executive profile, a summary, or a bullet) in the boxed pane, then type one lowercase letter in any Plate bullet before autosave completes — `replace()` sees the stale rendered name differing from the new draft value and reverts the unsaved boxed edit. (2) insert or reorder bullets in the boxed pane, then type in Plate — render-time bullet ordinals are applied against the newer draft's `baselineBullets` array, so e.g. a newly inserted bullet 0 is overwritten by the rendered text of the old bullet 0 even though that bullet was never touched in Plate.

Required fix: project deltas, not the full snapshot — capture the semantic snapshot once when the preview HTML mounts and apply only changes whose text differs from that render-time snapshot; or skip projection (and surface a refresh) when the current draft no longer matches the profile revision that produced the mounted HTML.

ebarti (commented):
Reviewed #832 vs fix/provider-error-observability (head a06f4211). The v9 migration chain (Go launcher + Python executors, rollback/recovery), schema manifests, position-summary plumbing, saved-order rendering, and the PDF export adapter all check out. Two findings on the new Plate-to-Profile text projection: it is driven by a native onInput event that slate-react suppresses for most edit types in real browsers (High), and its full-snapshot rebase can clobber newer unsaved boxed-editor edits (Medium).

PR #835, opened 2026-08-21
refactor(ci): simplify CI to path-filtered workflows and tame Dependabot

ebarti:
## Summary

- simplify CI to plain, self-triggering, path-filtered GitHub Actions workflows: each workflow declares the paths it owns and runs whole when a pull request or `main` push touches them — no planner, no aggregate, no reusable-workflow routing
- delete the CI meta-layer entirely: the route planner (`ci-plan.mjs`), the required-aggregate validator (`ci-required.mjs`), and every test that asserted the shape of workflows or CI config (`ci-plan.test`, `ci-required.test`, `ci-action-pins.test`, `dependabot-config.test`, `stacked-ci-workflows.test`, `check-docs-workflow-contract`, `check-python-workflow-contract`, the API `typescript-ci-workflow-contract` test, and the demo-site workflow-shape assertions), plus `.github/actionlint.yaml`, whose only purpose was suppressing actionlint's rejection of the preview `stacked` event the planner depended on
- cover every package-manager surface in Dependabot with tamed cadence: weekly staggered per ecosystem (npm Tue, uv Wed, gomod Thu, actions Fri), at most 3 open PRs each, 30/7/3-day major/minor/patch cooldowns, and grouped minor+patch updates (React runtime and Storybook/MSW and OpenTelemetry families grouped to avoid known-broken partial graphs)
- test Python dependency PRs against the exact `uv.lock` candidate graph: all CI installs use pinned uv 0.11.7 with `--locked`, never floating `pyproject.toml` ranges
- add a small `Distribution` workflow that runs the fail-closed release-contract audit (`distribution:audit`) when dependency locks or packaging inputs change — the surface Dependabot PRs touch
- bound every ephemeral Temporal test-environment lifecycle: all 74 `start_time_skipping` and 2 `start_local` call sites now go through `tests/temporal_env.py`, which retries a wedged startup with a fresh server (30s bound, 3 attempts) and abandons a wedged shutdown (30s) without overturning the test's own outcome; Python CI additionally uses a `/dev/shm` TMPDIR when available and reports the 25 slowest tests
- run all repo script unit tests via `scripts:test` in the new always-on `repo-scripts.yml` — the suite asserts contracts across inputs that sprawl the repository (docs launch copy, `launcher.go` version parity, release-workflow pins), so like the privacy gate it takes no paths filter — and drop the `scripts:test:{ci-meta,distribution,docs,demo}` subset aliases
- keep PR 835's genuine fixes: immutable Node 24-compatible action pins, per-test 3-minute pytest timeouts, the `release-check` locked tool group for the always-on privacy gate, and upstream Python compat fixes

## Why

The previous topology (and this PR's first iteration of it) made CI both unreliable and absurdly self-referential:

- Python CI installed floating `pyproject.toml` ranges, so a dependency PR such as #813 did not test the version it proposed — fixed here with `--locked` everywhere
- a 672-line diff-classifying planner, a schema-validated aggregate, and a `meta` CI job existed only to orchestrate and verify the CI itself, backed by tests that shell out to Ruby to parse workflow YAML and assert its shape ("tests for tests")
- the planner keyed on the preview `stacked` pull-request event and `pull_request.stack` context, and suppressed the actionlint error saying that event is invalid
- `main` has no branch protection or ruleset, so the "stable required check context" the aggregate existed to provide is not enforced anywhere

Plain path filters give the same cost control with zero machinery: docs-only PRs run the docs build, Python PRs run the Python suite, lockfile PRs run the distribution audit, and every stacked layer scopes itself to its own diff automatically. The privacy gate (`release-check.yml`) intentionally keeps no paths filter — any file in a public repo can carry PII — and its regression test stays because it guards a real safety invariant, not workflow aesthetics.

## Python CI hang root cause

Two `main` lanes (runs 31428971570 and 31596968394) were cancelled at exactly GitHub's six-hour default ceiling. Both logs end mid-`test_activity_enrich.py` with the runner reaping orphaned `pytest` and `temporal-test-server-sdk-python-1.31.0` processes: an intermittent, binary wedge inside a time-skipping `WorkflowEnvironment` start while the product path was fully mocked (an identical environment had started successfully 100ms earlier in the same process). Those runs installed floating deps via pip (temporalio 1.31.0); CI now installs the locked graph (1.26.0), and the bounded lifecycle self-heals the wedge under any SDK version. A separate slow-lane pathology — hosted-runner disk latency stretching the SQLite migration family from ~50s to ~620s — is addressed by the memory-backed TMPDIR; the macOS launcher workflow still proves the migration boundary on real disk. Per-test `pytest-timeout` (180s, `timeout_method = "thread"`: it dumps every thread's stack and hard-exits, because a signal cannot unwind a blocked worker-pool thread — proven by the pre-existing enrichment execution-lease `BEGIN IMMEDIATE` deadlock this PR surfaced and filed as #836 — with faulthandler at 240s as a second-chance dump) and 45-minute job ceilings remain the guarantees that a lane always completes.

## Validation

- new `test_temporal_env.py` lifecycle tests (wedged start retried, bounded failure after 3 attempts, wedged shutdown abandoned) — 3/3 passed; full temporal-marked set through the new helper against real test servers — 197/197 in 27s; marker parity preserved (197 temporal before and after)
- `corepack pnpm scripts:test` — 139/139 passed (the 50 deleted tests were the meta-layer)
- `corepack pnpm api:test` — 812/812 passed after removing the workflow-contract test
- `actionlint` v1.7.12 with no config file — clean across all workflows
- `python scripts/release_check.py --strict-prompt` under the locked release-check group — passed over the rewritten workflows; its 55 self-tests passed
- `uv lock --check` — lockfile consistent; `corepack pnpm docs:build` — 9,653 references resolve; `git diff --check` — clean
- independent review gate: PASS on head a2c1932de after two High findings (script-suite routing gap, signal-method timeout) were fixed and verified by the reviewer; CI fully green on the head, including the previously wedged Python 3.11 lane at 3m02s versus a 45-minute cancellation

## Out of scope (follow-up candidates)

- `release-distribution.yml` (1,526 lines) and the distribution build/release scripts keep their current shape; they are the release pipeline, not PR CI
- if a merge gate is ever wanted, add a branch ruleset listing the per-workflow check contexts (none is configured today)


.github/workflows/typescript.yml
```diff
@@ -26,66 +13,168 @@ on:
       - "pnpm-workspace.yaml"
       - ".github/workflows/typescript.yml"
   pull_request:
+    paths:
```
ebarti, line 16: PR Reviewer - pr-reviewer

Severity: High
Gate: Fail
Evidence: `scripts:test` now runs only in this workflow's `api` job, but the suite asserts repo-state contracts over inputs these path filters do not include. `scripts/docs-launch-copy.test.mjs` reads `docs/index.md`, `docs/comparison.md`, `docs/user/*.md`, `docs/claims-ledger.md`, `docs/.vitepress/config.ts`, `README.md`, `ROADMAP.md`, `LAUNCH_CHECKLIST.md`, `.env.example`; `scripts/version-contract.test.mjs` reads `launcher/internal/launcher/launcher.go`; `scripts/distribution-release.test.mjs` reads `.github/workflows/release-distribution.yml` and `.github/workflows/sync-homebrew-tap.yml`. None of those paths trigger this workflow. `ROADMAP.md`, `LAUNCH_CHECKLIST.md`, and `.env.example` now trigger no workflow at all; `docs/**` triggers only the docs-site dead-link build; `launcher/**` lost the `scripts:test` step it had on main (launcher.yml step "Test pinned distribution build contracts"). On main, typescript.yml's bare `pull_request` trigger ran `scripts:test` on every PR, so all of these classes were exercised at PR time.
Reproduction: Reasoned from code — input list extracted from the `read()` calls in the three test files; trigger coverage from this file, docs-site.yml, launcher.yml, and origin/main's typescript.yml/launcher.yml.

A docs-only, README/ROADMAP-only, or launcher-only PR can now merge with all-green checks while violating `scripts:test` contracts (public copy vs claims ledger, launcher/package version parity, release-workflow pin and ordering assertions). main's `scripts:test` then turns red and the failure surfaces on the next unrelated apps/packages/scripts PR, attributed to the wrong change.

Required fix: route `scripts:test` by its own inputs while keeping CI plain — either add the missing paths (`docs/**`, `README.md`, `ROADMAP.md`, `LAUNCH_CHECKLIST.md`, `.env.example`, `launcher/**`, `packaging/**`, `.github/workflows/release-distribution.yml`, `.github/workflows/sync-homebrew-tap.yml`) to this workflow's push and pull_request filters, or move the "Test scripts" step into one small dedicated workflow (checkout + pnpm install + `corepack pnpm scripts:test`) that declares that union as its paths.

docs/local-reliability-qa.md
```diff
@@ -20,14 +20,35 @@ final PR, then run product QA on the cumulative stack.
 | Browser flow | `corepack pnpm --filter @jobctrl/web e2e -- tests/<flow>.spec.ts` |
 | Public demo browser workspace | `corepack pnpm --filter @jobctrl/web e2e:demo-workspace` |
 | Public demo edge | `corepack pnpm demo-edge:check`, `corepack pnpm demo-edge:test`, and `corepack pnpm demo-edge:dry-run` |
-| Python worker | `uv --project workers/automation run --extra dev ruff check .` and `uv --project workers/automation run --extra dev pytest -q` |
+| Python worker | `uv --project workers/automation run --locked --all-extras ruff check .` and `uv --project workers/automation run --locked --all-extras pytest -q` |
```
ebarti, line 23: PR Reviewer - pr-reviewer

Severity: Medium
Gate: Fail
Evidence: The development machine's shell profile exports `UV_EXCLUDE_NEWER="8 days"`. With that variable set, every `--locked` command this PR standardizes fails: `uv --project workers/automation run --locked --all-extras pytest -q` prints "Resolving despite existing lockfile due to addition of global exclude newer 2026-08-13T12:30:04Z" then "error: The lockfile at `uv.lock` needs to be updated, but `--locked` was provided". The same failure hits `python:lint`, `python:build`, and both release-check.yml commands when reproduced locally. With the variable unset, `uv --project workers/automation lock --check` passes and all of the above succeed (release_check --strict-prompt passed; self-test 55/55) — the lockfile itself is correct.
Reproduction: Reproduced with command — ran this QA-table command and the release-check.yml command sequence in the PR worktree with and without `UV_EXCLUDE_NEWER` in the environment.

The PR deliberately moves the rolling cooldown out of uv (pyproject comment, `validateUvLockPolicy`, release_check.py) precisely because time-relative cutoffs make an unchanged lockfile fail `--locked` as time passes. A machine-level `UV_EXCLUDE_NEWER` or user-level uv.toml `exclude-newer` is the same contamination through a different door, and it breaks exactly the commands this page documents, with an opaque error. Nothing in the touched docs mentions it.

Required fix: add one line here (or in docs/local-development.md) stating that local environments must not set a global uv exclude-newer (`UV_EXCLUDE_NEWER` env var or user-level uv.toml), otherwise every `--locked` command fails with "lockfile needs to be updated"; and remove the `UV_EXCLUDE_NEWER` export from the dev machine's shell profile when landing this.

scripts/distribution-manifest.mjs
```diff
@@ -711,6 +719,38 @@ export function validateLicenseReview(licenseReview, inventoryById) {
   return reviewedComponents;
 }
 
+const SOURCE_RECORD_COUNT_KEYS = ["pnpmPackageRecords", "uvPackageRecords"];
+
+export function validateSourceDependencyCounts(baselineValue, currentValue) {
```
ebarti, line 724: PR Reviewer - pr-reviewer

Severity: Low
Gate: Pass
Evidence: On main, `auditDistributionContracts` hard-failed when `pnpmPackageRecords`/`uvPackageRecords` drifted from source-baseline.json ("source dependency counts drifted; run distribution:measure and update source-baseline.json intentionally"). This function downgrades those two keys to an advisory `hasRecordDrift`, and nothing consumes the flag: no gate in scripts/*.mjs or in either regenerated authority bundle reads it, so it only travels inside the report JSON. License evidence pins only the enumerated subject packages, so an unlisted transitive package change now passes PR audit and release finalization with no deliberate acknowledgment anywhere.
Reproduction: Reasoned from code — grepped for `hasRecordDrift`/`sourceDependencyReview` consumers; `validatePythonLicenseEvidenceLocks` binds only `EXPECTED_PYTHON_LICENSE_EVIDENCE` subjects.

This is coherent with the Dependabot goal (transitive bumps must not require a baseline edit per PR) and the baseline note was updated honestly. But the two counts still sit in `reproducibleCounts` looking like contracts while going permanently stale after the first transitive bump.

Required fix: make the drift visible instead of silent — emit a warning line in the `audit` CLI output when `hasRecordDrift` is true, or move the two record counts out of `reproducibleCounts` into an explicitly observational baseline field.

workers/automation/pyproject.toml
```diff
@@ -112,8 +116,14 @@ line-length = 120
 
 [tool.pytest.ini_options]
 asyncio_mode = "strict"
+faulthandler_timeout = 180
+timeout = 180
+timeout_method = "signal"
```
ebarti: PR Reviewer - pr-reviewer

Severity: High
Gate: Fail
Evidence: On this PR's head run (https://github.com/ebarti/JobCtrl/actions/runs/32481181792/job/96767685379), the Python 3.11 lane hung and was cancelled by `timeout-minutes: 45` (started 12:17:32Z, cancelled 13:02:48Z) while 3.12 passed in 5m32s and 3.13 in 4m17s. The job log shows exactly why `timeout_method = "signal"` cannot contain this hang class: at 12:30:21Z pytest-timeout fired and dumped stacks — an asyncio executor thread was blocked in `conn.execute("BEGIN IMMEDIATE")` at `workers/automation/src/jobctrl/infrastructure/enrichment/execution_lease.py:68` (via `enrichment/activities.py:363` under a live `temporal-test-server`) — and then the session produced nothing for the remaining 32 minutes. SIGALRM raises only in the main thread; it cannot unwind a worker thread blocked in a SQLite write-lock wait, so teardown wedged behind the stuck thread and the job burned the full 45 minutes before reporting an unattributable "cancelled".
Reproduction: Reproduced with command — polled the run to completion; stack evidence from `gh run view --log --job 96767685379`.

Two parts:
1. The lane itself is red on this head. The deadlock is pre-existing product/test behavior (execution_lease.py last changed in #751; pytest-randomly order-dependent — the same suite passed all three lanes on main on 2026-08-15), so the PR surfaced it rather than caused it, but the PR cannot merge on a cancelled required lane. Rerun the lane; if it reproduces, the enrichment-lease deadlock needs its own issue.
2. In-PR fix: `timeout_method = "signal"` demonstrably fails to terminate thread-blocked hangs, which is the dominant hang class in a suite that owns Temporal `WorkflowEnvironment` threads (this PR's own conftest markers classify files by exactly that). pytest-timeout's `thread` method dumps every stack and then hard-exits the process, converting this 45-minute silent cancellation into a ~3-minute attributable failure.

Required fix: set `timeout_method = "thread"` (or drop the key for pytest-timeout's default) so a wedged test kills the run promptly with the dump attached; `faulthandler_timeout = 180` then also stops double-firing at the same instant as the pytest-timeout dump (stagger it or remove it). Then rerun the Python 3.11 lane and get it green before merge.

.github/workflows/typescript.yml
ebarti: PR Reviewer - pr-reviewer

Verified fixed by a2c1932de: `scripts:test` now runs unconditionally on every pull_request and main push via the new `.github/workflows/repo-scripts.yml` (no paths filter, same SHA pins, read-only token, `persist-credentials: false`), and the duplicate step was removed from the typescript.yml api job. CI on head a2c1932de: "Script unit tests" pass in 37s (run 32486645956). Resolved; no longer gates.

workers/automation/pyproject.toml
ebarti: PR Reviewer - pr-reviewer

Verified fixed by a2c1932de: `timeout_method = "thread"` with `faulthandler_timeout` raised to 240s as the second-chance dump, and the underlying enrichment execution-lease `BEGIN IMMEDIATE` deadlock filed as #836. Additionally 70ae8a27f bounds every WorkflowEnvironment start (30s x 3 fresh-server retries) and shutdown (30s, suppressing only TimeoutError so the test outcome propagates) — I verified the 74 time-skipping and 2 local rewrite sites are textually uniform with zero other edits, and 197/197 temporal tests pass through the helper locally. On head a2c1932de the Python 3.11 lane passes in 3m02s (run 32486645812) versus the prior 45-minute cancellation. Resolved; no longer gates.

docs/local-reliability-qa.md
ebarti: PR Reviewer - pr-reviewer

Repo-side fix verified in a2c1932de: `docs/local-development.md` now warns that `UV_EXCLUDE_NEWER` (or a global uv `exclude-newer` config) makes every `--locked` command fail or rewrite the lockfile. The remaining action — removing the export from the local shell profile — is machine state outside this PR. Resolved for this PR.

scripts/distribution-manifest.mjs
ebarti: PR Reviewer - pr-reviewer

Acknowledged: the a2c1932de commit message records this as an accepted Low (the record-count advisory stays unconsumed by design). No objection; noting for the audit trail.

PR #860, opened 2026-09-03
feat(discovery): use live Chrome profile transport

ebarti:
## Why

Integrated Discovery and LinkedIn Enrich need the authenticated state of the user's currently running Chrome session. Anonymous requests and copied profiles lose that live state, which can surface as robots denials, login challenges, or snapshots of an unhydrated loading shell.

This change makes the paired JobCtrl extension in the user's existing Chrome profile the browser transport for those workflows. JobCtrl neither launches Chrome nor copies the profile, and session changes made by the user are available to subsequent browser tasks without synchronizing a duplicate profile.

## What changed

- add an in-memory browser-task broker between the API/worker and the selected extension installation
- route integrated broad-board, ATS, Workday, Smart Extract, robots, and detail-enrichment acquisition through the extension, with no direct-network or copied-profile fallback
- allow the extension to execute explicitly leased Discovery requests across HTTP(S) sites while keeping capture/autofill API calls loopback-only
- distinguish internal streaming-consumer shutdown from user cancellation so terminal Enrich reclaims unfinished jobs; replay-gated workflow cancellation retains exact-owner cleanup
- wait for LinkedIn's job-detail SPA to hydrate before snapshotting; loading shells fail retryably instead of looking like successful empty pages
- expose selected-installation and heartbeat status in Settings and reject Discover/Enrich dispatch while the extension is offline
- retain bounded concurrency, lease heartbeats, cancellation fencing, hard timeouts, same-origin redirect enforcement, public-destination validation, pacing, request budgets, and audit outcomes

## Safety and privacy

- no application-submission behavior is added or enabled
- browser task bodies and results stay in API process memory
- only the explicitly selected extension installation may lease work
- cookie values are not copied into worker payloads; the browser user agent is returned for robots-policy evaluation
- cross-origin redirects and unsafe destinations fail closed
- cancellation and timeout remove the task and close temporary inactive tabs

## Prior implementation validation

- `env -u UV_EXCLUDE_NEWER corepack pnpm check`
- API browser-broker/routes/server tests: 3 files, 256 tests passed
- web browser-capability/pipeline/settings tests: 4 files, 52 tests passed
- extension unit tests: 7 files, 45 tests passed
- extension production build passed
- packaged-extension Chromium E2E: 2 files, 8 tests passed, including delayed LinkedIn SDUI rendered-page capture
- focused Python Discovery/Enrich transport tests: 94 passed
- post-fix reliability suite: 74 passed; bridge/selectors/cancellation/lifecycle/reconciler suite: 53 passed
- real Temporal/SQLite race, ordering, and explicit cancellation cases: 4 passed using disposable local Temporal servers; the SDK test-server download was unreachable
- current LinkedIn SDUI, never-ready, and throttled-poll regression fixtures passed
- independent code review: no remaining findings
- `corepack pnpm docs:build`
- `git diff --check`

## Live runtime verification

A bounded real-account smoke reached the selected extension and admitted a job. It exposed two failures: source completion falsely canceled unfinished Enrich work, and LinkedIn’s current SDUI detail layout was not recognized by the readiness predicate. Both now have fixes and regression coverage, including a real Temporal/SQLite source-finish-during-capture test that completes and hands off the job.

The final post-fix authenticated Discover + Enrich smoke remains unverified. Fixture/package success is not completed production verification. No new live-provider workflow or application submission is included in this validation.

## Review corrections

- Isolate browser transport and detail-enrichment failures so one failed item does not abort unrelated work, while cancellation remains authoritative.
- Preserve rendered-page hydration when temporary-tab navigation rejects after useful content has loaded; retain bounded readiness checks and temporary-tab cleanup.
- Restrict externally connectable origins to the local application boundary and extend synthetic transport, robots, cancellation, packaged-extension and hydration regressions.
- Clarify that `enrich` is already rejected as a top-level action by the shared request schema; accepted composite workflows still perform Enrich internally. This is documentation and validation of the existing top-level contract.
- Synchronize with current main, retaining its direct API-client port implementation and removed forwarding adapter.

These corrections have focused regression coverage. Synchronized-head synthetic validation is recorded below. The authenticated real-provider Discover + Enrich proof remains unverified; synthetic/package tests do not close that High verification finding.

Review-fix validation at `4ea9c206`: 134 focused Python tests, 258 API tests, 30 web tests, 46 extension unit tests and 10 packaged-extension Chromium E2E tests passed; API/extension typechecks, Ruff, docs build and diff checks passed. Independent review passed 8 guarded Python cases, 12 extension units, the extension build, 3 Chromium hydration/redirect cases and 3 API admission cases. Independent synthetic product QA passed on published head `41b5c9ed`: six actual packaged-extension Chromium probes, nine guarded worker HTTP transport probes, fifteen malformed API requests rejected and a successful real broker lifecycle. All latest applicable CI workflows passed on that head. Code remediation passed review, but the overall review remains FAIL solely because the existing authenticated-provider High is unverified.


workers/automation/src/jobctrl/enrichment/detail.py
```diff
@@ -463,6 +504,55 @@ def scrape_detail_page(page, url: str, *, session: PolitenessSession | None = No
...
+    if not initial_safety.allowed:
+        return _mark_unsafe_url_block(result, initial_safety, blocked_url=url, t0=t0)
+    with session.guard(url) as decision:
+        if not decision.allowed:
+            result["status"] = "blocked"
+            result["politeness_outcome"] = decision.outcome.value
+            result["error"] = decision.reason
+            result["elapsed"] = time.time() - t0
+            return result
+        live_result = browser.rendered_page(url, timeout_seconds=60.0)
```
ebarti, line 564: **High — retryable live-browser failures abort the whole Enrich attempt (and the ATS / SmartExtract source family)**

Invariant: one job's page failure must not end the attempt for other jobs. Base `_scrape_detail_page_body` turned navigation failures and timeouts into `status: error` results handled per job.

Evidence: here `browser.rendered_page()` raises `TransientNetworkError` for every retryable task result (`live_browser.py:318`), including the content script's `navigation_failed` when a LinkedIn detail never hydrates within 12 s, which is what expired or removed postings produce. The per-job `except TransientNetworkError` at L2049 resets the job stage and re-raises, `_run_detail_scraper` releases the cohort and re-raises, and `_ENRICH_RETRY` retries the activity up to 3 times against the same ordering. Non-retryable failures land in the per-job `except Exception` and are isolated, so it is exactly the retryable class that aborts. The same `except (ConfigurationError, TransientNetworkError): raise` shape was added to the ATS loop (`production_wiring.py:649`) and SmartExtract `_run_all` (`smartextract.py:1529, 1559`), which previously isolated per-source errors; `_api_json` maps every 4xx other than 401/403/429, including a 400 `unsafe_discovery_browser_url` for an unresolvable configured host, to `TransientNetworkError` (`live_browser.py:398`).

Required outcome: a retryable browser-task failure marks only its own job or target failed-retryable while the remaining jobs/targets in the same attempt complete; only cancellation and extension unavailability abort the attempt. Regression test at the transport boundary with one job returning a retryable task failure and the others succeeding in the same attempt.

_Reasoned from code; not reproduced (needs a live extension)._

apps/extension/src/discovery-executor.ts
```diff
@@ -0,0 +1,409 @@
...
+    {
+      id: baseId + 1,
+      priority: 1,
+      action: { type: "block" },
+      condition: {
+        regexFilter: "^https?://",
+        resourceTypes,
+        tabIds: [tabId],
+      },
+    },
```
ebarti, line 356: **High — tab-scoped DNR block rule also blocks the rendered page's own cross-origin XHR and fetch**

Invariant: redirect enforcement must not change what the visited page itself can load.

Evidence: this block rule matches `^https?://` for `main_frame` and `xmlhttprequest` in the task tab at priority 1, and the allow rule covers only the exact source origin at priority 10. Chrome's DNR `xmlhttprequest` type covers page `fetch()` and XHR, so every cross-origin request issued by the page's own scripts is blocked. The `jobctrl.discovery.fetch` content-script path this was presumably meant to constrain is never sent by the executor (`content-script.ts:609-612` has a handler and no sender; `http_request` mode fetches from the service worker, `rendered_page` sends only probe and snapshot), so page requests are the only thing the rule affects. `docs/user/security.md:163` says a rendered page's ordinary subresource loading remains owned by Chrome and `docs/developer/security.md:163` says the rules allow "main-frame and Discovery-XHR traffic", which describe the opposite.

Consequence: career and job SPAs that hydrate listings or details from a second origin (hosted ATS API domains, api subdomains) render as empty shells; SmartExtract and enrichment see no data, and with the retryable-failure propagation the resulting failures can abort the attempt.

Required outcome: either rendered-page tasks leave page subresource fetches alone and rely on main-frame enforcement plus the API's final-URL validation, or the docs and threat model state the block and the unsupported page classes. In both cases an E2E fixture page that fetches listings from a second local origin should prove the chosen behavior (the current fixtures serve everything from the source origin).

_Reasoned from code; not reproduced._

apps/extension/src/discovery-executor.ts
```diff
@@ -0,0 +1,409 @@
...
+      lastError = error;
+    }
+    await delay(200, signal);
+  }
+  throw (
+    lastError instanceof Error
+      ? new Error(`Discovery content script did not become ready: ${lastError.message}`)
+      : new Error("Discovery content script did not become ready before its timeout.")
+  );
+}
```
ebarti, line 326: **Medium — a cross-origin main-frame redirect in rendered-page mode surfaces as a retryable timeout**

Invariant: a cross-origin redirect should fail closed quickly and terminally.

Evidence: when the DNR rule blocks the redirected main frame, Chrome shows an error page where no content script runs, so this probe loop runs until the task's abort and the catch at L86-96 reports `navigation_failed` with `retryable: true`. The worker maps that to `TransientNetworkError` and, with the per-job re-raise in `detail.py`, aborts the attempt and Temporal retries it, each retry burning the full task timeout (default 60 s). The redirect E2E case (`chromium-extension.e2e.test.ts:202-227`) uses the default `http_request` mode, so this path is untested.

Required outcome: a blocked navigation is reported promptly as non-retryable `unsafe_redirect`, with a rendered-page redirect E2E case.

_Reasoned from code; not reproduced._

apps/api/src/server.ts
```diff
@@ -990,6 +1240,15 @@ export function buildApp(options: BuildAppOptions): FastifyInstance {
       void reply.code(400);
       return undefined;
     }
+    if (body.stages.includes("discover") && !discoveryBrowserBroker.status().connected) {
```
ebarti, line 1240: **Medium — pipeline-level Enrich dispatch is not gated on the extension heartbeat despite the PR claim**

The PR body says Discover and Enrich dispatch are rejected while the extension is offline, and `docs/local-ts-api.md` documents the 503 only for Discover. This route gates only `discover`: a `run-stage` request with `enrich` among the stages dispatches, and the Pipelines Enrich trigger is not disabled (StageTriggerPanel disables only when `controls.discoverySource` is true). The job-level enrich retry/run routes are gated via `discoveryExtensionUnavailableResponse()`. With the extension offline the worker fails closed after cohort queueing: `scrape_site_batch` calls `ensure_available()`, every site records a `ConfigurationError`, and `_systemic_enrichment_error` returns a non-retryable error, so the user gets a failed run instead of a 503. The route tests (256 passed) cover the discover 503 and contain no enrich-offline case.

Required outcome: the same pre-dispatch 503 and disabled trigger for pipeline-level Enrich, with a route test; or the PR body and docs state this exception explicitly.

_Reasoned from code; not reproduced._

apps/extension/src/content-script.ts
```diff
@@ -553,26 +577,368 @@ function normalize(value: string): string {
...
+    return discoveryFailure("unsupported_page", new Error("The rendered page URL is invalid."), false);
+  }
+  if (!isSafePublicPageUrl(finalUrl)) {
+    return discoveryFailure("unsafe_redirect", new Error("Rendered Discovery page is not a public HTTP(S) URL."), false);
+  }
+  const bodyText = truncateUtf8(
+    document.body?.innerText ?? document.documentElement?.textContent ?? "",
+    4_000_000,
+  );
+  const bodyHtml = truncateUtf8(document.documentElement?.outerHTML ?? "", 4_000_000);
```
ebarti, line 689: **Medium — signed-in page content beyond the posting reaches LLM prompts and persisted descriptions without disclosure**

Privacy boundary check. This snapshot is the entire document's `innerText` and `outerHTML` (up to 4 MB each) as rendered in the user's signed-in session, not just the posting. SmartExtract keeps it as `full_html` and card `examples` (`smartextract.py:568, 635`) and sends cleaned page HTML to the LLM (`smartextract.py:1092-1103`); enrichment persists the extracted description (`detail.py:2169-2199`). Signed-in pages commonly include account-specific regions, for example header identity, candidate-portal greetings, or LinkedIn match and insight panels inside the detail area (an inference about site behavior, not proven from this diff). Before this PR that exposure required the separate copied-profile consent and applied to LinkedIn only; now it is the default for every host where the user is signed in. `docs/user/security.md:57` and `docs/user/data-and-safety.md:207` disclose only that cookie values are not copied.

Required outcome: the data-flow docs state that snapshots are captured in the signed-in session and may include account-personalized page content that is processed, persisted as posting text, and sent to LLM legs; or the worker restricts snapshot content before LLM use and persistence.

_Reasoned from code; not reproduced._

docs/user/data-and-safety.md
```diff
@@ -204,7 +204,7 @@ records something you did. Follow-ups are reminders and never act automatically.
 | Service | When used | Data involved |
 | --- | --- | --- |
 | LLM providers | Scoring, employer analysis, materials, contact extraction, stored interview prep | Posting text, relevant profile evidence, generated text, or opted-in fetched page text. |
-| Job boards, ATS APIs, posting pages | Discovery and enrichment | Search terms, URLs, and page/API requests. |
+| Job boards, ATS APIs, posting pages | Integrated Discovery and its enrichment drain, through the paired extension in the current Chrome profile | Search terms, URLs, and page/API requests; Chrome may also send cookies or other session state that already belongs to that site. Cookie values and the browser user agent are not copied into the worker task. |
```
ebarti: **Low — the claim that the browser user agent is not copied into the worker task is inaccurate**

Every succeeded task result carries `browserUserAgent` (`packages/contracts/src/schemas.ts:4694`; populated by the executor and content script; consumed by `LiveChromeRobotsCache` for robots evaluation), so "the browser user agent [is] not copied into the worker task" is not accurate as readers will understand it. It is not persisted, and `docs/user/security.md:322` already says it is returned for robots evaluation.

Required outcome: say cookies are never copied and the user-agent string is returned with results for robots evaluation only.

_Reasoned from code (schema, producers and consumer read)._

ebarti (commented):
Review of `e444ec6fd262e1991fde439c61fc3692a8057215` against base `ed5e530b4d93eca82365c828c6b2b4b88062c4b8` (independent `pr-reviewer` pass, coordinator-published). Posted as a comment review; the gate below is the review's own assessment, not a GitHub changes-request.

**Result:** Gate FAIL. Blocker 0, High 3, Medium 3, Low 1. Six findings are inline; the third High is unanchorable and stated here.

**High — Tier 3 operational proof incomplete.** `AGENTS.md` Tier 3 requires product and operational proof for a change to the user's browser-session transport and says unavailable evidence is not proof. The PR body states the post-fix authenticated Discovery smoke is still pending an extension reload and that fixture success is not production verification, and `e444ec6fd` changed the LinkedIn readiness predicate and the streaming-consumer cancellation semantics after the only real-account run (one job admitted). Required before merge: a completed authenticated Discover plus Enrich smoke at this or a later head, recorded in the PR with what was observed. (Not run in review: requires the owner's real Chrome profile.)

**Merge state:** GitHub reports CONFLICTING against current `main` (`8eaea25a2`). Merge-tree shows an import-block conflict in `apps/api/test/server.test.ts` (lines 29-34) and a modify/delete on `apps/web/src/shared/adapters/local/FetchApiClientAdapter.ts`, deleted on `main` by #863. Semantic check against #863: the auto-merged `write-model.ts` contains every symbol `retryFailedJobTargets` needs; `db.ts` is unchanged since base; `main`'s `server.ts` change is dead-helper removal only; `main`'s `createLocalPorts` uses `JobCtrlApiClient` directly and this PR already adds `discoveryBrowserBridgeStatus()` to that client, so the adapter hunk can be dropped without loss. The textual conflicts still need resolving.

**Verified in an isolated checkout at head (clean before and after; `uv.lock` untouched):** API broker/routes/server vitest: 3 files / 256 passed (reviewer, plus three further clean coordinator runs; one intermediate reviewer run exited without a test summary during environment provisioning and did not reproduce); extension vitest: 7 files / 45 passed; web vitest for BrowserCapabilitiesPanel, StageTimeline, StageTriggerPanel, SettingsPanel, demo executor/seed: 6 files / 87 passed; Python `test_live_browser_discovery.py` + `test_enrichment_politeness_gate.py`: 26 passed; `test_discover_reliability.py`: 74 passed. Not run here: `test_workflow_discovery.py` (spawns a Temporal test server), the headed extension Chromium E2E (temp profile and loopback fixtures confirmed from source; CI ran it green), typechecks and docs build (CI green at head).

**Investigated and refuted (6):** unauthenticated access to worker task routes (bearer enforced on the `/v1/extension/` prefix, browser-originated requests excluded); semantic conflict with #863; the web adapter gap; broker double-lease or lost results (single-threaded lease assignment, `leaseId` fencing, worker DELETE after read); cancellation resurrecting user-cancelled work (streaming consumer releases rather than terminalizes, patch-gated cleanup terminalizes only the exact owner, covered by new tests); public-destination bypass through the extension (API validates with DNS resolution at create, lease and result).

**Residual risk, not counted:** live JobSpy provider behavior under `credentials: include` with redirects refused is unverifiable offline; `http_request` mode refuses same-origin redirects too, so any API or robots endpoint that 301s fails non-retryably; MV3 service-worker lifetime during long `http_request` fetches is unverified and lease expiry would mask it as retries; the content script now runs passively on every HTTP(S) page (message-driven only, but permanently broader reach).

ebarti (commented):
Review of `41b5c9ede0e835fe9d1d7e449fda9a85dc454967` against merge-base `8eaea25a2fd9034798ef5bac18f2994a33a713c5` (current `main`; independent `pr-reviewer` pass, coordinator-published). The prior review's head was rebased away, so every thread was re-verified from scratch at this head. Posted as a comment review; the gate below is the review's own assessment, not a GitHub changes-request.

**Result:** Gate FAIL. Blocker 0, High 1, Medium 1, Low 3. Four findings are inline (status-less rendered pages, executor ignoring non-blocked navigation errors, "Discovery-fetch" doc drift, bulk-retry preview outside the DB guard). The High is the existing operational-proof finding and is stated here. All six prior inline threads are verified fixed at this head (details below).

**High (existing) — Tier 3 operational proof still incomplete.** The PR body at this head says the authenticated real-provider Discover + Enrich proof remains unverified, and the single real-account run (one job admitted) predates the readiness predicate, streaming-consumer cancellation, per-item failure isolation and DNR changes in the second and third commits. Under `AGENTS.md` Tier 3 that proof is required and unavailable evidence is not proof, so this is not satisfied at head. Required before merge: a completed authenticated Discover plus Enrich smoke at this or a later head, recorded in the PR with what was observed (job admitted, LinkedIn detail enriched through the extension, bridge task counts in Settings). (Not run in review: requires the owner's real Chrome profile.)

**Verified in an isolated checkout at head (clean before and after, `uv.lock` untouched):** API vitest broker + routes: 2 files / 17 passed. Extension vitest (discovery-executor, discovery-loop, content-script, local-api, privacy, popup): 6 files / 42 passed. Python `test_live_browser_discovery.py` + `test_enrichment_politeness_gate.py` + `test_discovery_production_wiring.py`: 58 passed; `test_discover_reliability.py`: 74 passed. One in-memory Python probe reproduced the Medium. Merge-tree against `main` is conflict-free. Not run here: `test_workflow_discovery.py` (SDK time-skipping test server, not the local temporal CLI), the packaged-extension Chromium and privacy E2E, web vitest/E2E/Storybook, typecheck and docs build (all green in CI at head).

**Existing-thread verdicts (verified at head, not resolved by this review):**
- discussion_r3945260458 (High, retryable failures abort the attempt): **fixed at head.** `live_browser.py:327-333` raises `LiveBrowserTaskError` (a `JobCtrlError`, not `TransientNetworkError`) for failed task results and `live_browser.py:412-413` maps create-time 400s to a non-retryable task error; `detail.py:2059-2072` isolates it per job with the task's code/retryability; `production_wiring.py:649-652` and `smartextract.py:1530/1560` re-raise only `ConfigurationError`/`TransientNetworkError`. Executed and passed: `test_live_browser_task_failure_isolated_from_remaining_enrich_jobs`, `test_live_browser_smartextract_failure_preserves_remaining_targets`, `test_live_browser_ats_failure_preserves_other_sources`, `test_unsafe_task_admission_is_a_target_failure`, `test_extension_disconnect_still_aborts_enrich_before_the_next_job`.
- discussion_r3945260463 (High, DNR blocks page XHR): **fixed at head.** `discovery-executor.ts:329-330` scopes both rules to `main_frame`; the unit test asserts `resourceTypes: ["main_frame"]`; the E2E "hydrates a rendered posting using the page's own fetch to a second origin" exists (CI, not run here); user/developer security pages updated. Residual doc drift in four other files is a new Low.
- discussion_r3945260469 (Medium, rendered-page redirect surfaces as timeout): **fixed at head.** `discovery-executor.ts:66-77, 100-102` adds a task-tab `webNavigation.onErrorOccurred` listener that aborts on `net::ERR_BLOCKED_BY_CLIENT` for a foreign origin and returns non-retryable `unsafe_redirect`; unit test "reports a blocked task navigation promptly" passed; the `it.each(["http_request","rendered_page"])` redirect E2E case exists (CI).
- discussion_r3945260471 (Medium, Enrich dispatch not gated): **fixed at head.** `PIPELINE_RUN_STAGES` (`schemas.ts:6`, unchanged from base) has no `enrich`, so `run-stage` cannot carry it; `server.test.ts` now covers `["enrich"]`, `["discover","enrich"]`, `["score","enrich"]` with 400 and no dispatch. Job-level run/retry (`server.ts:2025, 2070`), bulk retry with `runAfter` (`server.ts:1668-1681`) and bulk pending preparation (`server.ts:1752`) preflight the heartbeat; `docs/local-ts-api.md:166-169` documents the exception; StageTriggerPanel disables Discover while offline.
- discussion_r3945260473 (Medium, signed-in snapshot disclosure): **fixed at head.** `docs/user/data-and-safety.md` (external-services table and the new signed-in snapshot paragraph) and `docs/user/security.md` state that snapshots come from the signed-in session, may include account-personalized content, can be persisted as posting text and sent to LLM providers; README links the notice.
- discussion_r3945260480 (Low, user-agent claim): **fixed at head.** `docs/user/data-and-safety.md` now says cookie values are never copied and the browser user-agent string is returned with results for robots evaluation only.

**Investigated and refuted (9):** bearer enforcement on the worker-task routes (the `/v1/extension/` prefix rule at `server.ts:3894-3900` and the onRequest hook cover safe and unsafe methods; worker-only routes additionally require no Origin/Referer/Sec-Fetch headers; route test "does not let an extension-origin request create worker tasks" passed); broker double-lease or lost result (single-threaded lease assignment, `leaseId` fencing with 409 on stale completion, expired-lease requeue test passed, worker DELETE after read); cancellation resurrecting user-cancelled work (`_ActivityCancellationEvent(terminal_on_cancel=not stream_while_discovering)` makes the streaming consumer release rather than terminalize, and the patch-gated `_cancel_owned_enrichment` terminalizes only rows whose metadata matches the exact workflowId/temporalRunId at `activities.py:106-116`); public-destination validation at create, lease and result (create and result call `jobUrlValidator`, `/tasks/next` re-validates DNS then re-checks selection and lease, broker enforces same origin; route test "revalidates DNS at lease time" passed); externally connectable exposure (no `externally_connectable` key exists at head and the content script has no window message listener); JobSpy registry kwarg dropped by the signature filter (`run_discovery` accepts `adapter_registry` at `jobspy.py:2139`; `runner.py:1519` sets it directly); JobSpy silently succeeding while the extension is offline (adapter failures surface as typed error events that `jobspy.py:1969-2025` records as unit failures or `DiscoveryResumeRequired`); removal of the top-level-enrich 400 test (expanded to three stage combinations, not removed); content-script discovery handlers reachable from pages (`chrome.runtime.onMessage` is extension-only).

**Residual risk, not counted:** live JobSpy providers under `credentials: include` with all redirects refused are unverifiable offline, and LinkedIn's guest API commonly 302s, which now yields non-retryable `unsafe_redirect` per request; more than four concurrent requesters wait at most 30 s for admission and then abort their attempt with `TransientNetworkError`; MV3 service-worker death mid-task becomes a lease expiry followed by a worker 404 or queue-deadline abort and a Temporal retry; the pairing token now grants a local process an authenticated cross-site request primitive in the user's browser (accepted by the ADR; the developer security page names token disclosure as the main risk); the popup's `replace:false` status claim awaits the blocked-preparation continuation dispatch and can time out into "Discovery not selected" for a selected profile; the legacy Playwright/copied-profile enrichment branch and the unused content-script fetch handler remain as dormant code.

apps/extension/src/content-script.ts
```diff
@@ -553,26 +577,368 @@ function normalize(value: string): string {
...
+  if (!bodyText.trim() && !bodyHtml.trim()) {
+    return discoveryFailure("unsupported_page", new Error("The rendered page exposed no readable content."), true);
+  }
+  return {
+    status: "succeeded",
+    finalUrl: finalUrl.href,
+    statusCode: null,
```
ebarti: **Medium — rendered-page snapshots never carry the HTTP status, so removed postings are classified active and retryable**

Invariant: the enrichment audit and active-state verification must keep the same evidence they had under the Playwright path, and a removed posting must terminalize rather than retry.

Evidence: this line hard-codes `statusCode: null` for every rendered page. `live_browser.py:338` maps it to `status_code=None`, `detail.py` `_live_result_to_detail_page` passes `status=None`, and `snapshot_services.py:150-153` only returns `REMOVED` for 404/410 when the status is present; `detail.py` `_detail_failure_retryable` only returns non-retryable for 4xx when the status is an int. The base path `_page_to_detail_page(page, url, status)` used the navigation status. Probe on one identical "Job not found" snapshot: status 404 gives `removed / http_status / retryable=False`; status None gives `active / default_body_present / retryable=True`.

Consequence: every live enrichment records `http_status: null`, and a 404/410 ATS posting page without textual closed/removed markers is recorded as an active posting whose extraction failure is retryable, so it is re-fetched through the bridge on each retry instead of terminalizing as removed. Chrome exposes the top-level status as `PerformanceNavigationTiming.responseStatus` (109+), so the value is obtainable here.

Required outcome: rendered-page results report the navigation HTTP status (or the audit records it as unavailable without the verifier and retry classification silently degrading), with a regression test that a 404 rendered page classifies removed and non-retryable.

_Reproduced (in-memory Python probe against head code)._

apps/extension/src/discovery-executor.ts
```diff
@@ -0,0 +1,425 @@
...
+      browser.tabs.create({ active: false, url: "about:blank" }),
+      controller.signal,
+    );
+    if (!tab.id) {
+      return failed("navigation_failed", "Chrome did not create a Discovery tab.", true);
+    }
+    tabId = tab.id;
+    navigationListener = (details) => {
+      if (details.tabId !== tabId || details.frameId !== 0 || controller.signal.aborted) return;
+      if (details.error !== "net::ERR_BLOCKED_BY_CLIENT") return;
```
ebarti: **Low — non-blocked top-frame navigation errors are ignored, so unreachable hosts consume the full task timeout**

Invariant: a task whose navigation has already failed should not hold an executor slot until the hard timeout.

Evidence: the listener returns for every error except `net::ERR_BLOCKED_BY_CLIENT`. On `ERR_CONNECTION_REFUSED`, `ERR_CONNECTION_TIMED_OUT`, `ERR_CERT_*` or `ERR_NAME_NOT_RESOLVED` the tab shows a Chrome error page where no content script runs, so `waitForContentScript` (L293-321) polls every 200 ms until the task timeout (60 s default, 120 s max) and the catch at L103-112 reports `navigation_failed` retryable. The lease-time DNS check does not cover refused or timed-out connections.

Consequence: one of the four executor slots and the worker thread are held for the full timeout per attempt for each dead host.

Required outcome: a top-frame network error on the task tab fails promptly as retryable `navigation_failed` (`ERR_ABORTED` excluded), covered by a unit case.

_Reasoned from code; not reproduced (needs a headed extension)._

docs/user/discovery.md
```diff
@@ -141,6 +141,60 @@ mode, then keep the operations workspace open while work proceeds. The source
...
+hard task timeout aborts the request, closes a temporary tab when one exists,
+and prevents a stale result from being accepted. Response bodies are not added
+to `config.json` or a bridge queue on disk.
+
+The API validates a destination when it is queued and again immediately before
+lease. HTTP/API execution disables redirect following, which is required
+because vendor API roots can be JSON/plain text or redirect away and therefore
+cannot reliably host a content script. Before rendered-page navigation, the
+extension installs a tab-scoped rule pair that allows the task's exact origin
+and blocks all other HTTP(S) main-frame and Discovery-fetch destinations.
```
ebarti: **Low — this and three other docs still say the DNR rule pair blocks "Discovery-fetch"/"fetch" redirects**

Invariant: security-control descriptions must match the shipped rules.

Evidence: at head `redirectGuardRules` covers `main_frame` only (`discovery-executor.ts:329-330`) and the content-script `jobctrl.discovery.fetch` handler is never sent by the executor. The same overstatement appears at `docs/decisions.md:2372`, `docs/architecture/pipeline/stages.md:126` and `README.md:413`, while `docs/user/security.md:167`, `docs/local-ts-api.md:158` and `README.md:578` correctly say page-owned fetch/XHR is not matched.

Consequence: the ADR and this user page claim a control the security pages deny.

Required outcome: align the four statements to main-frame-only blocking and either drop the "Discovery-fetch" notion or the dead handler it refers to.

_Reasoned from code and docs._

apps/api/src/server.ts
```diff
@@ -1407,6 +1666,19 @@ export function buildApp(options: BuildAppOptions): FastifyInstance {
       return undefined;
     }
     if (body.runAfter) {
+      const previewDb = openDatabase(options.dbPath);
```
ebarti: **Low — the bulk-retry enrich preview opens the database outside the DB guard, turning a missing DB into a 500**

Invariant: routes return the documented 503 when the database does not exist.

Evidence: `openDatabase` uses `fileMustExist: true` (`db.ts:64`) and throws here before `requireWorkerReady` and `withWritableDb` run; `withWritableDb` (`server.ts:4611-4613`) is what returns `503 db_not_found`. Base ordering reached one of those 503 paths first.

Consequence: `POST /v1/jobs/bulk-retry-failed` with `runAfter: true` against an uninitialized app dir returns an unhandled 500 carrying the raw SQLite message.

Required outcome: the enrich preview runs inside the existing missing-DB handling so the route keeps returning 503.

_Reasoned from code._

workers/automation/src/jobctrl/enrichment/detail.py
ebarti: Verified addressed at published 69c56b4f40bbe0d205c50ece7c80cfbd79fce0ec. Failed task results raise LiveBrowserTaskError with the target's code/retryability; Enrich handles it per job and ATS/SmartExtract propagate only extension unavailability/cancellation classes. The regression matrix exercised per-job failure isolation, remaining ATS/SmartExtract sources, unsafe task admission, and disconnect propagation: 134 focused Python tests passed, independently repeated. Resolving the already-present fix.

apps/extension/src/discovery-executor.ts
ebarti: Verified addressed at published 69c56b4f40bbe0d205c50ece7c80cfbd79fce0ec: both DNR rules match main_frame only. The packaged Chromium test hydrates a posting through the page's own second-origin fetch and passed in both implementation and independent QA runs. Security summaries now consistently describe the main-frame-only rule. Resolving the already-present behavior fix and completed documentation alignment.

apps/extension/src/discovery-executor.ts
ebarti: Verified addressed at published 69c56b4f40bbe0d205c50ece7c80cfbd79fce0ec. The selected-tab top-frame ERR_BLOCKED_BY_CLIENT event for a foreign origin aborts promptly as non-retryable unsafe_redirect. Packaged Chromium tests passed for both HTTP and rendered-page public-to-loopback redirects and independently confirmed that the target was never requested. Resolving the already-present fix.

apps/api/src/server.ts
ebarti: WON'T DO for adding a pipeline-level Enrich trigger: that surface is not supported by the shared PIPELINE_RUN_STAGES request contract; enrich is rejected as a top-level run-stage value before dispatch. Current server regressions reject ['enrich'], ['discover','enrich'], and ['score','enrich'] with 400 and no dispatch. Supported job-level and bulk Enrich operations have the heartbeat preflight, with tests proving offline rejection preserves stage state and diagnostics. The PR body and API docs already clarify this contract. These admission checks passed on published 69c56b4f40bbe0d205c50ece7c80cfbd79fce0ec; resolving this unsupported-surface finding.

apps/extension/src/content-script.ts
ebarti: Verified the existing disclosure at published 69c56b4f40bbe0d205c50ece7c80cfbd79fce0ec. docs/user/data-and-safety.md explicitly states that signed-in snapshots may contain personalized content beyond the posting, that extracted posting text can be persisted, and that HTML/text may reach configured LLM providers; docs/user/security.md and README direct users to that boundary. The snapshot data flow was not suppressed. Docs build and link checks passed; resolving the addressed disclosure request.

docs/user/data-and-safety.md
ebarti: Verified the corrected external-services disclosure at published 69c56b4f40bbe0d205c50ece7c80cfbd79fce0ec: cookie values are never copied into task/results, while the browser user-agent string is returned for robots evaluation. This agrees with the result schema and consumer. Docs build passed; resolving the already-present correction.

apps/extension/src/content-script.ts
ebarti: Fixed and pushed in 69c56b4f40bbe0d205c50ece7c80cfbd79fce0ec. The content-script snapshot now reads navigation responseStatus and preserves valid HTTP status codes, retaining null only when unavailable. Before the fix, 200/404/410 producer tests reproduced null. Afterward, actual packaged Chromium 404 and 410 navigations returned those statuses; worker bridge tests carried them into removed/http_status with non-retryable classification. Both implementation and independent QA passed. Resolving.

apps/extension/src/discovery-executor.ts
ebarti: Fixed and pushed in 69c56b4f40bbe0d205c50ece7c80cfbd79fce0ec. Selected-tab top-frame navigation errors now abort promptly as retryable navigation_failed, preserving the browser error; ERR_ABORTED is excluded, unrelated tabs/subframes are ignored, and foreign-origin DNR errors remain unsafe_redirect/non-retryable. Four network/certificate regression cases timed out before the fix and now pass with cleanup assertions. Independent QA passed. Resolving.

docs/user/discovery.md
ebarti: Fixed and pushed in 69c56b4f40bbe0d205c50ece7c80cfbd79fce0ec. README, Discovery guide, ADR, and pipeline stage description now consistently say DNR blocks cross-origin main-frame redirects and leaves page-owned fetch/XHR under Chrome's normal policy. This matches the shipped main_frame-only rules and passing second-origin hydration E2E. Docs build validated 9657 references across 360 emitted files. Resolving.

apps/api/src/server.ts
ebarti: Fixed and pushed in 69c56b4f40bbe0d205c50ece7c80cfbd79fce0ec. The bulk-retry preview and readiness checks now run inside withWritableDb, before any retry-state mutation. The new uninitialized-app regression reproduced 500 SQLITE_CANTOPEN before the change and now returns 503 db_not_found, with no DB created or action dispatched. Existing offline Enrich tests still preserve diagnostics and attempt state. API checks and independent QA passed. Resolving.

ebarti:
PR Fixer - pr860

Published `69c56b4f40bbe0d205c50ece7c80cfbd79fce0ec` on the existing PR branch. Rendered snapshots preserve navigation HTTP status; failed top-frame network navigation releases its task promptly; bulk retry uses the existing missing-database guard; four DNR descriptions now match the main-frame-only rules.

All 10 inline threads are closed: four new fixes, five verified prior fixes/disclosures, and one unsupported pipeline-level Enrich finding explained in-thread. Top-level `enrich` remains rejected by the shared schema; supported Enrich entry points retain heartbeat preflight.

Validation:
- `corepack pnpm extension:test`: 55 passed.
- `corepack pnpm extension:check` and `corepack pnpm api:check`: passed.
- Focused API broker/routes/server suites: 259 passed.
- Guarded locked UV run of live-browser, enrichment-politeness, production-wiring and discovery-reliability suites: 134 passed.
- `corepack pnpm extension:e2e`: production build and 12 packaged Chromium cases passed, including real 404/410 navigation status, cross-origin page fetch hydration and redirect refusal.
- Ruff, `git diff --check`, and `corepack pnpm docs:build`: passed; 9657 documentation references checked across 360 files.
- Independent source review and synthetic QA: PASS; repeated 55 extension units, 40 focused API cases, 134 Python tests and six packaged Chromium cases.

The separate High operational-proof requirement remains open. An isolated API/Temporal/worker/web runtime is healthy and the existing owner Chrome extension successfully executed a public-page transport probe. That installed extension still returns no navigation status, so the current build must be loaded/reloaded before authenticated real-provider Discover + Enrich proof can be recorded. No live model call, application submission or owner-database mutation was performed. Synthetic results do not close that remaining High.

ebarti:
Reconciled this PR with main `e3750ca001dc65c32b5e243f439711952927a817` in `6b1e4fa306f2476eaede5e0ecf07598755b5aacc`, preserving both the live Chrome acquisition fixes and the merged recovery changes.

The integration fixes retain the shared cancellation signal, remove a duplicated assignment, update two calls to the typed public-fetch failure helper, and adapt the Temporal recovery fixture to the live Chrome capture boundary. Both new URL-denial regressions reproduced the stale-helper `NameError` before the fix. The fixture continues to exercise real workflow execution, persistence and ownership fencing.

Focused verification passed: 263 API tests plus types; 58 web tests plus types; 72 Python URL-safety, live-browser and politeness tests; six previously failing Temporal cases after the fixture correction; changed-file Ruff; docs build; and diff checks. The initial combined Python run's six fixture failures are retained in the validation record.

Independent review and reconciliation QA passed at the exact head: 12 focused cases covered both URL-denial phases, four cancellation paths, and six real Temporal recovery/ownership cases. The extension production build passed. The branch is pushed, and a fresh GitHub check confirms it can merge with current main.

Live authenticated Discover and Enrich verification remains pending after extension reload. This reconciliation does not claim that operational proof, and the PR remains open for that check.

ebarti:
The six Python CI failures came from incomplete isolation in the Temporal recovery fixture. Page capture and robots checks were synthetic, but the real browser-readiness preflight still contacted a local API. A running API masked this in the earlier focused validation; the same tests failed on GitHub where that service was absent.

Commit `dfb215cb6a284ea210ff1560990536197ef47663` gives the real browser client a strict synthetic status transport and an unreachable API URL. A teardown guard fails the test if it attempts to use the default network transport. Existing workflow, persistence, cancellation and ownership assertions remain in place.

The failure was reproduced on unchanged `6b1e4fa` by setting `JOBCTRL_API_PORT=1`. After the fixture correction, all 37 affected tests passed with the same unavailable API. This fixes a test isolation defect and does not complete the separate authenticated extension Discover/Enrich check.

The complete Python suite passed at the exact committed head: 3,956 passed, two existing skips, in 196.12 seconds. Ruff and diff checks passed. Independent review and scoped QA passed, including an unavailable-API Temporal case and five direct fixture controls for real readiness/parsing, disconnected/malformed responses, default network rejection and teardown detection of caught network attempts.

[Published-head Python CI](https://github.com/ebarti/JobCtrl/actions/runs/34277405776) passed on Python 3.11, 3.12 and 3.13 at `dfb215cb6a284ea210ff1560990536197ef47663`.

ebarti:
The live authenticated check exposed two remaining LinkedIn acquisition gaps, fixed in `a51f56b35b8010a2f7a5f80a65bff4a92b6dbf19`:

- Cold signed-in pages could exceed the 12-second content-readiness window. The extension now allows 30 seconds while retaining the executor's separate hard timeout, inactive tabs, cleanup, and navigation guards.
- The worker's deterministic extractor did not recognize the description container already recognized by extension readiness. It now extracts `JobDetails_AboutTheJob_*` after snapshot cleaning and excludes adjacent company content.

Both focused regressions failed before the fix. Validation passed: extension types and all 56 unit tests; all 12 packaged Chromium/privacy cases, including an actual 15-second delayed inactive-tab page; 66 Python transport, extraction and URL-safety tests; changed-file Ruff; docs build and 9,659 link references; `git diff --check`.

After reloading, the real Temporal Enrich workflow persisted a 5,918-character LinkedIn description with extraction tier `css_selectors`. The other target reached a durable retryable `navigation_failed` state; its timeout did not prevent the healthy target from completing. Prior attempts remain in both histories. A fresh Discover workflow completed without duplicate rows. Independent SQLite inspection confirmed two jobs, one enriched description, one retryable failure, and zero score rows. No model calls or application submissions ran; downstream preparation and compensation were disabled to isolate acquisition. Settings stayed connected and returned from one active task to zero waiting/active tasks.

Merged as `a1526649a395b9febb4c20ef30dd946945014a21`. A separate fetch and comparison confirmed main has exactly the tested head's tree (`256058133d800201ecff144ec6797a30b3577ca4`). All ten inline review threads are resolved. Changed-surface hosted CI, all three Python versions, and the other completed PR checks passed; the unchanged web E2E job was still running at the final snapshot. Post-merge CI is running. The unrelated local lockfile and canonical AGENTS.md edits were preserved.

PR #861, opened 2026-09-04
chore(dev): add guarded job-data purge tool

ebarti:
## Why

Development and QA sometimes need to reset a JobCtrl workspace's job-derived state while keeping the candidate profile, search configuration, source settings, templates, and policies intact. Ad hoc SQL or filesystem cleanup is difficult to reproduce and can leave workflow/projection history behind, making the next test run inherit stale state.

This PR adds a source-checkout maintenance tool for that internal workflow. It inventories first, requires an exact destructive confirmation, creates a recovery bundle, and verifies the preservation boundary after the transaction. It is intentionally separate from the user-facing Discovery/browser work and is not presented as a product feature.

## What changed

- add `corepack pnpm data:purge-jobs` as a read-only inventory command
- require `--confirm DELETE-ALL-JOB-DATA` before mutation
- refuse active/revivable work, non-local tenants, unexpected schemas, symlinked authorities, and artifact paths outside JobCtrl-owned roots
- back up the database and generated material before removing the Job aggregate, generated materials, and the job/Discovery execution ledger
- preserve and fingerprint candidate-profile data, search/source settings, templates, policies, configuration, contacts/outreach, and unrelated operational history
- document the command only in developer/architecture/QA surfaces

## Prior implementation validation

- `corepack pnpm api:check`
- focused destructive-boundary tests: 2 files, 11 tests passed
- complete API suite: 59 files, 825 tests passed
- `corepack pnpm docs:build`
- `git diff --check`

All destructive-path tests use disposable temporary workspaces. This validation did not run the command against a real user workspace.

## Review corrections and current validation

- Reject fresh worker heartbeats for the target database and detect concurrent database commits between inventory, backup and the immediate deletion transaction.
- Inventory retained reference rows and provisional missing-history executions; document the required operator reconciliation instead of suggesting cancellation can clear an absent execution.
- Add disposable-workspace regressions for concurrent writes, worker activity, retained references and recovery boundaries.
- Independent review and product QA passed at `8ca6adb481b566953ddfa997eb938ebd054ab51a`. No real user workspace was purged. This PR remains separately based on main.

Review-fix validation: 17 focused destructive-boundary tests and the full 831-test API suite passed, with API typecheck and docs build. Independent review passed 16 purge cases and 8 exact-documentation SQL probes. Independent product QA passed five actual CLI scenarios covering backed-up deleted rows, archived file hashes, eight retained-reference families, exact owner checks, fresh worker heartbeats and separate database writers before/after backup. All data was synthetic and disposable.


apps/api/src/job-data-purge.ts
```diff
@@ -0,0 +1,869 @@
...
+    activeWorkflowCount: countRows(
+      db,
+      `SELECT COUNT(*) AS count
+         FROM workflow_run_projections
+        WHERE tenant_id = ?
+          AND (
+            LOWER(status) NOT IN (${workflowPlaceholders})
+            OR (
+              LOWER(status) = 'terminated'
+              AND LOWER(COALESCE(error_code, '')) = 'reconciled_not_found'
```
ebarti, line 574: **Medium — provisional `reconciled_not_found` terminals make the purge refuse permanently with no operator path**

Invariant: the active-work guard should block only work that can still write this workspace, and every refusal needs an operator path.

This predicate treats `terminated` + `reconciled_not_found` rows as active. The worker's reconciler stamps that code after Temporal dev-server history loss and clears it only if the exact run reappears (`_reconcile_one_workflow_run`), so on such a workspace the command refuses forever with "Stop or cancel it first" although nothing is running and nothing can be canceled. Neither new docs section mentions the state, and the PR's own test asserts the refusal.

Required outcome: a documented or built-in resolution for provisional terminals (list the offending workflow ids in the inventory and state how to clear them once the Temporal namespace is verified empty, or an explicit verified override), with a test on the operator-visible path.

_Reasoned from code; refusal behavior confirmed by the PR test in CI._

apps/api/src/job-data-purge.ts
```diff
@@ -0,0 +1,869 @@
...
+    }
+
+    assertNoActiveWork(db);
+    const registeredFiles = classifyRegisteredFiles(appDir, registeredArtifactPaths(db));
+    const preservedBefore = preservationSnapshot(db, appDir);
+    assertWorkspaceAuthorities(authorities);
+    backupDirectory = createBackupDirectory(appDir);
+    authorities = assertWorkspaceAuthorities(authorities, { allowBackupCreation: true });
+    databaseBackupPath = path.join(backupDirectory, "jobctrl-before.db");
+    backupDatabase(db, databaseBackupPath);
```
ebarti, line 861: **Medium — backup-to-delete window is unguarded and a live JobCtrl runtime is never detected**

Invariant: everything the confirmed run deletes must be in the recovery bundle, and the run must refuse while another JobCtrl process can write this database.

`VACUUM INTO` completes here, outside the `BEGIN IMMEDIATE` opened by `transaction.immediate()` below. A concurrent API/worker commit in that window (new job, edit, artifact) is deleted by the transaction but absent from `jobctrl-before.db`. The in-transaction re-checks cover only active-work rows and the preserved tables. `worker_runtime_heartbeats` (fresh within `WORKER_RUNTIME_STALE_AFTER_MS`, keyed by `db_path`) is never consulted, so the documented "stop every JobCtrl process" precondition in `docs/local-development.md` is unenforced.

Required outcome: the confirmed run aborts without deleting when the database changed after the backup (for example `PRAGMA data_version`, or purged-table counts / max `event_id` captured after the backup and re-verified inside the transaction), and refuses on a fresh heartbeat for this `db_path`; regression test that commits a write between backup and delete.

_Reasoned from code._

docs/architecture/storage.md
```diff
@@ -87,6 +87,19 @@ licensed-feed coverage policy. Public Levels.fyi Markdown needs no credential.
 Credentials, feed paths/URLs, feed contents, and provider payloads do not belong
 in the settings file.
 
+The guarded job-data purge uses aggregate ownership rather than table age. The
```
ebarti, line 90: **Medium — documented boundary and no-op verdict omit rows that keep purged job/run references**

Invariant: the documented boundary must name every row that survives with a purged job/run reference, and the "nothing remained" verdict must not hide them.

Rows outside both the deletion and preservation lists (`JOB_OPERATION_TABLES` and the preserved-table set in `apps/api/src/job-data-purge.ts`): pending `manual_capture_queue` items (job_id NULL so no cascade; seeded by Discovery via `_enqueue_manual_capture`, shown by `listManualCaptureQueue`), `source_locator_candidates`, the append-only learning ledger (`learning_recommendation_jobs`, `learning_recommendation_evidence_jobs`, `tailoring_feedback_signal_reviews` whose source `tailoring_feedback_signals` rows cascade away), and `role_match_feedback_suggestions.evidence_json`. After a confirmed run the Discovery page still lists captures "required" for purged runs and learning evidence still cites purged job ids, while the CLI reports nothing remained.

Required outcome: either purge the Discovery-seeded pending captures and count them in the inventory, or document these rows as retained and have the inventory report them. The learning ledger is trigger-immutable, so documentation plus inventory disclosure is the observable outcome there.

_Reasoned from code (schema_v7.sql plus the purge lists)._

ebarti (commented):
Review of `cb02f4787279442186cac3a34e73c2f06e3dbf08` against base `ed5e530b4d93eca82365c828c6b2b4b88062c4b8` (independent `pr-reviewer` pass, coordinator-published).

**Result:** Gate PASS. No Blocker/High. Three Medium findings posted inline (active-work guard has no operator path for provisional terminals; unguarded backup-to-delete window and no live-runtime detection; documented boundary omits rows that retain purged job/run references).

**Verified:** `corepack pnpm api:check` and `git diff --check` pass in an isolated checkout at head; merge-tree against current `main` (`8eaea25a2`) is conflict-free; no schema/manifest/migration change landed on `main` since this base, and the `permanentlyDeleteJobs`/`purgeJobRows` boundary the tool relies on is unchanged. CLI-only reachability confirmed (no HTTP/JSON-RPC route, no production import). Focused suites `test/job-data-purge.test.ts` + `test/permanent-delete-v7.test.ts`: 2 files / 11 tests passed in the isolated checkout at head (temp workspaces only), matching CI run 33897004549.

**Limits:** no run against a populated real-shaped workspace (the author did not either); the CLI path is proven only via `spawnSync` against seeded temp workspaces. Residual note, not a finding: `JobDataPurgeCommittedError` instructs a restore even when only post-commit compaction/verification failed after a clean commit.

apps/api/src/job-data-purge.ts
```diff
@@ -0,0 +1,939 @@
...
+        backupDirectory,
+        databaseBackupPath,
+        jobOperationRowsDeleted,
+        jobsDeleted,
+        logicalDatabaseBytesAfter: fs.statSync(databasePath).size,
+        movedGeneratedFiles: plan.generatedFiles + plan.registeredLogFileCount,
+        noOp: false,
+      };
+    } catch (error) {
+      throw new JobDataPurgeCommittedError(error, backupDirectory, databaseBackupPath);
```
ebarti, line 934: **Low — post-commit compaction failure is reported as a restore-and-retry condition**

Invariant: recovery guidance after a committed purge must match what actually failed.

Evidence: `wal_checkpoint`, `VACUUM` and the post-commit verification asserts share one `catch` (L907-935) that throws `JobDataPurgeCommittedError`, whose message (L203-207) instructs restoring `jobctrl-before.db` plus the `files/` tree and retrying. `VACUUM` needs free space up to the database size, immediately after `VACUUM INTO` consumed the same amount under `backups/`, so `SQLITE_FULL` there is plausible on exactly the large workspaces this command targets.

Consequence: the operator restores a correctly committed and verified purge, retries, hits the same compaction failure, and accumulates one more full database copy per attempt.

Required outcome: report compaction failure separately (purge committed; bundle path; rerun compaction or free space; no restore) from post-commit verification failure (restore), with a regression fixture that fails `VACUUM` after commit.

_Reasoned from code; the existing post-commit test injects a `statSync` failure only._

docs/architecture/storage.md
```diff
@@ -87,6 +87,36 @@ licensed-feed coverage policy. Public Levels.fyi Markdown needs no credential.
...
+on a no-op. These are outside the deletion boundary and can still cite removed
+jobs, runs, or feedback signals:
+
+| Retained rows | Remaining references or effect |
+| --- | --- |
+| `manual_capture_queue` with `job_id IS NULL` | Includes pending Discovery captures and their `retry_context_json`; they can still appear as required on the Discovery page. Job-linked captures cascade with their Job. |
+| `source_locator_candidates` | Retains discovered source URLs and manual-action reasons. |
+| `learning_recommendation_jobs`, `learning_recommendation_evidence_jobs` | Immutable learning provenance retains historical JobIds. |
+| `learning_recommendation_evidence`, `tailoring_feedback_signal_reviews`, `tailoring_feedback_signal_contradictions` | Append-only evidence/reviews retain signal/review references and historical JobIds even after source signals cascade away. |
+| `role_match_feedback_suggestions` | Retains `evidence_json`, which may cite purged jobs or runs. |
```
ebarti, line 113: **Low — retained-reference disclosure omits two surviving reference families**

Invariant (thread discussion_r3945227655): the documented boundary and the inventory must name every row that survives with a purged job/run reference.

Evidence: `workflow_run_projections` rows of non-job types survive (type not in `JOB_DATA_WORKFLOW_TYPES`, `apps/api/src/job-data-purge.ts:23-31`) and `ContactResearchWorkflow` stores `input_summary_json.jobId` (`workers/automation/src/jobctrl/contact/workflow.py:82-88`), which the API extracts via `json_extract(..., '$.jobId')` (`apps/api/src/read-model.ts:5868`, `:5940`). `learning_recommendation_tombstones.affected_signal_id` also survives under the delete-prevention triggers after its signal cascades away. Neither appears in this table nor in `RETAINED_REFERENCE_TABLES` (`apps/api/src/job-data-purge.ts:77-85`).

Consequence: after a confirmed run the workflow list can still show a contact-research run keyed to a purged job id while the inventory and the no-op verdict present the retained families as complete.

Required outcome: add both rows to this table and to the inventory's retained counts, or state explicitly that the list is non-exhaustive and why.

_Reasoned from code and the exact v9 schema (in-memory build of schema_v7+v8+v9)._

ebarti (commented):
Review of `8ca6adb481b566953ddfa997eb938ebd054ab51a` against base `ed5e530b4d93eca82365c828c6b2b4b88062c4b8` (merge-base with `main`; independent `pr-reviewer` pass, coordinator-published). Posted as a comment review; the gate below is the review's own assessment, not a GitHub changes-request.

**Result:** Gate PASS. No Blocker/High/Medium. Two Low findings posted inline (post-commit compaction failure is reported as a restore-and-retry condition; the retained-reference disclosure omits non-job workflow projections that cite a purged job and learning tombstones). All three prior Medium threads are verified fixed at this head (details below).

**Verified in an isolated checkout at head (clean before and after):** `vitest run test/job-data-purge.test.ts` 16 passed; `vitest run test/permanent-delete-v7.test.ts` 1 passed; `git diff --check` on the range clean. A temp WAL database probe reproduced `PRAGMA data_version` detection inside `BEGIN IMMEDIATE` after `VACUUM INTO` (a foreign commit moved 2 → 3) with no false positive without one. An in-memory build of schema v7+v8+v9 (117 tables) mapped every FK/cascade, tenant column, trigger and AUTOINCREMENT fact against the purge's deletion, preservation and retained lists. Not run here: `api:check`, `docs:build` (CI green at head), and any populated real-shaped workspace (the CLI path is proven only through `spawnSync` against seeded temp workspaces, which use a rollback journal while real workspaces use WAL).

**Prior threads (verified at head, not resolved by this review):**
- Provisional `reconciled_not_found` terminals (discussion_r3945227651): fixed at head. `provisionalWorkflows()` is part of the plan, printed by the CLI with the doc anchor and appended to the refusal; `docs/local-development.md#resolve-provisional-missing-history-executions` gives a guarded exact-row `UPDATE`; the test reads that SQL from the doc, proves it matches only the exact provisional pair, then purges successfully (passed here). The worker reconciler re-probes only `reconciled_not_found` rows, so the cleared code is stable.
- Backup-to-delete window and live-runtime detection (discussion_r3945227654): fixed at head. `plannedDataVersion` is captured before inventory and re-read inside `transaction.immediate()` before any staging or deletion; `freshWorkerCount` keys on the realpath of `db_path`, `component='temporal-worker'`, age ≤ 45 s, checked before backup and again in-transaction; the worker heartbeats every 15 s with resolved paths. Tests for before/after-backup concurrent commits and fresh/stale/other-database heartbeats passed; the WAL probe reproduced detection.
- Retained-reference boundary (discussion_r3945227655): fixed at head through the documentation-plus-inventory route the thread allowed. `RETAINED_REFERENCE_TABLES` and job-unlinked `manual_capture_queue` counts print on inventory, purge and no-op; the storage.md table was added; the retained-rows test preserves stored values with a clean `PRAGMA foreign_key_check`. Two omitted families remain and are the second Low.

**Semantic compatibility with current `main` (`8eaea25a2`):** compatible; merge-tree is conflict-free. `git diff ed5e530b4..8eaea25a2` leaves `permanentlyDeleteJobs`, `purgeJobRows`, the detach helpers, `invalidateOperationsProjections` and `rebuildTenantDeleteProjections` byte-identical; #862 removed only unrelated legacy helpers; #863 added `.immediate()` wrappers to the contact/research/outreach refreshes and `runRefreshPass` plus per-runtime watermark names, none on the delete path (the purge's outer `transaction.immediate()` still nests the inner delete transaction as a savepoint). `db.ts`, `schema-manifest.ts`, `schema_v7/8/9.sql`, `test/v7-schema.ts` and `worker-runtime-telemetry.ts` are unchanged. The purge never touches `event_watermarks`, and with AUTOINCREMENT event ids post-purge events cannot fall below any consumer cursor. Reasoned from the diffs; the purge suite was not executed on a merged tree.

**Investigated and refuted (8):** watermark id reuse after mass event deletion (`job_events.event_id` is AUTOINCREMENT); `listing_ingest`/`canonical_identity`/`content_identity` as undeleted job-stage attempt metrics (they are `dedupe_span` tracing attributes); `PRAGMA foreign_keys` a no-op inside the transaction (enabled at `openDatabase` and verified by `permanentlyDeleteJobs`); `data_version` false positive or negative under WAL (probe); RESTRICT FKs on contacts/research/outreach/attempt metrics breaking deletion (nulled first by `detachIndependentJobReferences`, covered by permanent-delete-v7); projection re-creation from retained events (the predicate covers every DiscoveryRun*/PipelineStep*/lifecycle type the TS and Python folds consume; other job-owned events cascade); heartbeat path mismatch (worker stores the resolved path, purge compares realpaths); API/JSON-RPC reachability (only `purge-job-data.ts` imports the module; no server/main/launcher reference; `data:purge-jobs` is wired root → `@jobctrl/api` → `tsx src/purge-job-data.ts`).

**Residual risk, not counted:** generated-directory renames happen inside the SQLite transaction, so a kill between the renames and COMMIT leaves them in the bundle with rows intact (recoverable, not auto-restored); only the Temporal worker heartbeats, so an API or other writer is caught only when it commits in the guarded window (documented); Temporal history is outside the boundary (documented).

PR #865, opened 2026-09-05
refactor(web): reconcile saved review drafts in one cache

ebarti:
Delayed create, seed, save, render, and comment replies previously competed between the query cache and a second selector in Application review. Saved responses now reconcile once in the apply context, while a late save acknowledgement preserves text or formatting entered after its submitted snapshot.

- Order document state by draft/base-generation identity, revision, explicit rendered-state precedence and server timestamps; merge thread/reply and job feedback identities independently.
- Preserve the initiating tenant/job cache key, accepted-artifact comparison metadata, and the editor’s live focus/dirty state. Remove the view’s five-snapshot selector and reply merger.
- Keep completed-render evidence through promotion to a new active revision-0 draft, including its accepted baseline and risk labels; current editor status belongs to the new draft, and unrelated artifact identities clear the comparison.
- Add reversed-response, replacement-generation, independent-feedback, tenant-switch and save-acknowledgement regressions. The required browser fixtures cover delayed save/seed responses and mounted-editor accessibility.
- Add a test-only isolated API entry point and owned-path checks because ordinary action stubs do not intercept startup provider dispatch. This mode uses synthetic capabilities/credentials, denies provider subprocesses, rejects foreign/dangling-symlink paths and binds seed/teardown state to the same run. Production startup and normal wire contracts are unchanged.
- Include the accepted four-PR implementation plan and update the owning frontend/QA documentation. This is phase 1; later phases will be stacked above it and this stack remains unmerged.

Prior implementation validation (Tier 3):
- `corepack pnpm web:check` and `corepack pnpm api:check`: passed.
- Focused `corepack pnpm --filter @jobctrl/web test` over Application review, mutation/reconciliation and fixture-isolation/config suites: 5 files, 79 tests passed.
- `corepack pnpm --filter @jobctrl/web test-d`: 11 files, 13 tests passed, no type errors.
- `corepack pnpm web:build`, `corepack pnpm --filter @jobctrl/web storybook:build`, `corepack pnpm docs:build`, and `git diff --check`: passed. Existing bundle-size/docgen warnings remain.
- Review reproduced a promotion regression with a faithful new-generation fixture. The correction passes the affected reconciliation/mutation/view suites (3 files, 71 tests) and web typecheck.
- Earlier isolated product QA passed comparison/accessibility, same-job artifact detail, and save-A/type-B/ack-A. A delayed-seed fixture lacked rendered-line markers; it now supplies those markers and waits for a visible newly published thread. Independent review and QA both pass at `2428ce2ddd34a4c24c06298f33c3acfc551ef3c3`; all four isolated Chromium scenarios pass with scoped axe clean. The High review finding is resolved, and all applicable CI checks pass.

No real profile, database, artifacts, credentials, provider calls, or application submission were used for these checks.

## Review corrections and synchronization

- Preserve real typing and formatting while an initially empty draft receives its first saved identity.
- Bind that preservation to the editor session so switching between cached jobs with identical saved documents cannot carry unsaved edits into the newly selected job.
- The corrected view, mutation and reconciler suites pass 73 tests. Six isolated Chromium artifact-comparison scenarios pass, including pending-create typing and cached-job isolation.
- Independent review passed at `55fa84f0ced467ec7e87c8aa9c7faf4faf3ab019`, including the unchanged cached-job counterexample and 73 focused tests.
- Synchronize with current main and preserve the correction through all dependent layers. Range-diff confirms unchanged executable patches; ledger updates distinguish historical evidence from synchronized-head QA and CI.

Cumulative synchronized-head QA passed at `f1c8aae101a727001802d44a3433a6fdb0e28c08`; the completion evidence below is separate from the historical implementation results.

The review-fix web typecheck, web build and docs build also passed; the six Chromium scenarios included scoped accessibility checks.

## Cumulative completion evidence

Independent cumulative QA returned Gate PASS at code head `f1c8aae101a727001802d44a3433a6fdb0e28c08`: 198 guarded core/materials tests, 20 real Temporal cases with 14 balanced ephemeral-server lifecycles, eight real Chromium flows including both review-draft regressions and both structured-profile persistence cases, and three scoped axe checks. No QA case failed or was skipped. Owned pre-import admission, independently parsed JUnit results, process/port cleanup and the clean exact checkout all passed.

[Python CI](https://github.com/ebarti/JobCtrl/actions/runs/34110669959) passed on that code head: Python 3.11, 3.12 and 3.13 each passed lint, 3,742 tests with two unchanged opt-in browser skips, and sdist/wheel builds. [TypeScript CI](https://github.com/ebarti/JobCtrl/actions/runs/34110669878) passed API, web unit/types/build, Storybook, extension and browser E2E. All latest applicable workflows also passed on the synchronized phase heads; superseded cancelled runs were excluded.

### DIFF
--- a/apps/web/src/contexts/materials/components/ResumeAuditPins.tsx
+++ b/apps/web/src/contexts/materials/components/ResumeAuditPins.tsx
@@ -3248,10 +3248,24 @@ export function ResumePlateEditor({
   }, [draft?.latestRevision?.plateDocument, htmlState]);
   const [currentPlateValue, setCurrentPlateValue] = useState<Value | null>(initialPlateValue);
 
+  const reviewDocumentIdentity = draft ? `${draft.draftId}:${draft.baseGeneration}` : artifactId;
+  const savedDocument = useRef({ identity: reviewDocumentIdentity, signature: resumePlateValueSignature(initialPlateValue) });
   useEffect(() => {
-    setCurrentPlateValue(initialPlateValue);
-    setDraftSourceVersion((currentVersion) => currentVersion + 1);
-  }, [initialPlateValue]);
+    const signature = resumePlateValueSignature(initialPlateValue);
+    const previous = savedDocument.current;
+    if (previous.identity === reviewDocumentIdentity && previous.signature === signature) return;
+    savedDocument.current = { identity: reviewDocumentIdentity, signature };
+    // A saved response acknowledges its snapshot, not edits typed after it.
+    // Comment-only publications and identical acknowledgements never remount
+    // Plate, preserving formatting, focus and selection in the live document.
+    const currentSignature = resumePlateValueSignature(currentPlateValue);
+    if (previous.identity !== reviewDocumentIdentity || currentSignature === previous.signature) {
+      if (currentSignature !== signature) {
+        setCurrentPlateValue(initialPlateValue);
+        setDraftSourceVersion((currentVersion) => currentVersion + 1);
+      }
+    }
+  }, [currentPlateValue, initialPlateValue, reviewDocumentIdentity]);
 
   const currentDraftText = useMemo(
     () => (currentPlateValue ? resumeTextFromPlateValue(currentPlateValue) : ""),
--- a/apps/web/src/contexts/materials/components/ResumeAuditPins.tsx
+++ b/apps/web/src/contexts/materials/components/ResumeAuditPins.tsx
@@ -3250,6 +3250,7 @@ export function ResumePlateEditor({
 
   const reviewDocumentIdentity = draft ? `${draft.draftId}:${draft.baseGeneration}` : artifactId;
   const savedDocument = useRef({ identity: reviewDocumentIdentity, signature: resumePlateValueSignature(initialPlateValue) });
+  const submittedDocument = useRef<{ identity: string; signature: string } | null>(null);
   useEffect(() => {
     const signature = resumePlateValueSignature(initialPlateValue);
     const previous = savedDocument.current;
@@ -3262,7 +3263,9 @@ export function ResumePlateEditor({
     // Comment-only publications and identical acknowledgements never remount
     // Plate, preserving formatting, focus and selection in the live document.
     const currentSignature = resumePlateValueSignature(currentPlateValue);
-    if (previous.identity !== reviewDocumentIdentity || currentSignature === previous.signature) {
+    const acknowledgesSubmission = submittedDocument.current?.identity === reviewDocumentIdentity
+      && submittedDocument.current.signature === signature;
+    if (previous.identity !== reviewDocumentIdentity || (!acknowledgesSubmission && currentSignature === previous.signature)) {
       if (currentSignature !== signature) {
         setCurrentPlateValue(initialPlateValue);
         setDraftSourceVersion((currentVersion) => currentVersion + 1);
@@ -3282,6 +3285,14 @@ export function ResumePlateEditor({
     () => resumePlateValueSignature(currentPlateValue),
     [currentPlateValue],
   );
+  const handleSaveDraft = useCallback((source: "autosave" | "manual") => {
+    if (!currentPlateValue || !onSaveDraft) return;
+    submittedDocument.current = {
+      identity: reviewDocumentIdentity,
+      signature: resumePlateValueSignature(normalizeResumePlateValue(currentPlateValue)),
+    };
+    onSaveDraft({ editedText: currentDraftText, plateDocument: currentPlateValue, source });
+  }, [currentDraftText, currentPlateValue, onSaveDraft, reviewDocumentIdentity]);
   const documentKey = `${artifactId}:${draft?.draftId ?? "no-draft"}:${htmlUrl ?? "no-html"}:${draftSourceVersion}:${editorVersion}`;
   const canFormat = formattingApiReady && Boolean(currentPlateValue);
   const draftDirty = Boolean(currentPlateValue && currentDraftSignature !== initialDraftSignature);
@@ -3427,11 +3438,7 @@ export function ResumePlateEditor({
     const handle = window.setTimeout(() => {
       if (!currentPlateValue || lastAutosaveSignature.current === currentDraftSignature) return;
       lastAutosaveSignature.current = currentDraftSignature;
-      onSaveDraft({
-        editedText: currentDraftText,
-        plateDocument: currentPlateValue,
-        source: "autosave",
-      });
+      handleSaveDraft("autosave");
     }, autosaveDelayMs);
     return () => window.clearTimeout(handle);
   }, [
@@ -3442,6 +3449,7 @@ export function ResumePlateEditor({
     draft,
     draftDirty,
     draftLoading,
+    handleSaveDraft,
     onSaveDraft,
     savePending,
   ]);
@@ -3503,14 +3511,7 @@ export function ResumePlateEditor({
           disabled={saveDisabled}
           size="sm"
           type="button"
-          onClick={() => {
-            if (!currentPlateValue) return;
-            onSaveDraft?.({
-              editedText: currentDraftText,
-              plateDocument: currentPlateValue,
-              source: "manual",
-            });
-          }}
+          onClick={() => handleSaveDraft("manual")}
         >
           Save draft
         </Button>
--- a/apps/web/src/contexts/materials/components/ResumeAuditPins.tsx
+++ b/apps/web/src/contexts/materials/components/ResumeAuditPins.tsx
@@ -3255,8 +3255,9 @@ export function ResumePlateEditor({
     const previous = savedDocument.current;
     if (previous.identity === reviewDocumentIdentity && previous.signature === signature) return;
     savedDocument.current = { identity: reviewDocumentIdentity, signature };
-    // The first draft can acquire an identity without changing its saved document.
-    if (previous.signature === signature) return;
+    // Only the initial revision-zero arrival continues the current artifact's
+    // editing session; identical saved documents in another draft still reset it.
+    if (previous.identity === artifactId && draft?.latestRevisionNumber === 0 && previous.signature === signature) return;
     // A saved response acknowledges its snapshot, not edits typed after it.
     // Comment-only publications and identical acknowledgements never remount
     // Plate, preserving formatting, focus and selection in the live document.
@@ -3267,7 +3268,7 @@ export function ResumePlateEditor({
         setDraftSourceVersion((currentVersion) => currentVersion + 1);
       }
     }
-  }, [currentPlateValue, initialPlateValue, reviewDocumentIdentity]);
+  }, [artifactId, currentPlateValue, draft?.latestRevisionNumber, initialPlateValue, reviewDocumentIdentity]);
 
   const currentDraftText = useMemo(
     () => (currentPlateValue ? resumeTextFromPlateValue(currentPlateValue) : ""),
--- a/apps/web/src/contexts/materials/components/ResumeAuditPins.tsx
+++ b/apps/web/src/contexts/materials/components/ResumeAuditPins.tsx
@@ -3255,6 +3255,8 @@ export function ResumePlateEditor({
     const previous = savedDocument.current;
     if (previous.identity === reviewDocumentIdentity && previous.signature === signature) return;
     savedDocument.current = { identity: reviewDocumentIdentity, signature };
+    // The first draft can acquire an identity without changing its saved document.
+    if (previous.signature === signature) return;
     // A saved response acknowledges its snapshot, not edits typed after it.
     // Comment-only publications and identical acknowledgements never remount
     // Plate, preserving formatting, focus and selection in the live document.
--- a/apps/web/src/views/apply-review/ApplyReviewView.tsx
+++ b/apps/web/src/views/apply-review/ApplyReviewView.tsx
@@ -43,6 +43,7 @@ import { useResumeReviewDraftQuery } from "../../contexts/operations/hooks/useRe
 import { useResumeTemplatesQuery } from "../../contexts/profile/hooks/useResumeTemplatesQuery.js";
 import { formatDateTime } from "../../shared/lib/formatters.js";
 import { usePorts } from "../../shared/providers/PortsProvider.js";
+import { useTenantId } from "../../shared/providers/TenantProvider.js";
 import { Alert, AlertDescription, AlertTitle } from "../../shared/ui/alert.js";
 import { Button, buttonVariants } from "../../shared/ui/button.js";
 import {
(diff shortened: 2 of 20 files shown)


apps/web/src/views/apply-review/ApplyReviewView.tsx
```diff
@@ -1078,32 +1078,15 @@ function ResumeLineReview({
...
-    () => selectLatestResumeReviewDraft([renderedDraft, seededDraft, savedDraft, createdDraft, queriedDraft]),
-    [createdDraft, queriedDraft, renderedDraft, savedDraft, seededDraft],
-  );
-  const draft = useMemo(
-    () => mergeDraftThread(baseDraft, replyToComment.data?.thread ?? null),
-    [baseDraft, replyToComment.data?.thread],
-  );
+  const draft = draftQuery.data?.draft.jobKey === item.jobKey ? draftQuery.data.draft : null;
+  const renderedDraftResult = renderDraft.data?.draft.jobKey === item.jobKey
+    && renderDraft.data.draft.draftId === draft?.draftId ? renderDraft.data : null;
```
ebarti: PR Reviewer - final_review

Severity: High
Gate: Fail
Evidence: ApplyReviewView.tsx:1082-1083,1121-1133; reconcileResumeReviewDraft.ts:72-78; apps/api/src/resume-review-drafts.ts:251-300,541-558,1219-1285; apps/api/src/application-feedback.ts:1847-1892.
Reproduction: Reasoned from code; no live API/database execution.

Preserve the completed artifact comparison when rendering creates the next editing draft. On the ordinary Render replacement path, the API promotes draft D and registers a newer accepted material generation. Mutation invalidation refreshes the review queue with its new text/PDF artifact IDs, so the existing draftSeedKey effect calls create/load again. The API cannot return promoted D as an active draft and creates D2 against the new generation. The new reconciler correctly selects D2, but this equality check then discards D's successful render response. comparisonDraftTarget returns null, replacing the completed accepted-versus-rendered comparison with the empty prompt immediately after a successful render. This violates phase 1's required accepted-artifact comparison preservation. The existing comparison unit test refreshes the queue but always returns the original draft from createResumeReviewDraft; the browser fixture keeps the queue fixed, so neither covers the real transition.

Required fix: Keep successful comparison metadata bound to the initiating tenant/job and its accepted/rendered artifact identities across the resulting next-draft creation, while retaining the new draft as the single editing authority and rejecting obsolete-session results. Add a regression that changes both queue artifact IDs after render and makes the subsequent create/load return a new draftId, higher baseGeneration, and revision zero; the original accepted-versus-rendered comparison must remain visible.

ebarti: PR Reviewer - final_review

Verified at 2428ce2ddd34a4c24c06298f33c3acfc551ef3c3. Completed render metadata is now scoped to the initiating tenant/job and remains valid across its source/output artifact pair, independently of the next editable draft. The current draft still owns editor state and validation, the original accepted baseline/risk labels remain visible, and unrelated artifact identities clear the comparison.

Independently ran the ApplyReviewView, useApplyReviewMutations, and reconcileResumeReviewDraft suites: 3 files / 71 tests passed, including the new active generation with revision zero. The revised browser fixture now models the same transition and is covered by the separate QA gate. git diff --check passed. This High finding is resolved.

apps/web/src/contexts/materials/components/ResumeAuditPins.tsx
```diff
@@ -3248,10 +3248,24 @@ export function ResumePlateEditor({
...
-  }, [initialPlateValue]);
+    const signature = resumePlateValueSignature(initialPlateValue);
+    const previous = savedDocument.current;
+    if (previous.identity === reviewDocumentIdentity && previous.signature === signature) return;
+    savedDocument.current = { identity: reviewDocumentIdentity, signature };
+    // A saved response acknowledges its snapshot, not edits typed after it.
+    // Comment-only publications and identical acknowledgements never remount
+    // Plate, preserving formatting, focus and selection in the live document.
+    const currentSignature = resumePlateValueSignature(currentPlateValue);
+    if (previous.identity !== reviewDocumentIdentity || currentSignature === previous.signature) {
```
ebarti: **Medium — identity-only draft arrival resets the live Plate value, discarding text typed while create/load is pending**

Invariant: the unsaved editor value belongs to the mounted session; a late server response may replace it only when it carries a different saved document (plan: "a late acknowledgement cannot replace later local edits").

Evidence: the Plate document renders and is editable as soon as `htmlState` is ready, before any draft exists (L3549-3566; `PlateContent` has no read-only gate), and `usePlateEditor` is keyed by `documentKey` with `initialValue={currentPlateValue}` (L2784-2790, L3560), so a remount alone preserves typing. When the first create/load response arrives with `latestRevisionNumber: 0`, `initialPlateValue` is the unchanged HTML-derived value but `reviewDocumentIdentity` flips from `artifactId` to `draftId:baseGeneration`. This branch then takes `previous.identity !== reviewDocumentIdentity`, `currentSignature !== signature` holds because of the typed text, and `setCurrentPlateValue(initialPlateValue)` erases it. At the base commit the effect depended only on `initialPlateValue`, so this transition did not reset.

Consequence: keystrokes entered while the status reads "loading draft" (create/load can include a template-materials refresh) are silently lost. The same happens after a queue advance while the revision-0 replacement draft is still loading.

Required outcome: an identity change with an unchanged saved-document signature must not replace `currentPlateValue` or bump `draftSourceVersion`. Add a view regression that types during a pending `createResumeReviewDraft`, resolves it with a revision-0 draft, and asserts the typed text is still present with the editor dirty.

_Reasoned from code; no reproduction test was written in review._

ebarti (commented):
Review of `2428ce2ddd34a4c24c06298f33c3acfc551ef3c3` against base `782d3d5ff3c4eac17c8ff1c9be91f89a420bdc07` (independent `pr-reviewer` pass, coordinator-published). Phase 1 of the #865 → #866 → #868 → #869 stack.

**Result:** Gate PASS. No Blocker/High. One Medium posted inline (identity-only draft arrival resets the live editor value and discards text typed while create/load is pending).

**Existing thread:** the earlier High (discussion_r3942266144, completed-render comparison lost on promotion) is independently verified as fixed at this head: `comparisonDraftTarget` keeps the completed render while queue artifacts equal either the source draft's base pair or the render output pair, takes risk labels from the rendered draft, and clears on unrelated identities; the regression returns a new draftId with baseGeneration 3 and revision 0 after both queue artifact IDs change and asserts the comparison persists.

**Verified in an isolated checkout at head:** reconciler, mutations, ApplyReviewView, isolated-workspace and playwright-config suites: 5 files / 80 passed; `web:check`, `api:check`, web `test-d` (11 files / 13) and `git diff --check` clean. Ten in-memory ordering probes against the head reconciler confirmed: a late old-generation promoted response never restores a replaced draft; the later same-generation draft wins in both arrival orders; save rev2 beats promoted rev1; unknown-thread replies are added with their signal; invalid timestamps rank lowest; abandoned never outranks promoted. API semantics checked by code: feedback signals are job-scoped in every draft response, save bumps revision, render promotes and marks threads residual, no delete paths for threads/replies/signals. `apps/api/src/main.ts` does not import the test-only `e2e-server.ts`; contracts and `docs/local-ts-api.md` are unchanged.

**Limits and notes:** Playwright fixtures were not run here (CI Web E2E is green in non-isolated mode; isolated-mode proof is author-supplied). Isolated mode was inspected statically: gated by `JOBCTRL_E2E_ISOLATED=1` plus a marker-owned temp workspace with realpath equality, per-component symlink rejection, `reuseExistingServer` forced false, synthetic provider/credential ports, and teardown limited to the marker-verified directory. Residual: the reconciler treats `baseGeneration` as monotonic per job, which holds today because drafts are only created against approved/active artifacts. Observation: the Phase 1 completion row in `docs/plans/2026-09-06-simplification-stack.md` still cites `6d6f28a18` and says the revised browser fixtures await gates, while this PR's body reports both gates passing at `2428ce2`.

apps/web/src/contexts/materials/components/ResumeAuditPins.tsx
```diff
@@ -3248,10 +3248,27 @@ export function ResumePlateEditor({
...
+    if (previous.identity === reviewDocumentIdentity && previous.signature === signature) return;
+    savedDocument.current = { identity: reviewDocumentIdentity, signature };
+    // Only the initial revision-zero arrival continues the current artifact's
+    // editing session; identical saved documents in another draft still reset it.
+    if (previous.identity === artifactId && draft?.latestRevisionNumber === 0 && previous.signature === signature) return;
+    // A saved response acknowledges its snapshot, not edits typed after it.
+    // Comment-only publications and identical acknowledgements never remount
+    // Plate, preserving formatting, focus and selection in the live document.
+    const currentSignature = resumePlateValueSignature(currentPlateValue);
+    if (previous.identity !== reviewDocumentIdentity || currentSignature === previous.signature) {
```
ebarti: **Low — save acknowledgement overwrites later edits that restore the pre-save baseline**

Invariant (plan, protected behavior): the unsaved editor value belongs to the mounted session; a late acknowledgement cannot replace later local edits.

Evidence: this branch applies the acknowledged snapshot whenever the live signature equals the *previous* baseline rather than the acknowledged snapshot. With autosave (L3416-3435, 1.5 s): baseline S0 → type X (S1) → autosave sends S1 → the user deletes or undoes X back to S0 while the request is in flight → the acknowledgement for S1 lands → `currentSignature === previous.signature` holds and `currentSignature !== signature` holds, so `setCurrentPlateValue(initialPlateValue)` and the `draftSourceVersion` bump run.

Consequence: the text the user just removed reappears and Plate remounts (focus and selection lost). Narrow window (the local save round trip) and not a regression from base, which reset unconditionally, but it is a residual case of the exact behavior this PR sets out to remove.

Required outcome: a late acknowledgement must not replace a live value that differs from the acknowledged snapshot, regardless of whether it equals the prior baseline. Regression: type X, trigger save, delete X before the response resolves, release it, assert the editor shows the baseline (not X) and the editor element identity is unchanged.

_Reasoned from code; no reproduction test was written in review._

docs/local-reliability-qa.md
```diff
@@ -116,6 +116,25 @@ order preserve the saved sequence. Apply **Sort newest first** and verify it
...
+until the later edit is saved, and the rendered revision/comparison do not
+regress. Focused mutation fixtures also cover draft replacement with restarted
+revision numbering, independent replies/feedback, and tenant/job switches.
+
+The isolated browser mode (`JOBCTRL_E2E_ISOLATED=1`) uses the test-only API
+entry point with synthetic capability/credential responses and denied provider
+subprocesses. It requires `JOBCTRL_DOCS_SCREENSHOTS=1`, the existing owned
+screenshot-directory marker, and matching contained app/database/config/state/
+temporary paths before setup, API construction and teardown. Use the existing
+`createOwnedDocsScreenshotDirectory` and `createDocsScreenshotEnvironment`
```
ebarti, line 132: **Low — the documented isolated-mode recipe cannot start as written**

Invariant: the QA matrix entry for a required product proof must be reproducible from the documented steps.

Evidence: `createDocsScreenshotEnvironment` (`scripts/docs-screenshot-environment.mjs:69-83`) sets `JOBCTRL_DOCS_SCREENSHOTS`, `JOBCTRL_E2E_APP_DIR`, `JOBCTRL_E2E_DB_PATH`, `JOBCTRL_E2E_CONFIG_PATH`, `JOBCTRL_E2E_SERVICE_HOME` and the ports, but not `JOBCTRL_E2E_ISOLATED`, `JOBCTRL_DIR`, `JOBCTRL_DB_PATH`, `JOBCTRL_CONFIG_PATH`, nor a contained `TMPDIR` (it forwards the host one). Fed to `assertIsolatedE2eWorkspace` (`apps/web/e2e/fixtures/isolated-workspace.cjs:29-53`), the helper output plus the documented flag fails with `ENOENT: no such file or directory, lstat '<cwd>/undefined'` (undefined `JOBCTRL_DIR` at L35); after adding DIR/DB/CONFIG it fails with `Isolated E2E path escaped its owned workspace` (host `TMPDIR` at L46); it is admitted only once all five are set.

Consequence: an operator following this paragraph cannot start the documented proof; the guard fails closed (no safety impact) but the first error is misleading.

Required outcome: this paragraph lists the complete variable set the guard requires (or points at a helper that produces it) so the documented command runs.

_Reproduced in-memory against the head fixtures with a throwaway marker-owned temp directory._

ebarti (commented):
Review of `6da9a318f78757407cb3a6a25d1f1c5e3c045c89` against base `8eaea25a2fd9034798ef5bac18f2994a33a713c5` (current `main`; independent `pr-reviewer` pass, coordinator-published). Phase 1 of the #865 → #866 → #868 → #869 stack, verified from scratch after the rebase. Posted as a comment review; the gate below is the review's own assessment, not a GitHub changes-request.

**Result:** Gate PASS. Blocker 0, High 0, Medium 0, Low 2 (posted inline: an acknowledgement-versus-baseline edge in the Plate guard; the QA doc recipe omits the variables the isolated guard requires).

**Existing thread (verified at head, not resolved by this review):** the pending-create Medium (discussion_r3945236781) is fixed at head. The same-artifact revision-zero arrival continues the session (`ResumeAuditPins.tsx` L3251-3259) without touching `currentPlateValue` or `draftSourceVersion`; the view regression "preserves formatting while the initial revision-zero draft is loading" passed here and the Chromium typing scenario exists; the deliberate narrowing to the current artifact is justified by the cached-job counterexample, traced by code across all four switch combinations (cached→cached, artifact-only→cached, cached→artifact-only, artifact-only→artifact-only) and covered by the passing view test "keeps unsaved edits scoped when selecting an identical cached draft". The earlier High (completed-render comparison across promotion, thread already resolved) remains fixed: `comparisonDraftTarget` keeps the render while queue artifacts match the source or output pair, and the `it.each` regression passed.

**Verified in an isolated checkout at head (clean before and after):** web vitest, one process at a time: reconciler 5/5, mutations 9/9, isolated-workspace 5/5, playwright-config 4/4, ApplyReviewView 59/59 (82 tests); 18 in-memory reconciler probes against the head module (late old-generation promoted responses in both orders, later same-generation draft in both orders, revision over promoted, unknown-thread replies, invalid timestamps, abandoned versus promoted, ties, old-draft replies, stale thread snapshots after a newer save, signal-review precedence), all passing; a guard-versus-docs probe reproduced the second Low; `git diff --check` clean. Not run here: the isolated Playwright Chromium scenarios, typecheck/lint/build (CI green at head), docs build (the plan's four referenced docs exist).

**Checked by code:** `apps/api/src/main.ts` is unchanged and no production module references `test/e2e-server.ts` or `isolated-workspace.cjs`; the test entry asserts the marker-owned temp workspace before importing config/server and again before listen, supplies a synthetic capability dispatcher and credential store, a throwing Python runtime, the E2E action stub and `requireHealthyWorkerForActions: true`; isolated mode forces `reuseExistingServer: false`, the spec blocks all non-local origins, and teardown deletes only the marker-verified directory. Mutations bind `tenantId` at `onMutate`, publish through one reconciler keyed to the initiating tenant/job, and the draft query is only ever mounted disabled, so invalidations never refetch over reconciled state. Contracts and `docs/local-ts-api.md` are unchanged.

**Investigated and refuted (7):** stale refetch overwriting the reconciled cache (query disabled, sole caller); `ok:false` render/reply crashing the reconciler (contract shapes); callbacks lost on job switch (mutation-level callbacks, `context` carried in query-core 5.100.9); same-generation `createdAt` ties (ms ISO timestamps; ties keep cached; same-base creates reuse the active draft); production reachability of the test server; cross-job carry-over of unsaved edits (traced, test passed); thread-state regression on a newer-revision save (probe).

**Limits and observations, not counted:** Chromium evidence is author-supplied. The Phase 1 ledger row still cites pre-rebase heads `2428ce2dd`/`55fa84f0` and says cumulative synchronized-head QA remains pending while the PR body reports it passed at `f1c8aae10` (also not an ancestor of the current head); conservative rather than wrong, worth refreshing at the final head. The `playwright-config.test.ts` addition is a config-shape assertion; it protects the "never attach to a live server in isolated mode" property and matches the file's existing tests. New reply threads are appended rather than prepended in the unmatched-thread panel (no product impact). Typing during the promotion → new revision-0 draft window is still discarded, as at base.

apps/web/src/contexts/materials/components/ResumeAuditPins.tsx
ebarti: The existing pending-create correction is present at 0490c7f9. The real Chromium regression types while draft creation is held, releases the revision-0 response, and verifies the live text remains dirty. All seven isolated artifact-comparison scenarios passed, including cached-job isolation; independent review and the delayed-save product check also passed.

apps/web/src/contexts/materials/components/ResumeAuditPins.tsx
ebarti: Fixed in 0490c7f9. Manual save and autosave record the submitted normalized document and draft identity; a matching acknowledgement cannot replace a later live value, including undo back to the previous baseline. Both a view regression and a real Chromium type/save/Backspace/held-response regression failed before the fix and now pass, preserving editor identity, focus, dirty state and render gating. Verification: 74 focused tests, seven browser scenarios, web types/build and docs build; independent review and QA passed.

docs/local-reliability-qa.md
ebarti: Fixed in 0490c7f9. The isolated-mode recipe now lists the complete guard input: matching application/database/configuration paths, state and service-home paths, a contained TMPDIR, and both isolation flags. The documented setup was admitted by the guard and used for the seven passing browser scenarios. Independent review and startup/cleanup verification passed.

PR #866, opened 2026-09-05
refactor(web): keep profile drafts structured through editing

ebarti:
Profile editing repeatedly converted API objects into JSON strings and back for the form, structured controls, validation, and Plate projection. The form and editor now exchange immutable object drafts and serialize them when constructing the existing string-valued update request.

- Preserve nested unknown data, unedited fields, and numeric typing in the outgoing request. Schema parsing validates and supplies semantic reads; Plate writes only changed fields into the original draft.
- Retain source-bound conflict detection, selective undo, reordered entries, stale-autosave protection, and the existing profile/style/template request fields. Backend validation and normalized SQLite storage are unchanged; this does not add persistence for unsupported fields.
- Add an isolated test-only preview seam that generates escaped, semantically bound HTML from each current stored profile. Real profile GET/PATCH and preview HTTP routes remain in the browser fixture; generic Python/provider/credential execution stays denied.
- Close test-only execution gaps: explicitly fake/deny irrelevant API dispatch, copy launcher scripts into disposable roots with controlled environments, and isolate Python payload-fixture setup.
- Preserve the clicked native caret before audit-line selection rerenders Slate. Product QA reproduced a race where Slate restored an empty selection before its native selection event arrived; the regression now verifies the target caret, real typing and save/reload. Apply the existing toolbar color rules to its Font Select trigger so text remains readable.
- Record Phase 1's passed review, four browser scenarios and CI, and update owning frontend/QA contracts. This is Phase 2 of four, based on #865; the stack stays unmerged.

Prior implementation validation:
- Focused profile/form/editor/accessibility tests: 61 passed.
- Full web suite after the caret correction: 323 files, 2,020 tests passed; the subsequently added native-caret regression and affected Profile/Apply suites pass 96 tests. Type suite: 13 passed.
- Web and API typechecks, web build, Storybook build, docs build, and diff checks passed. Existing bundle-size/docgen warnings remain.
- New pure preview-fixture test: 1 passed using an explicitly owned environment established before imports.
- Full API suite through the reviewed owned pre-import environment: 59 files, 828 tests passed.
- Earlier implementation and fixture reviews passed. The corrected profile diagnostic passes real boxed/Plate save/reload, entry order, caret assertions and scoped axe. Final independent review and QA passed at a8bb11f3e: zero findings, all six cumulative browser scenarios and three scoped accessibility scans passed. All applicable CI checks passed. No PDF renderer verification is claimed.

## Review corrections and synchronization

- Remove the test that only asserted a future prop was absent, and retain the existing behavior-focused profile regression coverage.
- Reconcile the Phase 2 ledger with the recorded final validation while distinguishing the author-reported isolated browser run from the later published review, which did not rerun those scenarios.
- Independent review passed at `aba0fa2f0f92e86b04093e868a529e5b8768214b`.
- Rebase onto the corrected #865 parent, preserving pending-create typing and cached-job isolation. Range-diff differences are confined to the updated plan ledger.

Cumulative synchronized-head QA passed at `f1c8aae101a727001802d44a3433a6fdb0e28c08`; code-head CI also passed, as recorded below.

Review-fix validation passed 10 actual launcher tests, API typecheck and docs build. Independent review reran those 10 tests successfully.

## Cumulative completion evidence

Independent cumulative QA returned Gate PASS at code head `f1c8aae101a727001802d44a3433a6fdb0e28c08`: 198 guarded core/materials tests, 20 real Temporal cases with 14 balanced ephemeral-server lifecycles, eight real Chromium flows including both review-draft regressions and both structured-profile persistence cases, and three scoped axe checks. No QA case failed or was skipped. Owned pre-import admission, independently parsed JUnit results, process/port cleanup and the clean exact checkout all passed.

[Python CI](https://github.com/ebarti/JobCtrl/actions/runs/34110669959) passed on that code head: Python 3.11, 3.12 and 3.13 each passed lint, 3,742 tests with two unchanged opt-in browser skips, and sdist/wheel builds. [TypeScript CI](https://github.com/ebarti/JobCtrl/actions/runs/34110669878) passed API, web unit/types/build, Storybook, extension and browser E2E. All latest applicable workflows also passed on the synchronized phase heads; superseded cancelled runs were excluded.


docs/plans/2026-09-06-simplification-stack.md
```diff
@@ -355,8 +373,8 @@ Update this table with actual results; do not mark proposals implemented.
...
+| 2 | [#866](https://github.com/ebarti/JobCtrl/pull/866); implementation `1a54c7a6b` | Form/editor/projector JSON round trips removed; object drafts preserve original values and serialize at the request boundary | 61 focused tests; full web 323 files/2020 tests, 13 type tests, web/API checks and web/Storybook builds pass. Docs build and pure preview-fixture test pass. Full API suite deliberately run through the reviewed owned pre-import environment: 59 files/828 tests PASS. | Earlier implementation and fixture reviews PASS; final caret/color delta review and six cumulative browser scenarios pending. Diagnostic profile case passes real save/reload, entry order and scoped axe after fixing the native-caret publication race and toolbar Select colors. Full web suite and web types/build/Storybook rechecked after the shared owner correction; the new native-caret regression and affected Profile/Apply suites pass 96 tests. Renderer-class and line-end fixtures retain explicit caret assertions |
```
ebarti: **Low — Phase 2 completion record is stale relative to this PR's validation claims**

This row still says "final caret/color delta review and six cumulative browser scenarios pending" and cites implementation `1a54c7a6b`, while the PR body reports final review/QA passing at `a8bb11f3e` with six scenarios and three axe scans. This plan is the stack's status ledger ("Update this table with actual results") and Phases 3/4 gate on it.

Required outcome: update the row to the actual outcome at the reviewed head, or state explicitly that those runs are unrecorded.

_Reasoned from docs at head._

apps/api/test/dev-launcher-contract.test.ts
```diff
@@ -134,16 +143,46 @@ function processExists(pid: number): boolean {
...
+    JOBCTRL_USER_ENV_PATH: join(home, "JobCtrl", ".env"),
+    TMPDIR: tempDir,
+    TMP: tempDir,
+    TEMP: tempDir,
     ...extra,
   };
 }
 
 describe("dev launcher contract", () => {
+  it("does not forward ambient env-source or database overrides to launcher fixtures", () => {
```
ebarti: **Low — test asserts a test-local helper rather than launcher behavior**

"does not forward ambient env-source or database overrides" stubs ambient env and checks which keys the test-local `devScriptEnv` helper omits. No `scripts/dev` or product behavior is exercised, and the isolation it guards is already exercised by every launcher test that spawns the copied script with that env. It adds maintenance surface without proving launcher semantics, and this repository has previously declined tests whose subject is other tests.

Required outcome: remove it, or prove the isolation through an observable launcher run (for example, the spawned script cannot observe an ambient `JOBCTRL_DB_PATH`).

_Reasoned from code._

ebarti (commented):
Review of `a8bb11f3e755228fb8f8024160e1e041684f1364` against its parent `2428ce2ddd34a4c24c06298f33c3acfc551ef3c3` (#865 head; independent `pr-reviewer` pass, coordinator-published). Phase 2 of the #865 → #866 → #868 → #869 stack; only this PR's diff was reviewed.

**Result:** Gate PASS. No Blocker/High/Medium. Two Low findings posted inline (stale Phase 2 completion row in the plan ledger; a launcher test that asserts a test-local helper).

**Verified in an isolated checkout at head:** focused web vitest for profile-form, StructuredProfileEditor (+a11y) and ProfileEditor: 4 files / 62 passed; API vitest for e2e-profile-preview, server, qa-workflow, python-runtime: 4 files / 245 passed; dev-launcher-contract: 11 passed; `web:check`, `api:check`, web `test-d` (11 files / 13) pass. An in-memory `ProfileSchema` probe confirmed the write path clones the raw draft and only writes fresh scalar/bullet values, so unknown nested keys, numeric-as-string values and style unknowns survive the round trip; the wire request remains `{profileText, styleText, templateText}` (no contract change). Preview seam escaping and gating (`JOBCTRL_E2E_ISOLATED=1` plus owned-path assertions; production default renderer unchanged), the copied-launcher isolation, deleted `json-record` helpers, and the caret fix's guards were checked by code and existing regressions.

**Limits:** the two new "structured profile persistence" browser scenarios are `test.skip` unless `JOBCTRL_E2E_ISOLATED=1` (`apps/web/e2e/tests/profile-edit.spec.ts:518`), and the CI Web E2E job does not set it, so the six cumulative scenarios exist only as the author's local owned-runner result; they were not run here (no servers/ports in review). Native-caret fix verified by the jsdom regression, not in a real browser. Pre-existing and unchanged, not counted: Plate deltas applied while the boxed draft fails `ProfileSchema` return an unchanged result with no warning (`profile-form.tsx:150`).

docs/plans/2026-09-06-simplification-stack.md
ebarti: The corrected Phase 2 record is retained at cd8f88b6. It identifies the validated a8bb11f3e head and accurately distinguishes the author-reported six browser scenarios and three accessibility scans from the later independent review, which did not rerun those scenarios. The stale pending statement is gone.

apps/api/test/dev-launcher-contract.test.ts
ebarti: The test-local-helper-only assertion was removed in the retained structured-profile review correction at cd8f88b6. The launcher tests continue exercising spawned launcher behavior with their isolated environment. This restack preserved every reviewed patch unchanged.

PR #871, opened 2026-09-07
fix(dashboard): resolve source names without hiding failures

ebarti:
Source health and blocked-source digest cards exposed internal source IDs even when the registry had a readable name. Resolve display names in the API and worker digest, normalize existing JobStreaming source labels, and render the name with the stable ID as a fallback. Failure counts, block reasons, source identities, and retry information remain visible.

Adds optional contract fields, API and UI regressions for named and legacy payloads, and the owning API documentation.

Validation on cumulative head `a72c3078008482345c7ff14937cf8bc02511540d`:

- Ruff, API/web typechecks, all 828 API tests, 9 focused dashboard tests, 13 web type tests, production web build, docs build, and `git diff --check` pass. The fresh full Python run passed **3841 tests with 2 optional skips**.
- Independent cumulative review: **Gate: PASS**, no findings; 48 guarded Python cases, 4 API digest cases and all 9 dashboard cases passed.
- Independent cumulative product QA: **Gate: PASS**, no findings; 29 material-validation cases and 30 Temporal cases passed, with 19 real owned server lifecycles and balanced shutdowns. Three actual API surfaces and three desktop/mobile browser scenarios preserved canonical source data, failure/block information, readable labels and Inspect navigation; scoped accessibility, console and network checks passed.
- Prior SDK isolation and timeout/reservation gates were carried forward after source and dependency equality checks. The separate authenticated-provider/owner-extension operational requirement on #860 remains open.

These three PRs extend the existing unreleased stack. Owning contract and safety documentation accompanies the relevant phase; the final recovery PR carries the shared canonical product documentation and cumulative QA checklist.

This PR’s latest [Python CI](https://github.com/ebarti/JobCtrl/actions/runs/34125336936) and [TypeScript CI](https://github.com/ebarti/JobCtrl/actions/runs/34125336912) passed at its published head. All five TypeScript jobs passed, including E2E and Storybook; other applicable workflows passed and DCO was explicitly skipped. The final cumulative Python CI also passed lint, all 3841 tests (2 optional skips), and package builds on Python 3.11, 3.12 and 3.13.

The first Python 3.13 attempt timed out waiting for the unchanged worker-loss fixture’s 30-second checkpoint signal. The original failure was retained; one [failed-job-only retry at the same commit](https://github.com/ebarti/JobCtrl/actions/runs/34125336936/attempts/2) passed. No fixture, assertion or product source was changed for the retry.

PR #872, opened 2026-09-07
fix(pipeline): recover automatic preparation with exact ownership

ebarti:
Saved preparation work could remain stranded after worker interruption, a lost workflow-start acknowledgement, or the handoff from discovery's streaming enrichment to its terminal pass. Resume eligible Enrich, Score, Tailor, and Cover work from startup and heartbeat using bounded durable reservations and exact Temporal execution ownership. Preserve cancellation, attempts, cooldowns, spend limits, approved artifacts, and persistence fences; recovery does not start Discover or Apply.

Persist requirement-fit evidence with owned scores and release only the matching retryable missing-evidence Tailor block after a valid rescore. Reconcile closed owners before dispatch and prevent disconnected predecessors from committing late results. Provider login refresh/revocation endpoints retain their required HTTPS defaults after ambient credentials are cleared.

Includes deterministic state regressions, real Temporal recovery/cancellation fixtures, and owning runtime, operations, enrichment, and security documentation.

Validation on cumulative head `a72c3078008482345c7ff14937cf8bc02511540d`:

- Ruff, API/web typechecks, all 828 API tests, 9 focused dashboard tests, 13 web type tests, production web build, docs build, and `git diff --check` pass. The fresh full Python run passed **3841 tests with 2 optional skips**.
- Independent cumulative review: **Gate: PASS**, no findings; 48 guarded Python cases, 4 API digest cases and all 9 dashboard cases passed.
- Independent cumulative product QA: **Gate: PASS**, no findings; 29 material-validation cases and 30 Temporal cases passed, with 19 real owned server lifecycles and balanced shutdowns. Three actual API surfaces and three desktop/mobile browser scenarios preserved canonical source data, failure/block information, readable labels and Inspect navigation; scoped accessibility, console and network checks passed.
- Prior SDK isolation and timeout/reservation gates were carried forward after source and dependency equality checks. The separate authenticated-provider/owner-extension operational requirement on #860 remains open.

These three PRs extend the existing unreleased stack. Owning contract and safety documentation accompanies the relevant phase; the final recovery PR carries the shared canonical product documentation and cumulative QA checklist.

This PR’s latest [Python CI](https://github.com/ebarti/JobCtrl/actions/runs/34125358301) and [TypeScript CI](https://github.com/ebarti/JobCtrl/actions/runs/34125358399) passed at its published head. All five TypeScript jobs passed, including E2E and Storybook; other applicable workflows passed and DCO was explicitly skipped. The final cumulative Python CI also passed lint, all 3841 tests (2 optional skips), and package builds on Python 3.11, 3.12 and 3.13.

### DIFF
--- a/workers/automation/src/jobctrl/discovery/activities.py
+++ b/workers/automation/src/jobctrl/discovery/activities.py
@@ -356,7 +356,13 @@ async def discovery_enrichment_activity(
         )
     )
 
-    cancel_event = threading.Event()
+    from jobctrl.enrichment.activities import _ActivityCancellationEvent
+
+    # Finishing the source producers cancels the live consumer before the
+    # terminal enrichment pass. That activity stop is a handoff, not a user
+    # cancellation of the job. Actual workflow cancellation is settled by the
+    # exact-run cancellation reconciler after the workflow records its outcome.
+    cancel_event = _ActivityCancellationEvent(terminal_on_cancel=not payload.stream_while_discovering)
     on_job_enriched = _build_per_job_handoff(payload)
     try:
         info = activity.info()
@@ -389,7 +395,7 @@ async def discovery_enrichment_activity(
             starting_message="discovery enrichment starting",
             progress_message="discovery enrichment still running",
             poll_interval=1.0 if payload.stream_while_discovering else 15.0,
-            on_cancel=cancel_event.set,
+            on_cancel=cancel_event.request_stop,
             activity_name="discover:enrichment",
         )
         activity.heartbeat({"status": result.get("status", "ok")})
--- a/workers/automation/src/jobctrl/pipeline/automatic_preparation.py
+++ b/workers/automation/src/jobctrl/pipeline/automatic_preparation.py
@@ -1,0 +1,762 @@
+"""Resume durable preparation backlogs without starting another discovery run.
+
+The queued stage row is the dispatch reservation. Its frozen batch survives a
+lost Temporal acknowledgement; the deterministic workflow ID rejects a second
+execution of that reservation. Each workflow runs only one stage, so every
+successor is selected afresh through the canonical prerequisite predicates.
+"""
+
+from __future__ import annotations
+
+from dataclasses import dataclass
+from datetime import datetime, timezone
+import hashlib
+import json
+import sqlite3
+from typing import Any
+
+from temporalio.client import WorkflowExecutionStatus
+from temporalio.api.enums.v1 import TimeoutType
+from temporalio.common import WorkflowIDConflictPolicy, WorkflowIDReusePolicy
+from temporalio.exceptions import WorkflowAlreadyStartedError
+from temporalio.service import RPCError, RPCStatusCode
+
+from jobctrl import database
+from jobctrl.domain.identifiers import JobId, canonical_job_id
+from jobctrl.domain.tenant import LOCAL_TENANT
+from jobctrl.infrastructure.enrichment.sqlite_repository import SqliteEnrichmentRepository
+from jobctrl.state import record_job_event, set_stage_state
+
+_STAGES = ("score", "tailor", "cover", "enrich")
+_BATCH_SIZE = 25
+_ACTIVE_WORK = (
+    "ExecutionStatus = 'Running' AND (WorkflowType = 'DiscoverWorkflow' "
+    "OR WorkflowType = 'JobPipelineWorkflow' OR WorkflowType = 'JobPreparationWorkflow' "
+    "OR WorkflowType = 'JobUrlImportWorkflow' OR WorkflowType = 'ManualCaptureImportWorkflow')"
+)
+
+
+@dataclass(frozen=True)
+class RecoveryBatch:
+    workflow_id: str
+    stage: str
+    job_ids: tuple[JobId, ...]
+    min_score: int
+
+    def metadata(self) -> dict[str, Any]:
+        return {
+            "workflowId": self.workflow_id,
+            "stage": self.stage,
+            "jobIds": list(self.job_ids),
+            "minScore": self.min_score,
+        }
+
+
+async def reconcile_automatic_preparation(
+    client: Any,
+    *,
+    task_queue: str,
+    expected_app_dir: str,
+    expected_db_path: str,
+) -> int:
+    """Dispatch at most one bounded batch while the preparation worker is idle."""
+    from jobctrl.infrastructure.scoring.criteria_provider import read_min_fit_score
+    from jobctrl.llm import read_spend_budget_status
+    from jobctrl.pipeline.workflow import JobPipelineWorkflow, JobPipelineWorkflowInput
+
+    # Visibility errors propagate before any reservation is written. The worker
+    # retries this read on its next heartbeat instead of guessing that it is idle.
+    async for _execution in client.list_workflows(query=_ACTIVE_WORK):
+        return 0
+    if read_spend_budget_status().exceeded:
+        return 0
+    conn = database.get_connection()
+    await _reconcile_stopped_enrichment_owners(client, conn)
+    await _reconcile_stopped_activity_owners(client, conn)
+    await _reconcile_interrupted_reservations(client, conn)
+    batch = reserve_recovery_batch(conn, min_score=read_min_fit_score(default=7))
+    if batch is None:
+        return 0
+    try:
+        description = await client.get_workflow_handle(batch.workflow_id).describe()
+    except RPCError as exc:
+        if exc.status != RPCStatusCode.NOT_FOUND:
+            raise
+    else:
+        if description.status != WorkflowExecutionStatus.RUNNING:
+            _settle_closed_reservation(conn, batch, description.status)
+        return 0
+
+    payload = JobPipelineWorkflowInput(
+        tenant_id=str(LOCAL_TENANT),
+        stages=[batch.stage],
+        job_ids=batch.job_ids,
+        min_score=batch.min_score,
+        workers=1,
+        expected_app_dir=expected_app_dir,
+        expected_db_path=expected_db_path,
+        automatic_recovery=True,
+    )
+    try:
+        await client.start_workflow(
+            JobPipelineWorkflow.run,
+            payload,
+            id=batch.workflow_id,
+            task_queue=task_queue,
+            id_conflict_policy=WorkflowIDConflictPolicy.USE_EXISTING,
+            id_reuse_policy=WorkflowIDReusePolicy.REJECT_DUPLICATE,
+        )
+    except WorkflowAlreadyStartedError:
+        # Another worker or a lost acknowledgement already dispatched this
+        # exact reservation. Never mint a new ID to get around that decision.
+        return 0
+    return len(batch.job_ids)
+
+
+async def _reconcile_stopped_enrichment_owners(client: Any, conn: sqlite3.Connection) -> None:
+    """Fence a dead enrichment execution before releasing its durable cohort."""
+    from jobctrl.domain.enrichment import StaleEnrichmentExecutionLease
+    from jobctrl.domain.enrichment.aggregate import JobEnrichment
+    from jobctrl.domain.enrichment.value_objects import EnrichmentError, ExtractionTier
+    from jobctrl.enrichment.detail import cancel_enrichment_cohort
+    from jobctrl.infrastructure.enrichment.execution_lease import claim_enrichment_execution_lease_for_run
+
+    owners = conn.execute(
+        "SELECT DISTINCT json_extract(metadata_json, '$.workflowId') AS workflow_id, "
+        "json_extract(metadata_json, '$.temporalRunId') AS run_id FROM job_stage_states "
+        "WHERE tenant_id = ? AND stage = 'enrich' AND state IN ('queued', 'running') "
+        "AND json_extract(metadata_json, '$.workflowId') GLOB 'prepare-auto-local-*' "
+        "AND json_extract(metadata_json, '$.temporalRunId') IS NOT NULL",
+        (str(LOCAL_TENANT),),
+    ).fetchall()
+    for owner in owners:
+        workflow_id, run_id = owner["workflow_id"], owner["run_id"]
+        try:
+            description = await client.get_workflow_handle(workflow_id, run_id=run_id).describe()
+        except RPCError as exc:
+            if exc.status == RPCStatusCode.NOT_FOUND:
+                continue
+            raise
+        if description.status == WorkflowExecutionStatus.RUNNING:
+            continue
+        job_ids = tuple(
+            canonical_job_id(row[0])
+            for row in conn.execute(
+                "SELECT job_id FROM job_stage_states WHERE tenant_id = ? AND stage = 'enrich' "
+                "AND state IN ('queued', 'running') AND json_extract(metadata_json, '$.workflowId') = ? "
+                "AND json_extract(metadata_json, '$.temporalRunId') = ?",
+                (str(LOCAL_TENANT), workflow_id, run_id),
+            ).fetchall()
+        )
+        if description.status == WorkflowExecutionStatus.CANCELED:
+            cancel_enrichment_cohort(conn, job_ids, workflow_id=workflow_id, workflow_run_id=run_id)
+            continue
+        try:
+            claim_enrichment_execution_lease_for_run(
+                conn,
+                tenant_id=LOCAL_TENANT,
+                workflow_id=workflow_id,
+                run_id=run_id,
+                owner_token=f"recovery:{workflow_id}:{run_id}",
+                activity_phase=3,
+                activity_attempt=1,
+            )
+        except StaleEnrichmentExecutionLease:
+            # Cancellation may have committed its fence but lost the process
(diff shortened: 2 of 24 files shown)


ebarti (commented):
Review of `36fe02dff14f5efcfe3867f58e6d98074f5271e4` against its parent `5844354e9076d325ace9de09bd6b9c9aea94c145` (#871 head; independent `pr-reviewer` pass, coordinator-published). Only this PR's diff was reviewed (22 files, 1 commit), including the full new `automatic_preparation.py`, the callers it fences, and the Temporal-backed fixtures. Posted as a comment review; the gate below is the review's own assessment, not a GitHub changes-request.

**Result:** Gate FAIL. Blocker 0, High 2, Medium 2, Low 0. All four findings are inline: (1) High, Enrich batches that fail before any job claim are re-reserved forever without consuming attempts; (2) High, a real Discover cancel during streaming enrichment releases queued rows into automatic recovery instead of canceling them; (3) Medium, the owned scorer's repost-reuse branch leaves the row `running` until the reconciler relabels it as an orphan restoration; (4) Medium, one job's ownership loss mid-batch fails the activity and blocks the rest of the batch as non-retryable.

**Verified in an isolated checkout at head (clean before and after; `uv.lock` untouched):** Ruff on the 18 touched Python files (pass); `pytest tests/test_automatic_preparation.py tests/test_discovery_enrichment_handoff.py tests/test_codex_home_isolation.py tests/test_scorer.py` (100 passed); `pytest tests/test_automatic_preparation_temporal_qa.py` with its own ephemeral dev server via the local `temporal` CLI (33 passed, 66 s, no leftover processes); `git diff --check` (clean); six scratch reproductions kept outside the repository (6 passed: five confirm the four findings, one refutes the idle-scan-cost candidate). Not run here: docs build and the full Python suite (CI green at head).

**Investigated and refuted (8):** transaction leak in `_reconcile_stopped_activity_owners` (both helpers commit internally); startup/heartbeat double dispatch (the startup reconcile completes before the heartbeat task is created; a second process reattaches to the same deterministic reservation ID under `USE_EXISTING`); idle-heartbeat scan cost (`_candidates` over 4000 jobs measured at 62 ms); missing coverage for the requirement-fit block release (four `evidence_world` cases in the Temporal QA file exercise the release and non-release paths); canceled, cooled-down or spend-capped rows resumed by the selector (`_retry_due` accepts only `pending`/`failed` retryable rows under the attempt cap and cooldown, and the global budget gate returns before reserving); approved artifacts regenerated (`pending_tailor` requires a NULL tailored resume path and `pending_cover` a ready resume with PDF and no cover); recovery starting Discover or Apply (`_STAGES = ("score", "tailor", "cover", "enrich")`; `_ACTIVE_WORK` defers to running Discover/import workflows); Codex HTTPS defaults and credential exposure (the pinned Codex 0.144.4 binary carries `https://auth.openai.com/oauth/token` and `/oauth/revoke` as its own defaults, the literal values override the blanked `CODEX_NEUTRALIZED_AUTH_ENV` keys, and nothing logs them).

**Residual risk and observations, not counted:** the idle heartbeat holds a `BEGIN IMMEDIATE` lock across full candidate scans every 15 s (small today, grows with the jobs table); `README.md` and `docs/local-reliability-qa.md` are not updated for the new background behavior and the three new suites, which the PR body defers to the final PR of the stack, and `docs/plans/2026-09-06-simplification-stack.md` does not list #871–#873; finding (1) was established by code and a reproduction of the state machine, not by driving a real batch-level scraper failure through Temporal.

workers/automation/src/jobctrl/pipeline/automatic_preparation.py
```diff
@@ -0,0 +1,762 @@
...
+                                retryable=True,
+                            ),
+                            finished_at=now,
+                        )
+                        repository.save(aggregate, commit=False)
+                        attempts += 1
+                    if aggregate is not None:
+                        attempts = max(attempts, aggregate.attempt_count)
+                        error = aggregate.last_attempt.error if aggregate.last_attempt else None
+                    state = "exhausted" if attempts >= min(5, row["max_attempts"]) else "failed"
```
ebarti, line 225: **High — Enrich recovery re-dispatches indefinitely when a batch fails before any job claim**

Invariant (`operations.md` in this PR): automatic retries are bounded by attempts (`min(stage limit, 5)`), and "batches without a consumed attempt retain their stopped/block outcome, preventing a preflight retry loop."

Evidence: this function settles *queued* rows of a closed automatic execution as `failed`/retryable with `attempt_count` unchanged (L184-236; only `running` rows charge an attempt at L200-217). `_run_selected_enrichment` binds `temporalRunId` before the scraper runs (`enrichment/activities.py:286-297`), so any failure after that point and before the first per-job claim lands here. The scraper's own transient-error path does the same without an attempt and also strips `automaticPreparation` (`enrichment/detail.py:3363-3437`, called from L4042-4060 and L4119-4130). `_retry_due` derives the cooldown from the unchanged count (L544), so the row is due again after a constant 60-120 s and gets a fresh workflow ID (L362-365). Score/Tailor/Cover are protected by the consumed-attempt guard in `_release_interrupted_reservations` (L648-666) and by `_settle_closed_reservation`; Enrich has no equivalent.

Consequence: a persistent batch-level enrichment failure (a network outage raising `TransientNetworkError` in the LinkedIn pre-pass or first navigation, a gateway/browser setup exception, or an activity killed before its first claim) makes the worker start a new `prepare-auto-local-enrich-*` workflow and write lease plus `StageQueued`/`StageFailed` events for up to 25 jobs every cooldown period, indefinitely, with no attempt accounting; the documented ceiling never engages.

Required outcome: an Enrich batch whose exact execution closed without any consumed attempt keeps the same stopped/block outcome as the other stages, or each redispatch charges a durable attempt so the ceiling engages; add a regression test for the queued-row branch of this reconciler.

_Reproduced (scratch test outside the repository): four consecutive reserve → bind run ID → FAILED → reconcile cycles on one job produced four distinct workflow IDs with `attempt_count` fixed at 1; the scraper-path variant left the row `pending` with no `automaticPreparation` metadata and a new workflow ID on the next reservation._

workers/automation/src/jobctrl/discovery/activities.py
```diff
@@ -356,7 +356,13 @@ async def discovery_enrichment_activity(
...
     )
 
-    cancel_event = threading.Event()
+    from jobctrl.enrichment.activities import _ActivityCancellationEvent
+
+    # Finishing the source producers cancels the live consumer before the
+    # terminal enrichment pass. That activity stop is a handoff, not a user
+    # cancellation of the job. Actual workflow cancellation is settled by the
+    # exact-run cancellation reconciler after the workflow records its outcome.
+    cancel_event = _ActivityCancellationEvent(terminal_on_cancel=not payload.stream_while_discovering)
```
ebarti, line 365: **High — stopping a Discover run during streaming enrichment now leaves its queued cohort to automatic recovery instead of canceling it**

Invariant: a user cancellation terminalizes the exact enrichment cohort (PR body: "Actual workflow cancellation is settled by the exact-run cancellation reconciler"; `docs/user/enrichment-and-extraction.md` in this PR: "Canceled jobs … remain stopped"). At base the streaming consumer used a plain `threading.Event`, so `_settle_interrupted_enrichment_cohort` took the cancel branch and these rows became `canceled`.

Evidence: with `terminal_on_cancel=False`, a real cancel is indistinguishable from the handoff at the activity boundary, so `_settle_interrupted_enrichment_cohort` → `_release_unstarted_enrichment_cohort`, which rewrites queued rows to `pending` with metadata `{"recoveryReason": "workflow_selection_unprocessed"}` and drops `workflowId`/`temporalRunId` (`enrichment/detail.py:3407-3420`). `DiscoverWorkflow` runs no cancel-cohort activity on `CancelledError` (`discovery/workflow.py:128-138`). The heartbeat reconciler `_reconcile_canceled_enrichment_cohorts` (`cli.py:2769-2866`) joins stage rows to canceled projections through `json_extract(metadata_json, '$.workflowId')`, and `cancel_enrichment_cohort` skips rows whose metadata does not carry the run (`detail.py:3532-3535`), so released rows are unreachable. `_candidates`/`_retry_due` then treat those `pending` rows as due 60 s later.

Consequence: pressing Stop on a Discover run mid-stream no longer stops the queued enrichment jobs; automatic recovery enriches them within about a minute and then flows them into Score/Tailor/Cover (LLM spend), while rows that still carry the run's metadata are canceled by the reconciler, leaving one cohort split between canceled and auto-processed.

Required outcome: a real workflow cancellation still terminalizes the released queued rows (keep the run identity on released rows, or cancel by cohort/lease evidence), with a regression test covering cancel-during-streaming → canceled outcome → reconciler → no recovery reservation.

_Reproduced (scratch test outside the repository): the PR's own handoff scenario followed by a `canceled` projection row: `_reconcile_canceled_enrichment_cohorts` returned 0, an exact-cohort `cancel_enrichment_cohort` returned 0, the row stayed `pending`, and `reserve_recovery_batch` then reserved it for Enrich._

workers/automation/src/jobctrl/scoring/scorer.py
```diff
@@ -787,7 +838,7 @@ def score_job_by_id(
             job=job,
             source_score=reusable_repost_score,
         )
-        if outcome.ok and outcome.score is not None:
+        if outcome.ok and outcome.score is not None and not enforce_workflow_ownership:
```
ebarti: **Medium — owned scoring's repost-reuse branch returns without finishing the stage; the row is later relabeled as an orphaned-owner restoration**

Invariant: every automatic Score execution ends its own `running` row with an accurate terminal write and audit event (root-cause/auditability discipline: displayed history must have an explicit source of truth).

Evidence: the owned path moves the row to `running` up front (L762-801), but this branch skips `_record_score_stage_succeeded` and returns after `_persist_reused_score` commits the score through `_OwnedScoreRepository`. Nothing else finishes the row: the activity reports success and the workflow completes with the row still `running`. On the next heartbeat `_reconcile_stopped_activity_owners` → `recover_preparation_state_rows` → `_restore_committed` writes `succeeded` with `reason: orphaned_activity_restored_committed_score`, message "Committed result restored after its activity owner stopped.", and `attempt_count + 1` (`preparation_recovery.py:372-411, 462-471`). `_sync_score_eligibility_stage_state` (threshold skips and eligibility blocks for Tailor) never runs for that job. The `existing is not None` branch at L851-859 has the same shape. Reachable for reference-repost candidates (`_preferred_direct_score_for_repost`, L1429).

Consequence: false "owner stopped" audit history for a normal completion, an extra attempt charged, the `busy` guard stalling recovery for one tick, and the downstream Tailor row missing its below-threshold/eligibility outcome until something else reconciles it.

Required outcome: the owned reuse and existing-score branches write the same fenced `succeeded` transition, `StageCompleted` event, and eligibility sync as the LLM path; add a test asserting the row state immediately after an owned reuse.

_Reproduced (scratch test outside the repository) with only the reuse decision stubbed: after the owned call, state `running` and 0 `StageCompleted` events; after the reconciler, `succeeded` with `attempt_count=1` and the orphan reason._

workers/automation/src/jobctrl/scoring/tailor.py
```diff
@@ -616,6 +619,10 @@ def tailor_job_by_id(
         raise RuntimeError("tailor activity canceled before dispatch")
     stable_job_id = canonical_job_id(str(job_id))
     conn = get_connection()
+    if recovery_workflow_id and not owns_preparation_reservation(
+        conn, tenant_id=tenant_id, job_id=stable_job_id, stage="tailor", workflow_id=recovery_workflow_id,
+    ):
+        raise RuntimeError("tailor activity no longer owns its queued reservation")
```
ebarti: **Medium — losing one job's reservation mid-batch fails the whole automatic Tailor/Score batch and strands the rest as non-retryable blocks**

Invariant: recovery preserves cancellation and deletion per job and must not convert other jobs' work into a stopped state that automatic recovery never revisits.

Evidence: this check raises `RuntimeError` when the reservation is no longer owned, including a job deleted after the activity's start-of-batch prefilter (`owns_preparation_reservation`, `preparation_recovery.py:15-30`); the scorer raises the same way at L779. `run_material_jobs` calls `run_one` with no per-job exception handling (`materials/executor.py:31-36`) and `_run_selected_scores` likewise. With `maximum_attempts=1` the workflow fails; on the next tick `_reconcile_interrupted_reservations` releases only on Temporal timeouts (`automatic_preparation.py:609-621`), so `reserve_recovery_batch` → `_settle_closed_reservation` marks every still-queued job `blocked`/`PREPARATION_RECOVERY_STOPPED`, `retryable=0` (L723-758), a state `_retry_due` never reconsiders. Cover handles the same condition as a per-job skip (`cover_letter.py:612-615`), so the stages are inconsistent.

Consequence: deleting one job while a 25-job automatic Tailor batch runs (tens of minutes) leaves up to 24 unrelated jobs blocked with "Automatic preparation stopped before consuming this reservation." until each is retried manually. The same settlement also mislabels jobs that a *completed* batch merely skipped as ineligible.

Required outcome: a per-job ownership/deletion failure is recorded against that job (skip or fail) without aborting the batch, or the closed-batch settlement does not block never-attempted jobs after a non-timeout failure; add a regression test for ownership loss between prefilter and dispatch.

_Reproduced the raise for a job deleted after reservation (scratch test outside the repository); propagation to the activity and the blocked settlement are reasoned from the cited code and the PR's own `test_unconsumed_closed_batch_stops_instead_of_restarting_forever` (FAILED → blocked)._

workers/automation/src/jobctrl/pipeline/automatic_preparation.py
ebarti: Fixed in 73332bb134f8d1f81001513cd12252c99e8e2232. Automatic Enrich keeps unclaimed rows under their original reservation when the scraper releases its cohort or the exact execution stops. The closed-reservation policy now blocks a batch that consumed no attempt; the existing exact-timeout/consumed-attempt rule remains responsible for releasing eligible leftovers.

Regression coverage exercises both queued-owner and scraper-release paths over repeated reconciliation, preserving the attempt count and a single reservation. A real Temporal fixture raises a transient site failure before the first claim and verifies the stopped outcome across three later recovery ticks with only one workflow execution.

workers/automation/src/jobctrl/discovery/activities.py
ebarti: Fixed in 73332bb134f8d1f81001513cd12252c99e8e2232. Releasing the streaming cohort retains its exact workflow and Temporal run identity. This preserves the normal terminal-pass handoff while allowing the cancellation reconciler to terminalize released pending rows when that Discover run is canceled.

The handoff regression now covers both normal terminal reclamation and cancellation immediately after release: the canceled-run projection reconciles the exact row and automatic recovery cannot reserve it. The existing Discover cancellation, lease fencing, and recovery suites also pass.

workers/automation/src/jobctrl/scoring/scorer.py
ebarti: Fixed in 73332bb134f8d1f81001513cd12252c99e8e2232. Owned repost reuse, existing-score reuse, and fresh scoring share the same fenced success finalization, StageCompleted event, attempt accounting, and eligibility synchronization.

Regressions assert immediate succeeded state after real repost matching, preservation of prior attempt count, below-threshold Tailor state for an existing score, and refusal to overwrite cancellation between reuse selection and finalization.

workers/automation/src/jobctrl/scoring/tailor.py
ebarti: Fixed in 73332bb134f8d1f81001513cd12252c99e8e2232. Revoked Score reservations return a per-job failure outcome and revoked Tailor reservations return a skip, so the batch continues to other jobs. The atomic material admission check raises a dedicated reservation-loss exception handled per job in both sequential and parallel execution; activity cancellation and other exceptions retain their existing behavior.

Real Temporal regressions revoke the first reservation after cohort filtering in Score and Tailor, then verify that the other job consumes its own attempt and never receives PREPARATION_RECOVERY_STOPPED. Additional sequential/parallel material tests cover reservation loss at admission.

ebarti:
PR Fixer - pr872

Fixed all four review findings in 73332bb134f8d1f81001513cd12252c99e8e2232 and pushed the existing PR branch:

- Unclaimed Enrich failures keep their original bounded reservation and stop instead of redispatching indefinitely.
- Discover handoff retains exact execution identity so a real workflow cancellation reaches the released cohort.
- Owned Score reuse finalizes its success audit, attempts, and downstream eligibility under the same persistence fence as fresh scoring.
- Losing one job reservation no longer aborts the remaining Score/Tailor cohort; atomic material admission loss is isolated per job.

All four inline threads were replied to and resolved. No findings were dismissed. Remaining Blocker/High findings: None.

Tier 3 verification: the relevant 16-suite Python matrix passed **373 tests** in 76.36 seconds; independent review and QA both returned **Gate: PASS**, with **125 tests** passing independently in 68.90 seconds. Ruff passed on all ten changed Python files; `git diff --check` and the staged check passed. The tests include real Temporal workflows and canonical SQLite state with synthetic browser/LLM boundaries. No production data or authenticated providers were used. The full Python suite and CI on this new head are not represented by these focused local results.

PR #873, opened 2026-09-07
fix(tailor): preserve grounded role and summary evidence

ebarti:
A required experience role without achievement evidence was forced to generate a positioning bullet, while summary rewriting was instructed to preserve unsupported baseline quantities. Retain such a role's canonical details with an empty bullet list, and allow unsupported summary quantities to become qualitative wording while preserving grounded metrics and required evidence.

Align response schema, prompt, deterministic curation, repair guidance, and evidence provenance. Add regressions for evidence-free required roles and summary grounding, and update canonical product documentation plus the cumulative recovery QA checklist.

Validation on cumulative head `a72c3078008482345c7ff14937cf8bc02511540d`:

- Ruff, API/web typechecks, all 828 API tests, 9 focused dashboard tests, 13 web type tests, production web build, docs build, and `git diff --check` pass. The fresh full Python run passed **3841 tests with 2 optional skips**.
- Independent cumulative review: **Gate: PASS**, no findings; 48 guarded Python cases, 4 API digest cases and all 9 dashboard cases passed.
- Independent cumulative product QA: **Gate: PASS**, no findings; 29 material-validation cases and 30 Temporal cases passed, with 19 real owned server lifecycles and balanced shutdowns. Three actual API surfaces and three desktop/mobile browser scenarios preserved canonical source data, failure/block information, readable labels and Inspect navigation; scoped accessibility, console and network checks passed.
- Prior SDK isolation and timeout/reservation gates were carried forward after source and dependency equality checks. The separate authenticated-provider/owner-extension operational requirement on #860 remains open.

These three PRs extend the existing unreleased stack. Owning contract and safety documentation accompanies the relevant phase; the final recovery PR carries the shared canonical product documentation and cumulative QA checklist.

This PR’s latest [Python CI](https://github.com/ebarti/JobCtrl/actions/runs/34125370619) and [TypeScript CI](https://github.com/ebarti/JobCtrl/actions/runs/34125370632) passed at its published head. All five TypeScript jobs passed, including E2E and Storybook; other applicable workflows passed and DCO was explicitly skipped. The final cumulative Python CI also passed lint, all 3841 tests (2 optional skips), and package builds on Python 3.11, 3.12 and 3.13.

### DIFF
--- a/workers/automation/src/jobctrl/domain/materials/use_cases.py
+++ b/workers/automation/src/jobctrl/domain/materials/use_cases.py
@@ -1700,11 +1700,13 @@ def build_master_tailor_prompt(
     require_resume_master(profile)
     resume = get_resume_master(profile)
     required_experience_ids = get_required_experience_entry_ids(profile)
+    required_bullets = get_required_bullets_by_experience_id(profile)
     evidence_entry_ids = {
         item["experience_entry_id"] for item in get_achievement_evidence(profile)
     }
-    required_roles_without_evidence = [
-        entry_id for entry_id in required_experience_ids if entry_id not in evidence_entry_ids
+    required_roles_allowing_empty_bullets = [
+        entry_id for entry_id in required_experience_ids
+        if entry_id not in evidence_entry_ids and not required_bullets.get(entry_id)
     ]
     required_skill_ids = get_required_skill_category_ids(profile)
     all_experience_entries = get_experience_entries(profile)
@@ -1749,7 +1751,6 @@ def build_master_tailor_prompt(
         for entry in education_entries
     ]
 
-    required_bullets = get_required_bullets_by_experience_id(profile)
     tailoring_policy = get_tailoring_policy(profile)
     writing_style = get_writing_style(profile)
     custom_tailoring_prompt = get_custom_tailoring_prompt(profile)
@@ -1855,9 +1856,10 @@ HARD RULES:
 - If a required role has target-covered or explicitly pinned evidence, include
   only those bullets and no positioning-only filler. If it has neither, include
   exactly one positioning bullet citing an achievement from that role when one exists
-- For REQUIRED EXPERIENCE IDS WITHOUT ACHIEVEMENT EVIDENCE, return bullets: []
+- For REQUIRED EXPERIENCE IDS ALLOWING EMPTY BULLETS, return bullets: []
   and title: "". The code preserves the source role details. Do not invent a
-  positioning bullet or cite an achievement from another role
+  positioning bullet or cite an achievement from another role. This exception
+  applies only to roles with neither achievement evidence nor required bullet pins
 - non_requirement_reason is a required fallback classification. Choose pinned,
   positioning, or structure. When coverage_edge_ids is non-empty it is ignored;
   when coverage_edge_ids is empty it must truthfully classify the claim
@@ -1908,8 +1910,8 @@ WRITING STYLE:
 REQUIRED EXPERIENCE IDS:
 {json.dumps(required_experience_ids, ensure_ascii=False)}
 
-REQUIRED EXPERIENCE IDS WITHOUT ACHIEVEMENT EVIDENCE:
-{json.dumps(required_roles_without_evidence, ensure_ascii=False)}
+REQUIRED EXPERIENCE IDS ALLOWING EMPTY BULLETS:
+{json.dumps(required_roles_allowing_empty_bullets, ensure_ascii=False)}
 
 REQUIRED SKILL CATEGORY IDS:
 {json.dumps(required_skill_ids, ensure_ascii=False)}
--- a/workers/automation/src/jobctrl/domain/materials/use_cases.py
+++ b/workers/automation/src/jobctrl/domain/materials/use_cases.py
@@ -168,6 +168,7 @@ from jobctrl.domain.profile.achievement_metrics import (
 from jobctrl.model_defaults import DEFAULT_PIPELINE_LLM_MODEL_SPEC
 from jobctrl.domain.tenant import LOCAL_TENANT, TenantId
 from jobctrl.resume_profile import (
+    get_achievement_evidence,
     get_custom_tailoring_prompt,
     get_education_entries,
     get_experience_entries,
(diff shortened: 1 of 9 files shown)


workers/automation/src/jobctrl/domain/materials/use_cases.py
```diff
@@ -1690,6 +1700,12 @@ def build_master_tailor_prompt(
     require_resume_master(profile)
     resume = get_resume_master(profile)
     required_experience_ids = get_required_experience_entry_ids(profile)
+    evidence_entry_ids = {
+        item["experience_entry_id"] for item in get_achievement_evidence(profile)
+    }
+    required_roles_without_evidence = [
+        entry_id for entry_id in required_experience_ids if entry_id not in evidence_entry_ids
+    ]
```
ebarti: **Low — the prompt lists a role with pinned required bullets under "WITHOUT ACHIEVEMENT EVIDENCE" while the validator rejects the empty list it instructs**

Invariant: schema, prompt, field validation and curation must agree on what an evidence-free required role may return, which is this PR's stated goal.

Evidence: `required_roles_without_evidence` is computed from achievement evidence only. `_validate_master_json_fields` (`services.py:237-241`) rejects `bullets: []` when the same entry has `required_bullets_by_experience_id` pins. In-memory probe at head: an entry `earlier_role` with `bullets: []`, `achievement_evidence: []` and one pinned bullet is listed under `REQUIRED EXPERIENCE IDS WITHOUT ACHIEVEMENT EVIDENCE`, and `validate_json_fields` returns `Experience update 'earlier_role' must include bullets` for the instructed empty list. The configuration is reachable: `removeBullet` in `apps/web/src/contexts/profile/components/StructuredProfileEditor.tsx:453` leaves the pin map untouched, and neither `TailoringRules.from_dict` (`domain/profile/value_objects.py:642`) nor `packages/contracts/src/schemas.ts:1776` requires a pinned bullet to exist on the entry.

Consequence: for such a profile the generator is told to do the one thing the validator rejects, so every attempt fails `validation_failed` and the code-owned retry guidance cannot converge. The README and user-doc statement that evidence-free required roles keep their details without generated bullets does not hold for this case. The underlying inability to converge is pre-existing (every bullet must cite exactly one evidence id from its own role), so this PR adds the contradiction without changing the outcome.

Required outcome: the prompt list and the validator agree for pinned-but-evidence-free roles, either by excluding pinned roles from the list and documenting it or by rejecting the configuration explicitly, with a regression covering that combination.

_Reproduced by in-memory probe against head; reachability reasoned from code._

ebarti (commented):
Review of `a72c3078008482345c7ff14937cf8bc02511540d` against its parent `36fe02dff14f5efcfe3867f58e6d98074f5271e4` (#872 head; independent `pr-reviewer` pass, coordinator-published). Only this PR's diff was reviewed. Posted as a comment review; the gate below is the review's own assessment, not a GitHub changes-request.

**Result:** Gate PASS. No Blocker/High/Medium. One Low posted inline (the prompt lists a role with pinned required bullets under "WITHOUT ACHIEVEMENT EVIDENCE" while the validator rejects the empty list it instructs).

**Verified in an isolated checkout at head (clean before and after; `uv.lock` untouched):** focused pytest for the new and changed cases across the three touched test files (7 passed, 167 deselected); the three files in full, `tests/test_materials_use_cases.py`, `tests/test_requirement_led_tailoring.py`, `tests/test_tailor_provenance_integration.py` (174 passed); ruff on `services.py` and `use_cases.py` clean; `git diff --check` on the range clean. An in-memory probe confirmed that a required entry with `bullets: []` and no achievement evidence passes field validation, assembles as heading plus location/date lines with no bullet lines and no empty-section marker, yields zero bullets from `build_resume_document`, and renders through `build_resume_html` as the existing `resume-entry--no-bullets` article with no `<ul>`; `extract_achievement_metrics` returns `("37 years",)` for a digit quantity and `()` for "nearly four decades". Not run here: API/web vitest (no TypeScript in the diff), docs build (the new README anchor target `## Automatic Preparation Recovery` exists in `docs/architecture/pipeline/operations.md`), live judge or adversarial behavior.

**Investigated and refuted (6):** empty bullet list breaks assembly, PDF/HTML rendering or the API text-to-HTML path (assembler, document builder, HTML renderer and `resume-review-drafts.ts` all tolerate it); provenance for an evidence-free role records a fabricated evidence id (rows exist only per rendered bullet and `_evidence_ids_for_entry` draws only from plan evidence; the integration test asserts exactly one experience row); the summary rule lets the model introduce new numbers (`_claim_metric_binding_errors` rejects any summary metric absent from mapped evidence, backed by the unknown-metric quality check and the never-fabricate corpus scan); an invented positioning bullet passes by citing a nonexistent evidence id (`build_tailoring_plan` always builds a coverage graph, so `validate_generated_claim_mappings` rejects unknown ids; empty and other-role ids fail curation); prompt, validator and curation use different evidence sources (all three derive from `get_achievement_evidence`); the `tailored-resume.v4` schema-version bump breaks the API read model (no non-test TypeScript branches on the version string).

**Residual risk, not counted:** `build_judge_prompt` does not tell the judge that evidence-free roles legitimately carry no bullets, so a strict judge could report dropped required evidence; only live QA can observe this. The new "omit the quantity and use qualitative wording" guidance can elicit spelled-out quantities that neither `extract_achievement_metrics` nor the digit-based never-fabricate regex detects (pre-existing gap, judge-only coverage). Doc paragraphs added here for parent-stack behavior (preparation recovery, Codex SDK env isolation) were not verified against #871/#872 behavior beyond the anchor target.

workers/automation/src/jobctrl/domain/materials/use_cases.py
ebarti: Fixed in 31476bc5. Roles with required bullet pins are excluded from the empty-bullet prompt exemption. Grounding and validators remain unchanged. README and owning docs explain that unsupported pins require restoring evidence for the role or removing the pin. The 2x2 evidence/pin regression reproduced the contradictory prompt before the fix; 88 focused tests, Ruff and docs build now pass. Independent review and a two-role probe passed.

PR #874, opened 2026-09-07
fix(profile): synchronize baseline fields and preserve newer edits

ebarti:
Editing the fifth experience title could overwrite the fourth because repeated editor text was matched by content instead of its canonical field identity. Bind every baseline field to its exact source path, and retain local edits made while a save is in flight. Saved acknowledgements update the matching revision without replacing newer text.

Adds regression coverage for repeated titles, all structured baseline fields, canonical identity, and overlapping saves, plus owning Profile and QA documentation. The separate review-draft identity guards remain intact.

Validation at `74e2a3c919ad7f5a7d50a513ae948d949dd26441`:

- `corepack pnpm check`, `corepack pnpm docs:build`, and `git diff --check` passed.
- Aggregate components passed: 123 script tests, 27 demo unit tests and 4 demo Node tests, 828 API tests, production web build, 29 extension unit tests and 4 extension E2E tests, and 3,841 Python tests with two unchanged optional skips.
- `corepack pnpm launcher:test` passed separately with the documented explicit migration-test Python and Node runtimes. The earlier aggregate invocation reached this component but selected an unsuitable system Python; this is component-level evidence, not a claim that the interrupted aggregate passed.
- Independent review and product QA returned Gate: PASS with no findings on the patch-identical implementation. The actual Profile/API/browser path verified exact field binding and preservation of newer edits.

This PR extends the unreleased stack above #873. The following Fetch-recovery head passed the literal cumulative `corepack pnpm check` and `corepack pnpm test`; its product and dependency bytes remain unchanged through the test-only repair. The separate authenticated-provider requirement on #860 remains open.

The preview E2E now verifies successful HTML responses, the canonical initial name, and a distinct cache key plus the updated rendered name after a real save. Its former fixed `v=0`/`v=1` expectation did not account for the accepted-snapshot lifecycle under development StrictMode. The targeted real API/SQLite/browser case and web typecheck passed, and independent review returned Gate: PASS with no findings. Only this test file changed; all product and dependency bytes match the previously validated heads.

Final published head `e5dce1bbddeb0e7615afd1402d544990cdfe6fcf` passes every latest applicable workflow. [Python CI](https://github.com/ebarti/JobCtrl/actions/runs/34161616234) passed lint, 3,841 tests with two unchanged optional skips, and package builds separately on Python 3.11, 3.12 and 3.13. [TypeScript CI](https://github.com/ebarti/JobCtrl/actions/runs/34161616212) passed all five jobs: 2,031 web units, 13 public type tests, API/shared-domain tests, Storybook, extension and browser E2E. Web E2E passed 93 cases; seven opt-in skips comprise two screenshot-capture tasks and five owned-profile cases already covered by independent local product QA. The corrected preview-refresh case passes in this full E2E run. Docs, demo, scripts and release checks pass; DCO is explicitly skipped.


ebarti (commented):
Review of `e5dce1bbddeb0e7615afd1402d544990cdfe6fcf` against its parent `a72c3078008482345c7ff14937cf8bc02511540d` (#873 head, equal to the merge-base; independent `pr-reviewer` pass, coordinator-published). Only this PR's diff (14 files, +616/−67, two commits; the second touches only `apps/web/e2e/tests/profile-edit.spec.ts`) was reviewed. Posted as a comment review; the gate below is the review's own assessment, not a GitHub changes-request.

**Result:** Gate PASS. Blocker 0, High 0, Medium 1, Low 1, both inline (skill-item projection locks into a permanent conflict after a transient duplicate; preview-mismatch conflicts reuse the "Profile data changed" message).

**Verified in an isolated checkout at head (clean before and after; no uv command run):** web vitest one file at a time: `profile-preview-fields.test.ts` 3 passed, `profile-form.test.tsx` 30 passed, `ProfileEditor.test.tsx` 12 passed; `git diff --check` clean. In-memory probes of `profileWithPlateChanges` at head and at base (source excerpts bundled with the worktree's esbuild, outside every checkout) covered identity binding with identical titles and reordered entries, duplicate ids, non-string fields, `unmapped:` ids, and skill-item and bullet edits through a sibling-equal intermediate value; the Medium was reproduced there. Read end to end: the binder, form projection, ProfileEditor wiring, ResumeAuditPins DOM import and snapshot, the API E2E fixture, the CSS rule, the four docs and both E2E commits, plus the API preview/PATCH routes, the worker renderer markup and `sanitize_text`. Not run here: Playwright E2E, typecheck, Storybook, Python tests (CI green at head).

**Base-defect reproduction:** the PR body's "fifth title overwrote the fourth by content matching" did not reproduce at `a72c30780`; there an `experience:<id>:title` Plate change was silently ignored because `profileWithPlateChanges` had no branch for it. The base defect this PR does fix is the preview URL keyed on `useProfileMutationCount()`, which changed after every successful autosave, refetched and remounted the Plate document, and discarded newer typing. The fix addresses that cause: the revision now advances only on clean init, save, or reset, and a stale acknowledgement leaves the newer draft and preview untouched ("Saved; newer changes pending"). The #865 review-draft identity guard is unchanged and not exercised by the new attributes. The inaccurate narrative does not affect the gate but should not be carried into the plan ledger.

**Investigated and refuted (6):** the E2E revision merely hides a StrictMode flake (StrictMode's doubled effect batches into one render and one preview request; the new assertion requires a cache key distinct from every key seen before the save plus the saved rendered name, which still proves the preview refreshed with saved content, while the old fixed `v=0`/`v=1` expectation asserted an implementation detail); the #865 identity guard regressed via the shared DOM import (guard textually unchanged; review documents carry no `data-resume-profile-*` attributes so both comparison sides are null; the change extractor is used only by the standalone editor); binder/renderer normalization mismatch writing wrong values (`renderedText` mirrors `sanitize_text` and the date branch mirrors `normalize_resume_date_range`; any disagreement yields `unmapped:` and a refused edit, never a write to another field); stale preview for a reused `?v=N` key (`preview.html` is served `cache-control: no-store` and the revision counter only increases); duplicate or reordered entry ids mis-targeting writes (`findEntry` requires exactly one id match; after "Move Role 5 up" the probe changed only role-5 and preserved identical-titled siblings and unknown `futureData` fields; `profile-form.test.tsx` covers the same case); a mount-effect loop from an unstable `onPreviewSourceChange` (the only caller passes a memoized handler and the effect returns early while dirty or submitting).

**Residual risk, not counted:** the three new isolated E2E cases and the two pre-existing cases in `structured profile persistence` are skipped in CI (`JOBCTRL_E2E_ISOLATED` unset) and were not run here (`cd apps/web && JOBCTRL_E2E_ISOLATED=1 corepack pnpm exec playwright test --config=e2e/playwright.config.ts tests/profile-edit.spec.ts --grep "structured profile persistence" --project=chromium --retries=0`); no automated test binds real Python renderer output, so renderer markup or normalization drift would surface only as `unmapped:` conflicts in manual QA; a settle refetch that differs from the PATCH response can advance the preview revision a second time after a save (one extra preview request, no data effect).

apps/web/src/contexts/profile/forms/profile-form.tsx
```diff
@@ -198,50 +213,106 @@ function profileWithPlateChanges(
...
+        const index = indexes.length === 1 ? indexes[0] : undefined;
+        if (index === undefined) {
+          appliedTargets.delete(change.semanticId);
+          conflictCount += 1;
+          continue;
+        }
+        applySingleText(change, items[index] ?? "", (value) => {
+          items[index] = value;
+          setPathValue(updatedProfile, `${prefix}.items`, items);
+        });
```
ebarti: **Medium — skill-item projection locks into a permanent conflict after a transient duplicate**

Invariant: a Plate edit to one skill item keeps following that item across keystrokes, and the form draft never persists an intermediate value the editor no longer shows.

Evidence: this branch resolves the target index by content on every apply. `expected` is the last applied text or the baseline (L285), the match must be unique (L286-291), and nothing stores the resolved index. The bullets branch does store it (`bulletIndex`, L320-342, L358-361) and re-resolves by index first. When a keystroke makes the edited item equal to a sibling (deleting `Script` from `JavaScript` next to `Java`, or ` 3` from `Python 3` next to `Python`), that apply succeeds and the draft becomes `["Java","Java"]`. The next apply finds two matches, deletes the remembered target and counts a conflict; every later apply falls back to `change.baselineTexts` (`JavaScript`), which no longer exists in the array, so the conflict never clears.

Consequence: the editor shows what the user keeps typing while the draft, and the autosave that follows, hold the transient duplicate. The status message blames Profile data changes the user never made, and reopening the editor does not help because the persisted array is already wrong.

Required outcome: remember the resolved index in `appliedTargets` for skill items the way bullets do (check that `items[index]` still equals the remembered text, else fall back to the unique content match or the `item:n` ordinal), and add a `profile-form.test.tsx` regression that types item 2 through a value equal to item 1 and asserts the final text lands with zero conflicts.

_Reproduced with an in-memory probe of `profileWithPlateChanges` at head (skill items `["Java","JavaScript"]`, editing `skills:<id>:item:2`): after `"Java"` the value is `["Java","Java"]`; after `"Java "` the result is `{"conflicts":1,"applied":[],"value":["Java","Java"]}` and every subsequent keystroke returns the same; the identical sequence on experience bullets applied cleanly._

apps/web/src/contexts/profile/forms/profile-form.tsx
```diff
@@ -198,50 +213,106 @@ function profileWithPlateChanges(
...
+      }
+      continue;
+    }
+
+    if (change.semanticId.startsWith("unmapped:")) {
+      conflictCount += 1;
+      continue;
+    }
```
ebarti, line 323: **Low — preview-mismatch conflicts reuse the "Profile data changed" message**

Invariant: a refused resume-editor edit tells the user a cause they can act on.

Evidence: `unmapped:` ids come from the binder when the rendered preview text does not equal the canonical profile value (`profile-preview-fields.ts:28-30`, `:49-51`) and unconditionally for `education:<id>:details` (`:91`). This block counts them as conflicts, so the only message shown is the one at L446: "Some resume editor changes were not applied because the matching Profile data changed. Save or discard those Profile changes, then reopen the resume editor." In the stale-preview, renderer-drift and education-details cases no Profile data changed, and reopening the editor reproduces the same unmapped binding.

Consequence: the user is sent in a loop (save or discard, reopen, same warning) with no indication that the field is not editable from the preview or that the preview is out of date with the saved profile.

Required outcome: count unmapped edits separately from real conflicts and show a cause-specific message (for example "This part of the preview cannot be edited here; edit it in the Profile form" for `education:<id>:details`, and "The preview is out of date with the saved profile; reopen it" for mismatches), with a unit assertion on the message for an `unmapped:` change.

_Reasoned from code at head; not reproduced in a browser._

apps/web/src/contexts/profile/forms/profile-form.tsx
ebarti: Fixed in 43a12af8. The projection retains the edited skill index and the completed skill-list snapshot through transient duplicates. Structural or boxed changes fail closed instead of stealing a sibling; cumulative preview edits refresh shared snapshots after the whole pass. Regressions cover continuous typing, undo/save, boxed replacement, removal, remove-plus-add, unique relocation and sibling edits in both orders. Verification: 39 form tests, five independent adversarial probes, six real save/reload browser scenarios and six repeated accessibility scenarios passed. Independent production review and QA passed.

apps/web/src/contexts/profile/forms/profile-form.tsx
ebarti: Fixed in 43a12af8. Unmapped preview targets are counted separately from canonical-data conflicts and direct the user to edit those fields in Profile data. The write gate remains unchanged. Two regressions reproduced the misleading message before the fix and now verify the actionable guidance without changing canonical data. The full 39-test form suite and independent review passed.

PR #913, opened 2026-09-11
fix(realtime): model DryRunCompleted as a typed domain event

### LINKED ISSUE #896: fix(realtime): model DryRunCompleted as a typed domain event
Carry explicit dry-run completion semantics through event parsing, SSE subscription and frontend handling.

## Current evidence

- **BK-077 — valid backlog:** The worker emits DryRunCompleted and the projection maps it to dry_run_complete, but the shared event enumeration/SSE listener list omits it. This confirms a typed event coverage gap, not that all dry-run UI updates fail (other run events/refetches can refresh the projection).

This is a source-based reassessment at the linked immutable revision. Product tests were not rerun for this migration; investigation items retain their stated uncertainty.

## Acceptance

- [ ] Add the canonical payload/type/enumeration and route the event to the relevant cache updates.
- [ ] Prove dry-run completion remains distinct from ApplicationSubmitted.
- [ ] Exercise the emitted event through parser/SSE/router/browser status fixtures without a real submission.

## Source and implementation context

- https://github.com/ebarti/JobCtrl/blob/a1526649a395b9febb4c20ef30dd946945014a21/workers/automation/src/jobctrl/apply/launcher.py#L1173-L1191
- https://github.com/ebarti/JobCtrl/blob/a1526649a395b9febb4c20ef30dd946945014a21/workers/automation/src/jobctrl/infrastructure/projections/projection_builder.py#L4007-L4027
- https://github.com/ebarti/JobCtrl/blob/a1526649a395b9febb4c20ef30dd946945014a21/packages/domain-types/src/events/index.ts#L599-L703
- https://github.com/ebarti/JobCtrl/blob/a1526649a395b9febb4c20ef30dd946945014a21/apps/web/src/shared/adapters/local/SseEventStreamAdapter.ts#L103-L113
- https://github.com/ebarti/JobCtrl/blob/a1526649a395b9febb4c20ef30dd946945014a21/docs/backlog.md#L437-L442

## Backlog classification

- Priority: **P2**; suggested order: 14.
- Type: bug; area: web.
- Review state: Needs scoped implementation/design review
- Intake provenance: owner-requested backlog revalidation; this descriptive metadata does not authenticate execution authority.
- Execution: Owner-originated backlog; prioritization is not implementation authorization. Conditional proposals start with investigation/design acceptance.

Tracked by #881. Migration and categorization do not put this item into an executable queue.

ebarti:
DryRunCompleted now participates in the shared TypeScript and Python event registries and the browser's named SSE subscription. Completion refreshes apply/workflow details, job views and the dashboard while leaving submission state unchanged. The payload matches both worker emitters, including snake_case lifecycle fields and numeric or null worker IDs.

The Chromium regression uses the production API and native EventSource with one synthetic completion event. It verifies that projection changes alone leave the status stale, then the event updates the visible status to dry_run_complete without reconnecting, navigating, submitting an application or changing the job's submission fields.

Validation: all 12 local checks passed, including 182 domain tests, 449 focused frontend tests, 4 Python registry parity tests and 1 Chromium regression, with no skipped tests. Type checks, public web type tests, web build, docs and Ruff passed. Independent review and QA passed on the integrated candidate.

The preexisting ApplicationSubmitted run-ID mismatch is deferred to #920; its review thread remains open.

Closes #896.


packages/domain-types/src/events/apply.ts
```diff
@@ -25,6 +25,35 @@ export function createApplicationSubmitted(
...
+
+/** Launcher payload as persisted and delivered by SSE (lifecycle keys stay snake_case). */
+export interface DryRunCompletedPayload {
+  readonly jobId: string;
+  readonly run_id: string;
+  readonly result: "dry_run_complete";
+  readonly finished_at: string;
+  readonly duration_ms: number | null;
+  readonly dry_run: true;
+  readonly worker_id: number | null;
```
ebarti, line 38: [P2] Match the worker ID type to both emitters (fixed)

Initial review found that this payload declared worker_id as string or null, while ApplyRun stores int or None and both process_manager.py and launcher.py emit that numeric value unchanged. String-only fixtures concealed the mismatch. The handler currently does not read worker_id, so this was a type-contract defect rather than a demonstrated status-refresh failure.

The current head corrects the field to number or null and exercises numeric IDs. This thread records the finding discovered before publication; independent fix verification follows.

apps/web/src/contexts/apply/handlers.ts
```diff
@@ -95,6 +96,20 @@ export const emailApplicationCandidateRecordedHandler = (
...
+  invalidate(applyRunsKeys.detail(event.tenantId, event.payload.run_id)),
+  invalidate(workflowRunsKeys.lists(event.tenantId)),
+  invalidate(workflowRunsKeys.detail(event.tenantId, event.payload.run_id)),
+  invalidate(applyReviewKeys.queue(event.tenantId)),
+  invalidate(jobsKeys.detail(event.tenantId, event.payload.jobId)),
+  invalidate(jobsKeys.lists(event.tenantId)),
+  invalidate(dashboardKeys.summary(event.tenantId)),
+];
+
 export const applicationSubmittedHandler = (
```
ebarti, line 113: [P2] Preexisting submission detail invalidations use the wrong run-ID key

The launcher emits ApplicationSubmitted with payload.run_id (workers/automation/src/jobctrl/apply/launcher.py:1250), and the SSE envelope preserves that key. This existing handler reads payload.runId for the exact apply-run and workflow-run detail keys at lines 119 and 121, so launcher submissions invalidate an undefined detail identity instead of the actual run.

Both sides of the mismatch are present on the base branch. List, job and dashboard invalidations remain active and can still reconcile broader views; this does not establish that all submission status updates fail. Keep this open for a separately scoped repair that normalizes or reads the emitted run ID and covers exact detail invalidation with a launcher-shaped event. The DryRunCompleted change uses run_id correctly.

packages/domain-types/src/events/apply.ts
ebarti: Corrected worker_id to number or null and replaced string fixtures with numeric IDs. Independent review traced both emitters back to their numeric worker state; independent QA observed worker_id: 0 in the native SSE frame. Factory cases cover 0, 2 and null, and parser/adapter assertions preserve the numeric value.

The final commit passed 182 domain tests, 449 focused frontend tests, the Chromium dry-run regression, all admitted type/build checks and both independent gates. The browser status reaches dry_run_complete while submission fields remain unchanged.

Fix commit: f767cd8a5e7b75f04e05072f17775849ec63b308

packages/domain-types/src/events/index.ts
```diff
@@ -650,6 +655,7 @@ export const DOMAIN_EVENT_TYPES = [
   "ApplyRunEventRecorded",
   "EmailApplicationCandidateRecorded",
   "ApplicationEmailFeedbackIngested",
+  "DryRunCompleted",
```
ebarti, line 658: [P2] Keep the Python event registry in sync

Adding DryRunCompleted here without its Python mirror breaks the repository's ordered event-enumeration contract. All three Python CI jobs fail test_python_domain_event_types_match_typescript_union at workers/automation/tests/test_domain_event_parity.py:70: Python has ApplicationSubmitted at index 53 while TypeScript now has DryRunCompleted.

Add the same event before ApplicationSubmitted in workers/automation/src/jobctrl/domain/events/__init__.py and include the existing cross-language parity test in the focused verification. The worker emitters already produce this event; their submission behavior does not need to change.

ebarti: Added DryRunCompleted to the Python catalog in the same position as the TypeScript registry. The existing parity assertion is unchanged and now passes; the focused verification recipe makes this cross-language contract explicit.

Independent review and QA verified exact ordered equality across all 102 event names, including a single DryRunCompleted entry. All four parity tests execute with no skips, Ruff passes, and all twelve final local checks and both independent gates pass. The browser regression still proves status refresh without submission. The supported Python CI matrix is rerunning on the pushed commit.

Fix commit: aaac4cf543c7016651a32cd7f9d14e3099295848

PR #916, opened 2026-09-11
chore: adopt retained workflow check evidence handoffs

ebarti:
Upgrade the workflow pin to 0.5.1 and document the retained check-evidence handoff. Reviewers receive the evidence ID, artifact hash and explicit state root; an older compatible attempt can inspect its original artifact while retaining its existing pin and review activation. The skill-resolution example now uses the required JSON request file.

Validation: independent Tier 1 review passed; documentation build and link/redirect checks passed (9,917 references across 366 pages); git diff --check passed.

Source correction: ebarti/devflow#8.

### DIFF
--- a/docs/developer/workflow.md
+++ b/docs/developer/workflow.md
@@ -20,12 +20,23 @@ coordinating, implementing, reviewing, verifying and delivering (each prefixed
 `devflow-`). A design-only conversation uses a method without creating work.
 The old `devflow` name forwards to the entry. No hooks or scheduler are installed.
 
-`skill list` reports the installed catalog. `skill resolve` with a JSON request
-such as `{"name":"devflow-coordinating"}` returns the selected release's file.
+`skill list` reports the installed catalog. Save a JSON request such as
+`{"name":"devflow-coordinating"}` in `skill.json`, then run
+`scripts/devflow skill resolve --request-file skill.json --json` to return the
+selected release's file. JSON input is not a positional argument.
 For continuation, include the existing `--work-id`; read that immutable path
 rather than substituting newer global skill text. Missing historical stages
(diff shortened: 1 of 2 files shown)


docs/developer/workflow.md
```diff
@@ -20,13 +20,24 @@ coordinating, implementing, reviewing, verifying and delivering (each prefixed
...
+selected release's file. JSON input is not a positional argument.
 For continuation, include the existing `--work-id`; read that immutable path
 rather than substituting newer global skill text. Missing historical stages
 require compatible recovery or a recorded upgrade. `next` names each action's
 owning skill and the separate assigned role skill.
 
+Review/QA handoffs include each check's evidence ID, artifact hash and explicit
+private state root. Retained JSON lives at `<state-root>/artifacts/<hash>` and
+contains the command output and available JUnit report. Verify the hash and use
+the pinned verification skill's check-evidence reference when present. An older
```
ebarti, line 35: [P2] Retained compatible pins need a valid evidence-reading route

Independent review found that the initial wording unconditionally required the new check-evidence reference, which is absent from a retained 0.5.0 attempt. That could cause another missing-file failure or imply an unnecessary attempt upgrade. The final paragraph now makes the reference conditional and permits standard hash/file/JSON inspection while preserving the existing pin and activation.

ebarti: The final paragraph explicitly handles an older compatible pin without the new reference. Independent adopter review passed, the documentation build resolved 9,917 references across 366 emitted pages, and whitespace validation passed. The active historical attempt is preserved for its separate resume checkpoint.

Fix commit: 2dba0000b424290bd77b48934a0472883f618e1d

PR #929, opened 2026-09-15
fix: preserve truthful job review evidence and material readiness

ebarti:
Job and application review could present unsupported compensation, misleading material readiness, and analysis process commentary. This change withholds salary estimates without matching seniority evidence, makes requirement cards independently expandable with match-based defaults, and separates accepted material availability from application prerequisites.

Ideal-profile generation now rejects process commentary and preserves the last accepted analysis on a failed refresh. Refreshed requirements only receive fit and resume coverage when the stored evidence matches their analysis generation and source requirement identity.

Validation: 312 focused API, web, and worker tests passed; API/web type checks, production web build, documentation build, and Python lint passed. Independent review and browser QA passed, including keyboard/mobile disclosure behavior and stale-evidence refresh cases.

First layer of a two-PR stack. The next layer adds seniority-specific regional salary lookup.


docs/user/compensation-evidence.md
```diff
@@ -62,7 +62,14 @@ for comparison:
    weight. A cost-of-living-only range is allowed at low confidence. The raw
    factor and numeric range remain visible even outside the supported `0.1x` to
    `10x` review bound, with a prominent `factor_out_of_bounds` warning.
-5. **Materialize the last good result.** The latest direct or extrapolated fact
+5. **Check the source population.** An all-level or unknown-level benchmark is
+   contextual evidence for a job with a known seniority, not an estimate of that
+   seniority's pay. Without matching level evidence, the estimate is
```
ebarti: **Medium** — This rule currently holds only on the canonical benchmark path. The explicit refresh (Job Detail focused compensation refresh, Jobs toolbar backlog refresh, `compensation-refresh` CLI) runs `refresh_compensation_facts` → `backfill_from_jobs` → `estimate_market_compensation` (estimator `company-role-reported-compensation-v3`) and upserts the per-job row. Reproduced at 908276f5: one Levels.fyi "All levels" aggregate for "Principal Software Engineer" yields `estimated_range 60000-90000`, band low 0.45, level factor 0.75, no `weak_level_match` (`_level_score` returns 0.75 for unknown levels; threshold 0.55). The TS/Python read overrides match only the `canonical-benchmark-` prefix, so that range renders.

Stacked #931 changes `estimate_market_compensation` to withhold here (its updated `test_levels_public_market_fallback_uses_aggregate_instead_of_top_payer_range` asserts `insufficient_evidence`). Required for #929 on its own: land together with #931, or scope this sentence to the automatic canonical benchmark and add a fixture for the explicit path.

workers/automation/src/jobctrl/domain/materials/analysis_content.py
```diff
@@ -0,0 +1,42 @@
...
+import re
+
+from jobctrl.domain.materials.analysis import JobAnalysis
+
+
+# These expressions identify narration about producing the artifact, not mentions
+# of experts, analysis, or models as actual candidate skills/responsibilities.
+_PROCESS_COMMENTARY = re.compile(
+    r"\b(?:both|all|the|these|our|\d+)\s+(?:independent\s+)?"
+    r"(?:experts?|models?|drafts?|analyses|assessments?)\s+"
```
ebarti: **Medium** — `_PROCESS_COMMENTARY` also matches ordinary domain prose: any determiner before experts/models/drafts/analyses/assessments followed by one of these verbs. Reproduced against this regex: "ensures the models converge under distributed training", "owns the models describing customer churn", "ensures the assessments identify learning gaps", "the analysis identifies exploitable paths" and "all models are described in model cards" are rejected, while "Experts converge on…", "The two drafts agree…" and "The combined analysis suggests…" pass.

Draft legs retry with the unchanged prompt (`ensemble.py:79`; only the synthesizer receives the rejection text), so an ML, clinical or assessment role can exhaust every leg → `EnsembleError` → the tailor stage fails, and the `PROMPT_VERSION` bump re-analyses every job on next use. Required: a fixture proving legitimate role prose about models/assessments/analysis is accepted, and rejection feedback on draft retries as the synthesizer already gets.

.devflow/checks.toml
```diff
@@ -212,3 +212,25 @@ description = "Explicit provider login reuse recovers stale app-owned auth while
 argv = ["env", "-u", "UV_PROJECT_ENVIRONMENT", "-u", "VIRTUAL_ENV", "-u", "UV_EXCLUDE_NEWER", "-u", "UV_EXCLUDE_NEWER_PACKAGE", "uv", "run", "--project", "workers/automation", "--locked", "--all-extras", "--exclude-newer", "false", "pytest", "-q", "workers/automation/tests/test_setup_probes.py", "workers/automation/tests/test_codex_home_isolation.py", "--junitxml={report_path}"]
 min_executed = 1
 timeout_seconds = 300
+
+
+[checks.browser-debug-api]
+kind = "junit"
+description = "Application audit and compensation projections preserve material availability and source-population applicability"
+argv = ["corepack", "pnpm", "--filter", "@jobctrl/api", "exec", "vitest", "run", "test/apply-audit.test.ts", "test/market-compensation-estimates.test.ts", "test/application-feedback.test.ts", "test/application-feedback-v7.test.ts", "test/read-model-v7.test.ts", "--reporter=junit", "--outputFile={report_path}"]
+scenarios = ["A01", "A03"]
```
ebarti: **Medium** — `scenarios` is new to this file, and these values are this work item's acceptance IDs rather than repository scenarios. In the pinned devflow, `domain/rules.py::missing_scenarios` marks a contract's `verification.scenarios` covered by the union of passing evidence `scenario_ids` and `acceptance_ids`, with no work scoping, and the operation guide defines recipe `scenarios` as "the predefined scenarios this recipe proves". No catalog in this repository defines A01–A04, so any later work whose contract names `A01`/`A02`/`A03`/`A04` (acceptance IDs are a documented alias) reports them covered as soon as one of these recipes passes there. The `browser-debug-*` names also describe a session rather than the invariants, unlike `discovery-api`, `provider-auth-recovery` and the other recipes. Required: descriptive stable recipe names and either named repository scenarios or no `scenarios` key (same at lines 217, 228, 235).

apps/web/src/contexts/materials/components/EmployerAnalysisPanel.tsx
```diff
@@ -549,66 +548,82 @@ function RequirementItem({
...
-            tone={assessment.tone}
-          >
+        <CardAction className="flex flex-wrap items-center gap-2">
+          <StatusBadge title={assessment.title} tone={assessment.tone}>
             {assessment.label}
           </StatusBadge>
+          <CollapsibleTrigger
+            render={
+              <Button
+                aria-label={`${open ? "Collapse" : "Expand"} requirement: ${requirement.text}`}
```
ebarti: **Low** — The trigger's visible label is "Show evidence"/"Hide evidence" (line 565) but its accessible name is "Expand requirement: …"/"Collapse requirement: …". WCAG 2.5.3 (Label in Name) requires the accessible name to contain the visible text; speech-input users saying "click Show evidence" cannot activate it, and screen readers announce a name that never appears on screen. axe's `label-content-name-mismatch` rule is experimental and off by default, so `EmployerAnalysisPanel.a11y.test.tsx` and Storybook do not catch it, and the unit/e2e tests target the aria-label directly. Required: an accessible name that includes the visible text and stays unique per card (for example "Show evidence for requirement: …"), with tests following the visible label.

ebarti (commented):
## Independent review — PR #929 at 908276f5

Reviewed base `ee552bbe` (main) → head `908276f5`, full merge-base diff (38 files) in an isolated detached checkout. Gate: **PASS** (no Blocker/High). Findings: 3 Medium, 1 Low, posted inline. 8 further candidates were investigated and dismissed with evidence.

**Verified**
- API vitest: apply-audit, market-compensation-estimates, application-feedback, application-feedback-v7, read-model-v7, projections — 141 passed.
- Web vitest: EmployerAnalysisPanel (+ a11y), CompensationEvidence, ApplyReviewView, JobDetailDrawer — 99 passed. domain-types — 183 passed.
- Worker pytest: the seven `browser-debug-worker` files plus test_projection_builder, test_requirement_led_tailoring, test_v6_to_v7_job_detail_projections — 232 passed; ruff clean on all changed Python paths.
- Canonical-path withholding: fresh materialization and the legacy-row overrides in TS `mapEstimateRow` and Python `_market_estimate_from_row` agree on state, band, score, level factor, evidence levelScore and reasons; no amount leaks; the projection version 3→4 rebuild is proven on both sides.
- Evidence binding: identical id + whitespace-normalised text rule in `requirement-evidence.ts` and `EmployerAnalysisPanel.tsx`; the artifact binding keys match what Python has written into `quality_plan.requirement_directives` since f991dc0b, so older artifacts show `not recorded` rather than silently losing coverage.
- Failed analysis refresh writes nothing (validation precedes `repository.save`), so the last accepted generation stays in repository, projection and API.

**Not run**: Playwright e2e/visual specs (need a live stack; the two PNGs show the expanded matched card with the new button, the collapsed default is asserted via DOM only), `api:check`/`web:check`/web build (CI green; no finding depends on them). Author browser-QA claims were not independently reproduced.

**Code references for the docs comment (outside this diff)**: `workers/automation/src/jobctrl/domain/compensation/market.py` `_level_score` (unknown → 0.75) and the 0.55 critical-factor threshold; `infrastructure/compensation/sqlite_market_repository.py` `backfill_from_jobs` → `save_estimate`; `infrastructure/compensation/refresh.py:102`.

**Residual risks (no inline comment)**
- Requirement-fit binding exists only in the API read paths and the web panel; persisted `requirement_fit_report_json` stays unbound in both projection builders (consistent today; any new consumer needs the same guard).
- `getJobDetail` now calls `resumeMaterialPreviewForJob` (disk read plus audit/template state) on every detail read just to select the artifact id; cost unmeasured.
- Daily digest `reviewNeededMaterials` no longer counts jobs blocked only by global attestations or an in-flight refresh; likely intended, not documented.
- Scorer fallback on a failed refresh is log-only, with no failure event (pre-existing).

A review PASS does not establish merge readiness beyond the checks listed here. Reviewer: `pr-reviewer` (claude-fable-5-1, max effort); coordinator verified anchors against the reviewed head.

docs/user/compensation-evidence.md
ebarti: Fixed in c6403d41: the step now attributes the population check to the automatic canonical benchmark, and the later paragraph scopes the reclassification to older canonical-benchmark estimates on passive reads and projection rebuilds, which is exactly what `mapEstimateRow` and `_market_estimate_from_row` enforce at this head. The explicit refresh path (`estimate_market_compensation`) adopts the same rule in the stacked #931, whose updated `test_levels_public_market_fallback_uses_aggregate_instead_of_top_payer_range` asserts `insufficient_evidence`; estimator code is intentionally untouched in this layer so the two PRs do not conflict.

workers/automation/src/jobctrl/domain/materials/analysis_content.py
ebarti: Fixed in da802de3: the detector now requires a process subject (experts/drafts/analyses/model legs, or models/assessments only with a plurality or ensemble determiner) at a sentence or clause start followed by an agreement or conclusion verb, never treats a passive participle as a process verb, and counts a bare "the analysis" only with a conclusion verb. All seven sentences you listed are accepted, and the four that slipped through ("Experts converge on…", "The two drafts agree on…", "The combined analysis suggests…", "Each expert independently identified…") are rejected. `_draft_with_retry` now re-asks with the same rejection feedback as the synthesizer through a shared `_with_rejection_feedback`.

Proof in `test_employer_analysis_ensemble.py`: `test_candidate_prose_accepts_domain_roles_about_models_assessments_and_analysis`, `test_candidate_prose_rejects_process_narration`, `test_draft_leg_retries_process_commentary_with_rejection_feedback` (second prompt carries the rejection, no leg failure) and `test_persistent_draft_process_commentary_is_a_recorded_leg_failure` (exhausted leg recorded as `AnalysisContentError`, survivor proceeds). Every pre-existing case still passes (141 passed across the five employer-analysis suites, ruff clean). `docs/architecture/materials.md` now states that draft and synthesis retries carry the feedback.

.devflow/checks.toml
ebarti: Fixed in 18fc0299: the recipes are now `review-evidence-api`, `review-evidence-web` and `review-evidence-worker`, and the `scenarios` keys are removed; argv file lists are unchanged. Verified with tomllib: the file parses, 38 recipes, none carrying `scenarios`.

apps/web/src/contexts/materials/components/EmployerAnalysisPanel.tsx
ebarti: Fixed in 45e4f6ba: the accessible name is now `Show evidence for requirement: <text>` / `Hide evidence for requirement: <text>`, so it starts with the visible label and stays unique per card; the visible text is unchanged, so the checked-in visual snapshots remain valid. `EmployerAnalysisPanel.test.tsx`, `JobDetailDrawer.test.tsx`, `jobs-drawer.spec.ts` and `route-visual-qa.spec.ts` now select the control by the visible label (`/^Show evidence/`, `/^Hide evidence/`, or the full name where a screen-level query needs uniqueness). 91 web tests pass and `corepack pnpm web:check` (whose tsconfig includes `e2e/**`) passes; the Playwright specs themselves run in CI.

ebarti:
Addressed review feedback in da802de3, 18fc0299, 45e4f6ba, c6403d41. da802de3 narrows the process-commentary detector to process subjects at sentence/clause start, feeds prose rejections back into draft-leg retries, and adds accept/reject, retry-feedback and leg-failure fixtures. 18fc0299 renames the three check recipes to `review-evidence-*` and drops the unscoped `scenarios` keys. 45e4f6ba makes the requirement disclosure's accessible name start with its visible label and updates the unit and Playwright selectors. c6403d41 scopes the compensation doc's source-population rule to the automatic canonical benchmark and passive reads (the explicit refresh adopts it in #931).

Checks run: pytest on the five employer-analysis suites (141 passed) and ruff (clean); vitest on EmployerAnalysisPanel, its a11y test, JobDetailDrawer and ApplyReviewView (91 passed); `corepack pnpm web:check` (pass, includes e2e specs); `corepack pnpm docs:build` (pass, 10049 references resolve); `.devflow/checks.toml` parses. Playwright e2e/visual specs were not run locally.

